PR #343, opened 2015-09-16
Introduce reference periods

fpagnoux:
Introduce some reference periods in openfisca_core/periods.py in order to simplify period uses in formula.

Eg : 

Before : 
```python
        period = period.start.offset('first-of', 'month').period('month')
        two_years_ago = period.start.offset('first-of', 'year').period('year').offset(-2)
        last_month = period.start.period('month').offset(-1)
```
After :
```python
        period = period.this_month
        two_years_ago = period.n_2
        last_month = period.last_month
```

### DIFF
--- a/openfisca_core/periods.py
+++ b/openfisca_core/periods.py
@@ -673,6 +673,29 @@ class Period(tuple):
         return self[0]
 
 
+    # Reference periods
+
+    @property
+    def last_3_months(self):
+        return self.this_month.start.period('month', 3).offset(-3)
+
+    @property
+    def last_month(self):
+        return self.this_month.offset(-1)
+
+    @property
+    def n_2(self):
+        return self.start.offset('first-of', 'year').period('year').offset(-2)
+
+    @property
+    def this_year(self):
+        return self.start.offset('first-of', 'year').period('year')
+
+    @property
+    def this_month(self):
+        return self.start.offset('first-of', 'month').period('month')
+
+
 def instant(instant):
     """Return a new instant, aka a triple of integers (year, month, day).
 


cbenz:
@eraviart I find it is clever and improves the readability. Let's discuss about it.

openfisca_core/periods.py
```diff
@@ -673,6 +673,29 @@ def unit(self):
...
+    @property
+    def last_3_months(self):
+        return self.this_month.start.period('month', 3).offset(-3)
+
+    @property
+    def last_month(self):
+        return self.this_month.offset(-1)
+
+    @property
+    def n_2(self):
```
cbenz: be more explicit?

MattiSG: Do you mean more doc or another name? "n-2" is very common in law.

MattiSG:
:+1:

openfisca_core/periods.py
cbenz: Here are my 2 questions:

* `n_2` is not obvious for me without explanation since I didn't understand that the `_` had to be interpreted as `-`. We talked about naming it `2_years_ago` instead.

* Hardcoding the offsets in the function name isn't nice, what do you think about `last_months(2)` or `n_years_ago(2)`?

And/or some doc.

MattiSG: I'm not sure there's a sensible way between changing the core `offset`/`period` API and this “just give me straight what I use 90% of the time”.

I think these are syntactic sugar over generic period manipulation functions and that they make sense mostly if hardcoded. Then again, if there was a way to isolate these helpers from `openfisca-core`, it would probably be better, but the current lack of modularity makes that kind of adjustments pretty hard.

fpagnoux: :+1: Matti, the idea is really to give shortcuts for periods used in the vast majority of formulas, so I don't think putting flexibility here would be relevant. 

About `n_2`, let's think about a more transparent term.  I thought about `2_years_ago`, but it's not really what we mean. If we are in April 2015, two years ago is April 2013, not 2013. Would  `n_2_year` be more explicit ? 

I also wondered if it was a good idea to put it in core, but the `period.last_year` notation seemed pretty intuitive to me, and I didn't have a better idea.

cbenz:
Very clever initiative :+1:

PR #348, opened 2015-09-28
Uniformize variable declaration : allow inputs to be declared as a formula without a function

fpagnoux:
As discussed with @eraviart last Friday, declaring formulas and input in different ways has a few drawbacks :
- More complexity for the developer, who has to know 2 ways of declaring a variable that look like each other with a slightly different syntax. 
- Possible mistakes (comas are needed in one syntax, not the other)
- When looking for a variable declaration with a global text research, inputs are hard to find.

It would be nice if we could have only one syntax. A possible implementation (started here) is to just allow `SimpleFormulaColumn` to not have a function. 

Before : 
```python
reference_input_variable(
    name = "ass_precondition_remplie",
    column = BoolCol,
    entity_class = Individus,
    label = u"Eligible à l'ASS",
    )

@reference_formula
class ass(SimpleFormulaColumn):
    column = FloatCol
    label = u"Montant de l'ASS pour une famille"
    entity_class = Familles

         return period, x
```

After :

```python
@reference_formula
class ass_precondition_remplie(SimpleFormulaColumn):
    column = BoolCol
    entity_class = Individus
    label = u"Eligible à l'ASS"

@reference_formula
class ass(SimpleFormulaColumn):
    column = FloatCol
    label = u"Montant de l'ASS pour une famille"
    entity_class = Familles

         return period, x
```

### DIFF
--- a/openfisca_core/formulas.py
+++ b/openfisca_core/formulas.py
@@ -1240,16 +1240,11 @@ def new_filled_column(base_function = UnboundLocalError, calculate_output = Unbo
     else:
         assert issubclass(formula_class, SimpleFormula), formula_class
 
-        function = specific_attributes.pop('function', UnboundLocalError)
+        function = specific_attributes.pop('function', None)
         if is_permanent:
-            assert function is UnboundLocalError
-        if function is UnboundLocalError:
-            assert reference_column is not None and issubclass(reference_column.formula_class, SimpleFormula), \
-                """Missing attribute "function" in definition of filled column {}""".format(name)
+            assert function is None
+        if reference_column is not None:
             function = reference_column.formula_class.function
-        else:
-            assert function is not None, """Missing attribute "function" in definition of filled column {}""".format(
-                name)
         formula_class_attributes['function'] = function
 
     # Ensure that all attributes defined in ConversionColumn class are used.


fpagnoux:
I'm not sure entirely sure this is ready to merge right now, but tests seem to be ok on my computer. Maybe the core team to have a look to make sure everything is not broken :)

benjello:
Why not but one should rename formula sot variable somehow. A formula without a function is a bit awkward.

fpagnoux:
Good point !
Something like that ?

```python
@reference_variable
class ass_precondition_remplie(SimpleVariable):
    column = BoolCol
    entity_class = Individus
    label = u"Eligible à l'ASS"
```
Btw, what does `reference` means in `reference_formula` ?

benjello:
reference is for "root" tax_benefit_system vs reform IIRC @cbenz @eraviart

fpagnoux:
Ok, I understand better, thanks.

cbenz:
Furthermore I would suggest renaming even `@reference_formula` to `@reference_variable`, because a variable can have or not a formula. The formula *is* the function.

We could even replace the term *reference* by either *vanilla* (like @benjello says often), *country* or *tax_benefit_system*

fpagnoux:
`@vanilla_variable` does not seem very explicit to me. What do you think about `@law_variable` ?

benjello:
@reference_variable is good for me

fpagnoux:
I tend to agree with @cbenz that for contributors who are only using Openfisca to run simulation about the actual law and are not trying potential reforms (e.g. embauche or mes-aides), the term `reference` doesn't mean much and is pretty unclear.

Ideally, it would be nice to have a term that makes sense both for the distinction reference vs reform, and for use cases without reform.

benjello:
If we use core_variable it will mess-up with openfisca-core.
 Should we use reference_variable or base_variable and have two distinct term for reform and extension ?

cbenz:
I close this pull request but opened a new issue: #355

Issue #349, opened 2015-10-03
Problem with erratic rates.


adrienpacifico:
When the observations are close enough (e.g 10 euros of gap between each salaire de base), the marginal rates are not correct.

See [this notebook](https://github.com/adrienpacifico/bug-openfisca/blob/master/decote_en_dur_3%20bug%20report.ipynb) as an exemple.

benjello:
Please try with a single and then we will look at the coupe. I suspect that you mix a foyer value with indvidual value and thus inducing a mismatch in vector sizes.

benjello:
You can take a look at https://github.com/openfisca/openfisca-matplotlib/tree/master/openfisca_matplotlib
for drawing rates. Notebook examples would be nice ;-)

adrienpacifico:
> Please try with a single and then we will look at the coupe. I suspect that you mix a foyer value with indvidual value and thus inducing a mismatch in vector sizes.

Here it is : [Notebook rates single](https://github.com/adrienpacifico/bug-openfisca/blob/master/decote_en_dur_3%20bug%20report%20single.ipynb)
The problem is still here. I am suspecting wrong rounding (or something else) with the core.

> You can take a look at https://github.com/openfisca/openfisca-matplotlib/tree/master/openfisca_matplotlib
for drawing rates.

Yes thanx, test_graph.rates() doesn't work for me @home (If I wish I could make it work, I think).

The problem with rates doesn't occur if I'm doing only 100 simulations, but at 200 I'm stating to get some artefacts (that I can smooth), and around a 1000 I'm getting what you've seen on my notebook.

But at 1000 simulations the spread between two net wages (or revenu brut global) is around 30 euros... I don't think we should observe that with such large spread.

benjello:
@adrienpacifico: some of the IR bases and the IR is rounded so it may explain that.
Can you produce two simple test-case with a high jump so we can try to understand things here before calling in @eraviart and @cbenz ;-)

adrienpacifico:
Yep, I will.

benjello:
@adrienpacifico Should we close this ?

adrienpacifico:
Maybe, I'm still not satisfied with the provided solution and I'm kind of working on one. This still could be a space where we can talk about it. 

But I you prefers to close it, to keep only "real" issues opened, you can without hesitating.

adrienpacifico:
Thanx by the way for your help, that helped me to understand the issue I was facing.

PR #350, opened 2015-10-10
Add trim option to marginal rate computation

benjello:


benjello:
@adrienpacifico: you can use this to trim the large marginal rate you may hit because of nonlinearities like seuil de non versement and kinks.
@cbenz you can merge this with no impact for other openfisca users

adrienpacifico:
@benjello 
Thanx a lot. I did not try your trim function. 
To deal with that I'm currently using a smoothing function that gives the average value between n-1 and n+1 element of the array if the variation is too large.

What should be the typical value of your trim parameter ?

benjello:
trim is a list of two elements. The marginal rates that are higher than the max of these two parameters or less than the min are converted to nan and thus not appear in the plots.

adrienpacifico:
Thanx for you answer.
It solves big arterfacts that are higher than a given value (due to threshold, kinks, etc). But it doesn't solve  my fat lines issues :-) (as reported in [#349](https://github.com/openfisca/openfisca-core/issues/349)).

I need to find a good way to deal with that, other than reducing the number of observations :-).
(désolé je me sers un peu de cet espace comme si c'était un canard en plastique).

P.s : you might want to put a docstring for the function such that ones know that trim is supposed to be a two element array/list/tupple/whatever else.

benjello:
Your problem is due to rounding. Reduce the number of points.

Le sam. 10 oct. 2015 16:44, adrienpacifico <notifications@github.com> a
écrit :

adrienpacifico:
Yes I know I've been trying to do something a bit more fancy today, it's
not working perfectly but it's still something to improve.
Le 10 oct. 2015 18:21, "Mahdi Ben Jelloul" <notifications@github.com> a
écrit :

benjello:
@cbenz, you can merge this without trouble

PR #352, opened 2015-10-19
Allow SimpleFormula function to return a scalar number

fpagnoux:
To be able to replace [cryptic lines](https://github.com/openfisca/openfisca-france/blob/master/openfisca_france/model/prestations/education.py#L52) such as :
```python
nb_enfants_college = zeros(len(rfr))
```
by 
```python
nb_enfants_college = 0
```
If a function returns a scalar, it will be automatically be turned into a numpy array of the right size.

### DIFF
--- a/openfisca_core/formulas.py
+++ b/openfisca_core/formulas.py
@@ -529,8 +529,14 @@ class SimpleFormula(AbstractFormula):
             u"Function {}@{}<{}>() --> <{}>{} returns an output period that doesn't include start instant of" \
             u"requested period".format(column.name, entity.key_plural, str(period), str(output_period),
                 stringify_array(array)).encode('utf-8')
-        assert isinstance(array, np.ndarray), u"Function {}@{}<{}>() --> <{}>{} doesn't return a numpy array".format(
-            column.name, entity.key_plural, str(period), str(output_period), array).encode('utf-8')
+
+        if not isinstance(array, np.ndarray):
+            try:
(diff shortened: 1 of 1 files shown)


benjello:
+1. I think we had a cleaner way tha these zeros but after quick search, i didn't find it. And I am busy with other stuff today

fpagnoux:
I think the other way was like [here](https://github.com/openfisca/openfisca-france/blob/master/openfisca_france/model/prestations/minima_sociaux/aah.py#L129) to calculate another variable of the same entity and then to multiply by 0.
```python
nb_eligib_aah = 0.0 * br_aah
```

benjello:
We habe another way of doing that using an attribute of entitites that count the nomber of disctinct entitities IIRC

benjello:
@fpagnoux,  @cbenz : error produced when returning a scalar instead of a vectorare usually very useful information when you debug your code. I am afraid that errors can easily sneak in without notice. Any scalar coul easily propagate. It might less problematic for your test-case use but for survey data it can be destructive. So I have mixed feelings about this PR.

cbenz:
OK, I thought you were OK with your first comment.

We are discussing it again with @fpagnoux

cbenz:
After discussion we decided to cancel the merge (43f081e8ba2d8c714bc8c6405f2cb9f4afec74f7) and to open a new issue: #354

cbenz:
=== Merge cancelled ===

Issue #347, opened 2015-09-22
Don't accept circular references


eraviart:
When AL requires AL (month - 1) and AL (mont - 1) requires RSA (month-4 to month - 2) break cycle immediately and return default value.

Try to do this for every formula, not only RSA & AL.

cbenz:
Now OpenFisca handles some cyclic dependencies cases, see unit tests https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/tests/test_cycles.py

PR #358, opened 2015-10-29
Manage circular formula calls

fpagnoux:
Connected to #347 
Depends on https://github.com/openfisca/openfisca-france/pull/349 and https://github.com/openfisca/openfisca-web-api/pull/45

Please make sure openfisca france tests are ok before merging.

What has been done here : 
- Repair the infinite loop detection that was partly implemented but disconnected
- Detect the circular definition with a time offset (when a formula at time T depending on the same formula at another time). Return the default value of the variable in this case, except if recursion has been explicitly allowed.
- Refactor the `Entity` class. Take it out of the computation chain, and let `Simulation` handle holders_by_name. 
- Make it possible to transmit parameters in the computation chain without having to explicitly declare them at each level, by using `**parameters`

I'm still facing non passing tests. In some reforms like `inversion_revenus`, there is a detection of an infinite loop. I don't fully understand the problem, so I'd be happy to have some help from openfisca team :) @cbenz @eraviart

### DIFF
--- a/openfisca_core/formulas.py
+++ b/openfisca_core/formulas.py
@@ -108,7 +108,7 @@ class AbstractEntityToEntity(AbstractFormula):
         keys_to_skip.add('_variable_holder')
         return super(AbstractEntityToEntity, self).clone(holder, keys_to_skip = keys_to_skip)
 
-    def compute(self, period = None):
+    def compute(self, period = None, **parameters):
         """Call the formula function (if needed) and return a dated holder containing its result."""
         assert period is not None
         holder = self.holder
@@ -264,7 +264,7 @@ class DatedFormula(AbstractGroupedFormula):
 
         return new
 
-    def compute(self, period = None):
+    def compute(self, period = None, **parameters):
         dated_holder = None
         stop_instant = period.stop
         for dated_formula in self.dated_formulas:
@@ -273,7 +273,7 @@ class DatedFormula(AbstractGroupedFormula):
             output_period = period.intersection(dated_formula['start_instant'], dated_formula['stop_instant'])
             if output_period is None:
                 continue
-            dated_holder = dated_formula['formula'].compute(period = output_period)
+            dated_holder = dated_formula['formula'].compute(period = output_period, **parameters)
             if dated_holder.array is None:
                 break
             self.used_formula = dated_formula['formula']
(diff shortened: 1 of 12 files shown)


openfisca_core/formulas.py
```diff
@@ -486,31 +484,46 @@ def compute(self, period = None, requested_formulas_by_period = None):
...
-        if period_requested_formulas is None:
-            requested_formulas_by_period[period_or_none] = period_requested_formulas = set()
-        else:
-            assert self not in period_requested_formulas, \
-                'Infinite loop in formula {}<{}>. Missing values for columns: {}'.format(
+
+        # We keep track in requested_values of the formulas that are being calculated
+        # If self is already in there, it means this formula calls itself recursively
+        # The data structure of requested_values is: {formula: [period1, period2]}
+        if self in requested_values.keys():
```
cbenz: `.keys()` is not necessary in Python

openfisca_core/formulas.py
```diff
@@ -486,31 +484,46 @@ def compute(self, period = None, requested_formulas_by_period = None):
...
+            # A formula can call itself with a time offset (different periods), but we want to make sure this doesn't result in an infinite race
+            # towards the dawn (or end) of time. Thus the number of recursion allowed must be explicitely specified as a paramater. If this number
+            # is exceeded, we return the default value of the column.
+            max_number_recursive_calls = 0 if not 'max_number_recursive_calls' in parameters else parameters['max_number_recursive_calls']
+
+            if len(requested_values[self]) > max_number_recursive_calls:
+                dated_holder = holder.at_period(period)
+                dated_holder.array = self.zeros() + column.default
+                if debug and not max_number_recursive_calls:
+                    log.info("Recursive call detected on formula {}. Recursion is by default deactivated, default value {} will be returned for period {}. Use max_number_recursive_calls parameter in compute or calcultate to allow recursion."
```
cbenz: `log.debug`

benjello:
@fpagnoux : inversion_revenus is a bit odd. Did you detect problems elsewhere ?

fpagnoux:
Only in `inversion_revenus`, travis integrations tests (which don't include it) are ok.

cbenz:
Merged into master

Issue #355, opened 2015-10-23
Simplify variables vocabulary and syntax


cbenz:
Following #348 discussion, here are some ideas.

Merge `@reference_formula` decorator with `SimpleFormulaColumn` metaclass.

Before:

```python
# Input variable
reference_input_variable(
    name = "age",
    column = IntCol,
    entity_class = Individus,
    )

# Variable with formula
@reference_formula
class irpp(SimpleFormulaColumn):
    column = FloatCol
    entity_class = FoyersFiscaux

    def function(self, simulation, period):
        return period, x
```

After:

```python
# Input variable
class age(Variable):
    column = IntCol
    entity_class = Individus

# Variable with formula
class irpp(Variable):
    column = FloatCol
    entity_class = FoyersFiscaux

    def function(self, simulation, period):
        return period, x
```

benjello:
It is clean and i follow you on that side. But didn't you loose the information by removing the  reference_formula decorator ?

cbenz:
No problem, sorry I didn't explain.

The Variable inherited class would be different in openfisca-france and extensions:

- In OpenFisca-France:

  ```python
  from .formulas import Variable

  class irpp(Variable):
      column = FloatCol
      entity_class = FoyersFiscaux

      def function(self, simulation, period):
          return period, x
  ```

- In an extension:

  ```python
  Reform = reforms.make_reform(...)

  class irpp(Reform.Variable):
      column = FloatCol
      entity_class = FoyersFiscaux

      def function(self, simulation, period):
          return period, x
  ```

benjello:
Ok then.

fpagnoux:
Any news on this topic ? I can help this week if needed.

cbenz:
We forget to handle a case when a reform updates a variable giving the `reference = ` attribute: 150325583ee13ec2900a7b5b76b22777f73b2d65 and test b84c00e33d315422c6e9d21936f1843250fdd106

cbenz:
Tests still don't pass with https://github.com/openfisca/openfisca-france-indirect-taxation

cbenz:
Merged in `master` of all projects, pushed, Travis tests ongoing (but all were OK on my machine).

In case of emergency please just do `git reset --hard <ref>` where `<ref>` is the last commit before merge.

cbenz:
Good work everyone!

cbenz:
[...]

PR #364, opened 2015-12-02
Simplify variables declaration (without decorator)

fpagnoux:
Connected to #355


fpagnoux:
Careful: huge impacts on openfisca-france (and probably elsewhere)

PR #374, opened 2016-01-22
Improve tax scale combination

benjello:


openfisca_core/legislations.py
```diff
@@ -48,7 +48,9 @@ def combine_tax_scales(self):
         """Combine all the MarginalRateTaxScales in the node into a single MarginalRateTaxScale."""
         combined_tax_scales = None
         for name, child in self.iteritems():
-            assert isinstance(child, taxscales.AbstractTaxScale), child
+            if not isinstance(child, taxscales.AbstractTaxScale):
+                log.info('Skipping {} with value {} because it is not a tax scale'.format(name, child))
```
cbenz: Please log unicode strings only (add the "u") :)

By the way is it really an info or is it an error, with a better error message?

benjello: It should not stop the computation so i won't call it a n error. It is more a warning than an info.

PR #377, opened 2016-02-02
Add base function requested_period_last_or_next_value

fpagnoux:
Introduces a new base function requested_period_last_or_next_value, similar to requested_period_last_value except it can take a value from the future is there is none in the past.

Exemple : 
Input : `alt = True` for a child on 2016-01
If I request `alt`in 2015-12, I will get True.

It is helpful for calculating values in the past more accurately, which is useful for ppa_fictive calculations.

Depends on https://github.com/openfisca/openfisca-france/pull/427
Depends on https://github.com/openfisca/openfisca-parsers/pull/10

### DIFF
--- a/openfisca_core/base_functions.py
+++ b/openfisca_core/base_functions.py
@@ -1,0 +1,99 @@
+import numpy as np
+
+def permanent_default_value(formula, simulation, period):
+    if formula.function is not None:
+        return formula.function(simulation, period)
+    holder = formula.holder
+    column = holder.column
+    array = np.empty(holder.entity.count, dtype = column.dtype)
+    array.fill(column.default)
+    return period, array
+
+
+def requested_period_added_value(formula, simulation, period):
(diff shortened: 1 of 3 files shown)


fpagnoux:
Does this look ok to everyone ? 
It is pretty low impact

fpagnoux:
As this is getting old, I'm planning on merging this tonight if there is no objection. 
If someone (@laem?) want to give it a quick review, the only changes are :
- base functions are moved to a different file (formulas.py is way too big)
- a new base function is added as mentioned.

benjello:
No objection.

laem:
Looks good to me, although I don't understand these functions fully yet.

I started writing a guide about how to write variables, which was starting to tackle this kind of date subtleties. @cbenz Have you found time to read it ? I can try to finish the draft and publish it on https://github.com/openfisca/openfisca-gitbook/

PR #378, opened 2016-02-02
Allow optional params in formulas

fpagnoux:
This PR is  really opened for discussion, as it may have some non trivial implication.

The integration of PPA in openfisca was challenging, as the mechanism is a bit different from usual prestations.
The value of ppa is the average of a quantity called `ppa_fictive` for the 3 previous months (M-3 to M-1). What is challenging is that this ppa_fictive depends on the values of  some quantities evaluated on the month of the demand (M). 
Therefore to calculate `ppa_fictive`, we need to know what month M we need to refer to. One solution could be to have `ppa_fictive` as a subfunction of `ppa`, but this would decrease a lot testability and ability to see the different steps of the calculation. This PR proposes to introduce a mechanism of optional parameters : `ppa` would call `ppa_fictive` with an extra parameter containing the month of demand. 
This brings flexibility, but disturbs a little the openfisca model : the idea of "the value of the ppa_fictive for 2015-01" would be a slightly ill-defined concept as it depends on an extra parameter.

### DIFF
--- a/openfisca_core/formulas.py
+++ b/openfisca_core/formulas.py
@@ -545,7 +545,7 @@ class SimpleFormula(AbstractFormula):
                     input_variables_infos = [],
                     variable_name = column.name,
                     ))
-            formula_result = self.base_function(simulation, period)
+            formula_result = self.base_function(simulation, period, **parameters)
         except CycleError:
             self.clean_cycle_detection_data()
             if max_nb_cycles is None:
@@ -1054,12 +1054,22 @@ def calculate_output_divide(formula, period):
 def dated_function(start = None, stop = None):
     """Function decorator used to give start & stop instants to a method of a function in class DatedVariable."""
     def dated_function_decorator(function):
-        function.start_instant = periods.instant(start)
-        function.stop_instant = periods.instant(stop)
-        return function
+        function_wrapper = get_function_wrapper(function)
+        function_wrapper.start_instant = periods.instant(start)
+        function_wrapper.stop_instant = periods.instant(stop)
+        return function_wrapper
 
     return dated_function_decorator
 
+def get_function_wrapper(function):
+    def function_wrapper(formula, simulation, period, **parameters):
+        nb_arguments = function.func_code.co_argcount
+        if (nb_arguments > 3):
+            extra_parameters_names =  function.func_code.co_varnames[3:nb_arguments]
+            filtered_parameters = {param: parameters.get(param, None) for param in extra_parameters_names}
+            return function(formula, simulation, period, **filtered_parameters)
+        return function(formula, simulation, period)
+    return function_wrapper
 
 def last_duration_last_value(formula, simulation, period):
     # This formula is used for variables that are constants between events but are period size dependent.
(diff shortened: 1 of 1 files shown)


fpagnoux:
Exemple : 
In ppa:
```python
ppa = sum(simulation.calculate('ppa_fictive', period2, referencePeriod = period)
            for period2 in [m_1, m_2, m_3]) / 3
```
In ppa_fictive: 
```python
def function(self, simulation, period, referencePeriod = None):
    period = period.this_month
    referencePeriod = referencePeriod if referencePeriod else period
    forfait_logement = simulation.calculate('rsa_forfait_logement', referencePeriod)

```

fpagnoux:
What do you think ? Has this been already explored in the past ?

benjello:
I am not against adding parameters to function. In this case, however you can find a way to keep a testable solution although an ugly one: you can define ppa_fictive_m_1, ppa_fictive_m_2 and ppa_fictive_m_3  and they share all three the same reference period.
Since this is particularly ugly, I wont mind to have a parameter to distinguish between the three different ppa_fictive but it would be related to the previous months. 
````python
def function(self, simulation, period, fictive_period = None):
    period = period.this_month
    fictive_period = fictive_period if fictive_period else period
    forfait_logement = simulation.calculate('rsa_forfait_logement', period)
````
You can also use those as helpers and test them by an extension that would build a bridge to these helpers.

I am open but my advices are try not to disturb the actual model, try to avoid create multiple variable (although it increases testability, it increases complexity, memory burden for other usage etc).

 Hope it helps

MattiSG:
> you can define ppa_fictive_m_1, ppa_fictive_m_2 and ppa_fictive_m_3 and they share all three the same reference period.

Indeed, we had thought about this solution  :)
This could be a way to make visible the three different periods, but wouldn't decrease duplication. I guess it depends on how much these three different months are “concepts”, or formulas well defined enough to be made computable on their own. But in the end, they should probably all call a (private) function and pass it a parameter. The question is more: can formulas have additional parameters, or is it just an implementation detail?

benjello:
My point was mainly about which period should be the relevant period argument.
I definitely would go for including the computation of ppa_fictive _in_ ppa via a helper to avoid additional parameters. And have ppa_fictive_m_1, ppa_fictive_m_2 and ppa_fictive_m_3 callign this helper in an extension for test purposes. But i am not a software engineer ;-)
@eraviart @cbenz what do you think ?

fpagnoux:
We indeed thought about the `ppa_fictive_m_1` solution, but the main reason not to go for it was that some function called by `ppa_fictive` also need to know the referencePeriod (mainly for reinjecting computed prestations).
Then `ppa_fictive_m_1` would depend on `ppa_base_ressources_m_1`, which depends on `ppa_ressources_hors_activite_m_1`, and this becomes a nightmare :)

fpagnoux:
> You can also use those as helpers and test them by an extension that would build a bridge to these helpers.

Independently of this PR, being able to test and access intermediate result without making them a "pure" openfisca variable (especially if we know they will be calculated only once) is clearly interesting.
This could be an idea of enhancement (it came up in several conversations) but imo the Extension method is a bit heavy.

benjello:
We  can allow non-pure openfisca variable to have additional parameters but not the pure openfisca variables ;-)
I am open to everything but what may ease testing may confuse many other users, this is why i thought to write an extension to test the ppa since I don't expect to have a many cases like this one.
But once again, i do not feel entitled to point _the_ right choice.

fpagnoux:
Er, there are indeed many drawbacks about these extra parameters... All the cache system has to be rethought and that sounds like trouble. I think your solution could be better, provided we have a easier way of building extensions. This is also something we need for Paris : we have a dozen files of code about Paris prestations, and integrating them in a single reform file is something we won't do.
I'm closing this PR.

PR #379, opened 2016-02-23
Make possible to add extra parameters to a formula

fpagnoux:
This is a revival of  https://github.com/openfisca/openfisca-core/pull/378, but differently implemented, with the needed cache adaptation.

The idea is to allow a formula to take an extra parameter. For instance : 
```py
class ppa_fictive(Variable):
    # (...)
    def function(self, simulation, period, reference_period):

class ppa(Variable):
    # (...)
        period2 = period.last_month # for example
        simulation.calculate('ppa_fictive', period2, extra_params = [period])
```

The cache has been adapted so that is doesn't return a value cached for a specific set of extra parameters when called with a different one. In this case, the content of the cache would be:
```py
holder._array_by_period[period] = 
    { (param_value_1,) : result1, (param_value_2,) : result2 }
```
I slightly refactored it, trying to expose two clearly named methods to the outside : `put_in_cache(self, value, period, extra_params = None)` and `get_from_cache(self, period, extra_params = None)`.

I also adapted the conversion to JSON (here and in the API). 
```json
"ppa_fictive": {
    "2015-11": {
        "{reference_period: 2015-12}": 533.10595703125,
        "{reference_period: 2016-01}": 377.430419921875
    },
    "2015-10": {
        "{reference_period: 2015-12}": 428.10595703125,
        "{reference_period: 2016-01}": 272.430419921875
    },
    "2015-12": 283.0704345703125,
    "2015-09": 0
},
```
As long as there is only one result, for a variable and a period, there will be no difference with the actual behavior. The json will be more complex only if a formula has been called with two different set of extra parameters. This happens for instance with ppa if you compute it for several months in a row.


fpagnoux:
@benjello, we tried to use an extension solution you suggested but we dropped it as:
- There would have been around around 20 weird `formula_name_m_k` in it
- There would have been a lot of recomputation (once in the proper openfisca computation, once when we call `formula_name_m_k`.
- Having the extensions variables computed and returned to mes-aides would have needed some tricks (like calling `ppa_mes_aides` that calls all the other ones.

benjello:
No objection from my point of view.

laem:
Same for me. It's an interesting feature.

PR #382, opened 2016-03-30
Fix sum_by_entity

fpagnoux:
The result size must be the target entity'one, not necessarily the calling formula.


fpagnoux:
Not a big deal, so if I don't get a review I'll merge it by Thursday night ;)

PR #385, opened 2016-05-19
Implement mechanism to opt out of cache for some variables

fpagnoux:
When using it for a large population, having too many variables in cache make openfisca performances drop.

This introduces a mechanism that enables some variables not to be cached.

The names of the variables we don't want to cache must passed to the tax benefit system in a set : 
```python
tax_benefit_system.cache_blacklist = set(['intermediate'])
```
Then an extra parameter must be given when creating the simulation to activate this behaviour:
```python
simulation = scenario.new_simulation(debug = True, opt_out_cache = True)
```

See the [test file](https://github.com/openfisca/openfisca-core/blob/cache_opt_out/openfisca_core/tests/test_opt_out_cache.py) for a full exemple.

@cbenz @eraviart @laem @benjello any feedback ? Is naming clear enough ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.1.0 – [diff](https://github.com/openfisca/openfisca-core/compare/1.0.0...1.1.0)
+
+* Implement cache opt out system
+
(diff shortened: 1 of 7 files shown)


fpagnoux:
See the use in `openfisca-france` [here](https://github.com/openfisca/openfisca-france/pull/447).

PR #388, opened 2016-06-09
Remove metaclasses

fpagnoux:
The initial goal of this PR was to get rid of the metaclass system. As is it pretty deep in the core of openfisca, a lot of other changes comes with it. **Feedback and review from other contributors would be  particularly appreciated here**

#### Metaclass system
In the former system, openfisca variables were loaded into the tax benefit system statically through metaclasses. Concretely, when python was importing the different modules files, the variables would be **automatically**  loaded in the entities **classes**. Then only the tax benefit system would be created, and would be able to access these variables through the entity classes. 
This is pretty lopsided for many reasons. The lifecyles of the different objects is pretty hard to handle, as many things happen at the loading of the python modules. This make it very hard for reforms to change the tax benefit system (hence the rough API).
This mechanism is removed with this PR. See next comment about the new TaxBenefitSystem API to check the new mechanism.

--------------------------------------

To do :
- [ ] Update documentation 
- [x] Remove loading extensions through git
- [x] Make update_variable more robust (see note)


fpagnoux:
#### Creating a tax benefit system.
A tax benefit system is a set of :
* Variables (some of them calculable)
* Entities for which these variables are defined
* Legislation parameters. 

**Example**
``` py

# This hasn't change: I need to create my entities and scenario.

class Scenario(AbstractScenario):
	#...

class Familles(AbstractEntity):
	#...

class Individus(AbstractEntity):
	#...


# Variable classes can still inherit Variable, DatedVariable, EntityToPersonColumn or PersonToEntityColumn.
class age_en_mois(Variable):
	#...

class birth(Variable):
	#...

class salaire_brut(Variable):
	#...

# I want to define a particular tax benefit system.
class TestTaxBenefitSystem(TaxBenefitSystem):

	# The construction of the TBS is done in its constructor.
    def __init__(self):

    	# I first need to call the parent constructor for a basic initialization, providing the entities in a simple list.
        entities = [Familles, Individus]
        TaxBenefitSystem.__init__(self, entities)

        self.Scenario = Scenario

        # I add a variable to the TBS
        self.add_variable(age_en_mois)

        # I add several variables at the same time.
        self.add_variables(birth, salaire_brut)

        # I add all the ones declared in a python file
        self.add_variables_from_file("/path/to/openfisca_variables.py")

        # I recursively browse a directory to add all the variables declared in it.
        self.add_variables_from_directory("other/path/to/openfisca/variables/")

        # I declare the main parameter file
        path_to_root_params = 'path/to/root_params.xml'
        self.add_legislation_params(path_to_root_params)

        # I add another parameter file
		path_to_crds_params = 'path/to/crds_params.xml'
		# It will be added to the tree into the node `csg.activite`
        self.add_legislation_params(path_to_crds_params, 'csg.activite')
```

benjello:
Very nice work ! Congratulations !
Everything is simpler and clearer. 
Did you work on the reform (parametric, structural) par too ?

fpagnoux:
#### Creating a reform

A reform is a usually small set of modifications applied to a tax benefit systems (i.e. to its parameters or variables), in order to study the impact of legislation modification.

The way to build them changes significantly with this PR:

Example:
```py
# I declare a new variable

class charge_loyer(Variable):
    column = columns.FloatCol
    entity_class = entities.FoyersFiscaux
    label = u"Charge déductible pour paiement d'un loyer"

        return period, (...)

# I declare another one aiming at replacing an existing variable

class charges_deduc(Variable):
    label = u"Charge déductibles intégrant la charge pour loyer (Trannoy-Wasmer)"
    reference = "charges_deduc"

        return period, (...)

# I declare a legislation modifier. The API hasn't changed and is still rough to handle.
def modify_legislation_json(reformed_legislation):
    reform_legislation_subtree = {
        "@type": "Node",
        "description": "Charge de loyer",
        "children": {...},
        }
    reformed_legislation['children']['charge_loyer'] = reform_legislation_subtree
    return reformed_legislation


# I can finally declare my reform
class trannoy_wasmer(Reform):
    name = u'Loyer comme charge déductible (Trannoy-Wasmer)'

    # I need to implement this method.
    # Its role is to transform the reference tax benefit system into the reformed one.
    # When this method is called, self is a copy of the reference tax benefit system.
    def apply(self):

        # I add a new variable
        self.add_variable(charge_loyer)

        # I replace an existing one
        self.update_variable(charges_deduc)

        # I modify the legislation
        self.modify_legislation_json(modifier_function = modify_legislation_json)

        # I can also neutralize a column
        self.neutralize_column('rsa')

# I can then use my reform

tbs = SomeTaxBenefitSystem()

# I can my slightly different tax benefit system
reformed_tbs = trannoy_wasmer(tbs)

```

benjello:
Looks very good, indeed.
Did you take car of the reference attribute etc so the simulation.calculate can handle both the reform and the original tbs ?
Sorry for dropping question without testing but i do not have much time right now but i will definitively
try your PR my tbs and reform.

fpagnoux:
Some code has been removed, as was either already said to be deprecated, or didn't seem to be used:
- `XmlBasedTaxBenefitSystem`
- `update_legislation` and `updated_legislation_items` in reforms.

fpagnoux:
@benjello  yep as you can see [here](https://github.com/openfisca/openfisca-core/blob/remove-metaclasses/openfisca_core/tests/test_reforms.py#L19) you can still use `reference = True` to simulate without taking the reform into account.

fpagnoux:
No worries this has big impacts on the API so all questions and feedback are really welcome so that we end up with something satisfying for everyone ;)

benjello:
If update_legislation and updated_legislation_items were not replaced I may still use them ...
I will give you feedback ASAP.

fpagnoux:
#### Creating an extension:

An extension is a set of variables and parameters that can be added to a tax benefit system, but which purpose is too specific to be added to the general shared code base of this tax benefit system (e.g. Paris social prestations in `openfisca-france`)

An extension is loaded this way:

```py
tbs = SomeTaxBenefitSystem()
tbs.load_extension('path/to/extensions/folder')
```

The extension folder must have a specific architecture:
```
    extension_folder/parameters.xml # Optional parameters file. Must be called `parameters.xml
    extension_folder/{some_formula}.py # File containing formulas
    extension_folder/{other_formula}.py
    extension_folder/{some_formula}.yaml # Optional test files
    extension_folder/{other_formula}.yaml
```

laem:
:tada: 
>  +* XmlBasedTaxBenefitSystem is deprecated, and MultipleXmlBasedTaxBenefitSystem renamed to TaxBenefitSystem 

:+1: 
I'll have to adapt #387 following the same principles. 

> I declare a legislation modifier. The API hasn't changed and is still rough to handle.
def modify_legislation_json(reformed_legislation):

This should be simplified by YAML params.

cbenz:
Wow it's a good design, I especially like the `add_variables_from_file` / `add_variables_from_directory` methods.

Some quick questions (didn't dive into the code yet):
- do the metaclasses still return a `Column` object?
- or did you just disable the "auto-registering" of the variables in the global tax_benefit_system instance?

openfisca_core/scenarios.py
```diff
@@ -185,7 +185,7 @@ def fill_simulation(self, simulation, use_set_input_hooks = True, variables_name
                     # All parallel axes have the same count and entity.
                     first_axis = parallel_axes[0]
                     axis_count = first_axis['count']
-                    axis_entity = simulation.entity_by_column_name[first_axis['name']]
+                    axis_entity = simulation.getVariableEntity(first_axis['name'])
```
cbenz: In Python the convention would be camel case: `get_variable_entity` :)

cbenz:
Also, very good to get rid of `Simulation.entity_by_column_name` and `Entity.column_by_name`

openfisca_core/taxbenefitsystems.py
```diff
@@ -102,42 +98,115 @@ def new_scenario(self):
...
+    def add_variables(self, *variables):
+        for variable in variables:
+            self.add_variable(variable)
+
+    def load_extension(self, path):
+        if path.endswith('.git'):
+            import tempfile
+            import subprocess
+            temp_dir = tempfile.mkdtemp()
+            clone_command = ' '.join(['git clone', path, temp_dir, '&> /dev/null'])
```
cbenz: Waw that's perhaps not the right place :)

I guess the usage is:

```python
tbs.load_extension('https://github.com/xxx/yyy.git')
```

I'd like to decouple the `git clone` from the loading of the extension.

We can do:

```bash
$ pip install git+https://github.com/xxx/yyy.git
```

then

```python
tbs.load_extension_from_package('yyy')
```

where `yyy` is a Python package name.

fpagnoux:
@benjello I don't have a strong opinion about `update_legislation` and `updated_legislation_items`. They were marked as deprecated so I thought they were not used anymore. If you want, let me know and I'll put them back, as anyway @laem will probably offer new helpers when he is done working on the parameters.

openfisca_core/scenarios.py
```diff
@@ -185,7 +185,7 @@ def fill_simulation(self, simulation, use_set_input_hooks = True, variables_name
                     # All parallel axes have the same count and entity.
                     first_axis = parallel_axes[0]
                     axis_count = first_axis['count']
-                    axis_entity = simulation.getVariableEntity(first_axis['name'])
```
fpagnoux: @cbenz thanks for the quick fix.

openfisca_core/taxbenefitsystems.py
fpagnoux: Yay I admit this may be a little weird. I thought I would need it, but actually I'm not using it, so I'll just remove the feature. Adding extensions from a package seems interesting too, I'll keep that in mind.

openfisca_core/taxbenefitsystems.py
```diff
@@ -102,42 +98,115 @@ def new_scenario(self):
...
+        column = variable.to_column(self)
+        self.column_by_name[column.name] = column
+
+        return column
+
+    def add_variable(self, variable_class):
+        return self.load_variable(variable_class, update = False)
+
+    def update_variable(self, variable_class):
+        return self.load_variable(variable_class, update = True)
```
fpagnoux: Note : We should check if there is a reference attribute given with the new variable. If so, we should replace the indicated variable and not be based on the name.

openfisca_core/taxbenefitsystems.py
```diff
@@ -29,22 +27,26 @@ class AbstractTaxBenefitSystem(object):
...
+        self.column_by_name = collections.OrderedDict()
+        self.automatically_loaded_variable = set()
+        self.legislation_xml_info_list = []
+        self._legislation_json = legislation_json
+        self.set_entities(entities)
+
+    def set_entities(self, entities):
+        if entities is None:
+            from entities import BasicPersonEntity
+            entities = [BasicPersonEntity]
```
fpagnoux: Note: this should be tested

fpagnoux:
This seems GTM to me. @benjello what about `update_legislation` and `updated_legislation_items` N

setup.cfg
```diff
@@ -29,6 +29,6 @@ previous = true
 hang-closing = true
 ; E128 continuation line under-indented for visual indent
 ; E251 unexpected spaces around keyword / parameter equals
-ignore = E128,E251
+ignore = E128,E251,F405
```
laem: Is there a description of F405 to add before ?

benjello: F405 name may be undefined, or defined from star imports: module

PR #391, opened 2016-06-30
Fake PR

fpagnoux:
Just to make sure the test fails when there is no changelog update or no version update/


fpagnoux:
Behaves as expected.

PR #394, opened 2016-07-12
Add assertion in get_variable_entity to detect absent variables

benjello:


cbenz:
Hmm, what's the reason for this check?

I'm not convinced: I would prefer [`TaxBenefitSystem.get_column`](https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/taxbenefitsystems.py#L160-L161) to check that the column exists, not the caller.

I suggest:

```python
class ColumnNotFound(Exception):
    pass

def get_column(self, column_name):
    column = self.column_by_name.get(column_name)
    if column is None:
        raise ColumnNotFound(u'Column "{}" not found'.format(column_name))
    return column
```

@fpagnoux Do you agree? Would it break any callers?

benjello:
I need to see then name of column which is not found when debugging.

fpagnoux:
It would break several callers, as we use get_column() to know if a column exists. 

@benjello, can you tell us with more details what the problem is ? Under which circonstances is a column not found ?

fpagnoux:
I think I understand the problem:
```py
simulation.calculate('lol', period)
*** AttributeError: 'NoneType' object has no attribute 'entity_key_plural'
```

benjello:
You got it. When you are late on renaming, it tiens to a nightmare.

Le mer. 13 juil. 2016 14:16, Florian Pagnoux <notifications@github.com> a
écrit :

> I think I understand the problem:
>
> simulation.calculate('lol', period)*** AttributeError: 'NoneType' object has no attribute 'entity_key_plural'
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/openfisca/openfisca-core/pull/394#issuecomment-232337513>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ABHxdw1v85HCD_M5ZGifoqW6yD7iCXovks5qVNcfgaJpZM4JKlrB>
> .
>
-- 
Mahdi Ben Jelloul

fpagnoux:
@cbenz 
I think `simulation.get_or_new_holder` should check the columns exist.

fpagnoux:
Also `ColumnNotFound` -> `VariableNotFound`, let's progressively get rid of these `columns` ;)

fpagnoux:
See #395

benjello:
Obsolete since  #395 is merged

PR #395, opened 2016-07-15
Detect absent variables

fpagnoux:
Replaces #394

Explicit the error when a variable is not found

@benjello can you please check that this behaviour suits your need ?


benjello:
Seems good to me. Well done @fpagnoux

PR #399, opened 2016-08-04
Load extensions from pip packages

fpagnoux:
Enables this usage:

```py
tax_benefit_system.load_extension('openfisca_paris')
```

Provided that `openfisca_paris` has been installed via pip before. This is useful to define the extensions to load via the api config file. See https://github.com/openfisca/openfisca-web-api/pull/61.


cbenz:
Thanks!

PR #404, opened 2016-08-24
Update introspection data

cbenz:
This allows to enhance data served by the web API which itself feeds the Legislation Explorer.

### DIFF
--- a/openfisca_core/taxbenefitsystems.py
+++ b/openfisca_core/taxbenefitsystems.py
@@ -111,19 +111,18 @@ class TaxBenefitSystem(object):
         attributes = dict(variable_class.__dict__)
 
         existing_column = self.get_column(name)
+        if existing_column:
+            if update:
+                attributes['reference'] = existing_column
+            else:
+                # Variables that are dependencies of others (trough a conversion column)can be loaded automatically
+                if name in self.automatically_loaded_variable:
+                    self.automatically_loaded_variable.remove(name)
+                    return self.get_column(name)
+                raise VariableNameConflict(
+                    u'Variable "{}" is already defined. Use `update_variable` to replace it.'.format(name))
 
-        if existing_column and not update:
-            # Variables that are dependencies of others (trough a conversion column)can be loaded automatically
-            if name in self.automatically_loaded_variable:
-                self.automatically_loaded_variable.remove(name)
-                return self.get_column(name)
-            raise VariableNameConflict(
-                "Variable {} is already defined. Use `update_variable` to replace it.".format(name))
-
-        if existing_column and update:
-            attributes['reference'] = existing_column
-
-        # We pass the variable_class just for introspection for parsers.
+        # We pass the variable_class just for introspection.
         variable = variable_type(name, attributes, variable_class)
         # We need the tax benefit system to identify columns mentioned by conversion variables.
         column = variable.to_column(self)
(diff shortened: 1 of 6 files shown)


openfisca_core/variables.py
```diff
@@ -1,3 +1,6 @@
+# -*- coding: utf-8 -*-
```
fpagnoux: 👍

openfisca_core/taxbenefitsystems.py
```diff
@@ -137,21 +136,20 @@ def add_variable(self, variable_class):
...
-            potential_variables = [getattr(module, c) for c in dir(module) if not c.startswith('__')]
+            potential_variables = [getattr(module, item) for item in dir(module) if not item.startswith('__')]
             for pot_variable in potential_variables:
                 # We only want to get the module classes defined in this module (not imported)
-                if ((isclass(pot_variable) and
-                     issubclass(pot_variable, AbstractVariable) and
-                     pot_variable.__module__.endswith(module_name))):
+                if isclass(pot_variable) and \
+                        issubclass(pot_variable, AbstractVariable) and \
+                        pot_variable.__module__.endswith(module_name):
```
fpagnoux: From https://www.python.org/dev/peps/pep-0008/#maximum-line-length:

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

openfisca_core/variables.py
```diff
@@ -20,54 +23,39 @@ def introspect(self):
...
             entity_class = entity_class,
-            formula_class = formula_class,
+            formula_class = self.formula_class,
             reference_column = reference,
             comments = comments,
-            line_number = line_number,
+            start_line_number = start_line_number,
             source_code = source_code,
             source_file_path = source_file_path,
-            base_function = self.attributes.pop('base_function', UnboundLocalError),
```
fpagnoux: Why don't we need this anymore ?

fpagnoux:
As we discussed on the phone, and for the future, I think we should generally let time for code reviews before merging PRs.

This is **especially** true for:

- PRs on openfisca-core. We all rely on this fundamental package, and would gain for its code being high quality. We have to deal with a lot of technical debt today, and non-challenged quick additions are the best way to increase it. I'm not saying this is necessarily the case here, but we should avoid this bad practise in general.
- PRs with breaking changes (major version update). If a change may break some things for users, it is very likely to be worth a discussion with them. It would help everyone to understand the impact, and would allow to challenge the change. 
- To a lesser extend, PR thats add new functionalities (minor version update). Users can have constructive feedback about the new features usages, and implementation.

I understand we all have constraints that push us to want to go fast, and that sometimes production emergency happen. In that case, please consider releasing  your quick fix in a temporary version with pip (3.0.0.dev0 for instance). I don't think we should restrict ourselves to wait for reviews for every single PR, especially if they are about specific things that only impact our respectives scope.

However, for changes that matter, let's please make to sure reach consensus before merging.

fpagnoux:
If we all agree on this, I'd be in favour of puting the precedend post on the [contribution guidelines](http://doc.openfisca.fr/en/contribute/guidelines.html)

laem:
I would also add that we need an extensive text explanation of the change, which helps a lot for code review and discussion.

cbenz:
I agree about @fpagnoux . I am sorry for the rush I was in yesterday and won't do it again so quickly.

OK too about @laem

PR #406, opened 2016-08-25
Adapt base_functions to extra params

fpagnoux:
Should fix https://github.com/openfisca/openfisca-france/issues/523


benjello:
GTM (and urgent)

PR #412, opened 2016-10-20
Loosen versionning enforcing

fpagnoux:
When only the README (or any other file that is not _in_ the package) is modified, don't force a version update.

### DIFF
--- a/travis-run-tests.sh
+++ b/travis-run-tests.sh
@@ -1,22 +1,24 @@
 #! /usr/bin/env bash
 
-set -x
-
 current_version=`python setup.py --version`
-if [[ "$TRAVIS_BRANCH" == "master" && "$TRAVIS_PULL_REQUEST" != false ]]
-then
-    if git rev-parse $current_version
-    then
-        set +x
-        echo "Version $version already exists. Please update version number in setup.py before merging this branch into master."
-        exit 1
-    fi
 
-    if git diff-index master --quiet CHANGELOG.md
+if ! git diff-index --quiet master openfisca_core
+then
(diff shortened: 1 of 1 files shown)


fpagnoux:
@cbenz I'm not a shell expert so I couldn't fit the three conditions on one line. If this is good I'll transpose it to all other repositories.

travis-run-tests.sh
```diff
@@ -1,22 +1,24 @@
 #! /usr/bin/env bash
 
-set -x
-
 current_version=`python setup.py --version`
-if [[ "$TRAVIS_BRANCH" == "master" && "$TRAVIS_PULL_REQUEST" != false ]]
+
+if ! git diff-index --quiet master openfisca*/**
```
cbenz: I think this line is sufficient:

```
git diff-index master openfisca_core
```

because:

```
~/D/o/openfisca-core (master) $ echo a >> openfisca_core/entities.py 
~/D/o/openfisca-core (master) $ echo a >> openfisca_core/scripts/json_with_test_case_to_input_variables.py 
~/D/o/openfisca-core (master) $ git diff-index master openfisca_core
:100644 100644 9f3ea7f0ae76e7a8d728c0ddc80b383ca10ca948 0000000000000000000000000000000000000000 M	openfisca_core/entities.py
:100755 100755 b34f6d077c0b2dff7f4e551ae0a0a0a6cfc091d3 0000000000000000000000000000000000000000 M	openfisca_core/scripts/json_with_test_case_to_input_variables.py

```

cbenz: Also be careful with the `'!'`, I'm not sure it means `== 0` in Bash, or else it is a construct I don't know.

PR #422, opened 2016-11-30
Fix bug in entity.sum

fpagnoux:


fpagnoux:
@benjello this seems to solve the problem you mentionned, can you try with this branch ?
Tests still don't pass, but this doesn't seem to be related.

PR #425, opened 2017-01-04
Make Entity provide its own JSON representation

cbenz:
Precedes https://github.com/openfisca/openfisca-web-api/pull/73

### DIFF
--- a/openfisca_core/entities.py
+++ b/openfisca_core/entities.py
@@ -23,6 +23,16 @@ class Entity(object):
             raise Exception("Entity {} has no attribute {}".format(self.key, attribute))
         return projector
 
+    @classmethod
+    def to_json(cls):
+        return {
+            'isPersonsEntity': cls.is_person,
+            'key': cls.key,
+            'label': cls.label,
+            'plural': cls.plural,
+            'roles': cls.roles_description,
+            }
+
     # Calculations
 
     def check_variable_defined_for_entity(self, variable_name):
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ from setuptools import setup, find_packages
 
 setup(
     name = 'OpenFisca-Core',
-    version = '7.0.1',
+    version = '7.1.0',
     author = 'OpenFisca Team',
     author_email = 'contact@openfisca.fr',
     classifiers = [
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 7.1.0
+
+* Add `Entity.to_json` method. Used by OpenFisca-Web-API in `/entity` endpoint in particular.
+
 ## 7.0.1
 
 * Declare `Openfisca-Parsers` as an optional dependency of `Openfisca-Core`.


openfisca_core/entities.py
```diff
@@ -53,6 +53,23 @@ def __call__(self, variable_name, period = None, options = [], **parameters):
         else:
             return self.simulation.calculate(variable_name, period, **parameters)
 
+    # JSON
+
+    @classmethod
+    def to_json(cls):
```
fpagnoux: @MattiSG: 
Some architectural question : should it be a core's class concern to convert itself to JSON so that it's served by the API ?

**Cons :**
- We put code in the core that is not related to generic operations, but a specific usecase (serving the API)

**Pros:**
- If we put this code in the API, we'll have to get many of the class attributes (here `plural`, `label`, `key`) instead of just calling the `to_json()` method. The API code will be somewhat more coupled with the structure of the class.

I'm more leaning towards putting this code in the API. @cbenz is there another part of OpenFisca, out of the API, that needs to convert entities to JSON ?

MattiSG: It's a common pattern that `to_string` and `to_json` methods are part of the model itself. It indeed is the one with most knowledge of its state and how to serialise it. Encapsulation trumps [single-responsibility](https://en.wikipedia.org/wiki/Single_responsibility_principle) in that case, as a matter of [cohesion](https://en.wikipedia.org/wiki/Cohesion_(computer_science)).

openfisca_core/entities.py
```diff
@@ -53,6 +53,23 @@ def __call__(self, variable_name, period = None, options = [], **parameters):
...
+        entity_dict = {
+            'key': cls.key,
+            'label': cls.label,
+            'plural': cls.plural,
+            }
+        if cls.is_person:
+            entity_dict['isPersonsEntity'] = cls.is_person
```
MattiSG: Why add a variable boolean within a disjunction? Either always pass the boolean or add a fixed value if `cls.is_person` is true.

MattiSG: This name seems very weak. Why is there a specific boolean entry rather than a `type` property, for example?

setup.py
```diff
@@ -7,7 +7,7 @@
 
 setup(
     name = 'OpenFisca-Core',
-    version = '4.1.5',
+    version = '4.1.6',
```
MattiSG: Adding a new public API method is not a patch bump, but a minor bump. This method did not exist beforehand, and is thus backward-incompatible.
See [SemVer](http://semver.org).

openfisca_core/entities.py
```diff
@@ -53,6 +53,23 @@ def __call__(self, variable_name, period = None, options = [], **parameters):
         else:
             return self.simulation.calculate(variable_name, period, **parameters)
 
+    # JSON
```
MattiSG: What is the aim of this comment? If it only says `JSON` on a method called `to_json`, better remove it altogether.

cbenz: This is a pattern often found in the codebase to classify the methods of a class (see [here](https://github.com/openfisca/openfisca-core/blob/f9fa2813b3790115ee8663a31eb939931eb264b9/openfisca_core/entities.py#L26) or [there](https://github.com/openfisca/openfisca-core/blob/f9fa2813b3790115ee8663a31eb939931eb264b9/openfisca_core/entities.py#L87) and [there](https://github.com/openfisca/openfisca-core/blob/f9fa2813b3790115ee8663a31eb939931eb264b9/openfisca_core/entities.py#L73).

Sure, in this case it's abusive! I'll remove it.

openfisca_core/entities.py
cbenz: This is a common pattern, for `to_json` methods, to export only the fields which make sense. As a consequence, they avoid `key: null` or, in this case, `key: false`.

I agree that it would be better to export a `type` key whose value would be either `person` either `group`.

I kept the Web API function as is (see [here](https://github.com/openfisca/openfisca-web-api/blob/d790243123c72fd198cc29a63d6dd4dedb316510/openfisca_web_api/controllers/entities.py#L12-L20)) but I'm OK to enhance it *via* this PR.

setup.py
cbenz: OK

cbenz:
@MattiSG 

> Could you please link to the matching PR in web-api that removes the code there?

https://github.com/openfisca/openfisca-web-api/pull/75

openfisca_core/entities.py
MattiSG: Unless needed for backward compat, I'm in favour of enhancing the representation  😃

MattiSG:
> openfisca/openfisca-web-api#75

This states “Cancels openfisca/openfisca-core#425”… I thought they were both working together as one? 😥

> There is a way to keep the encapsulation and improve readability

In my OOP book, model representation transformations should be instance methods managed by the class, and “hydration” methods (i.e. building an object from flat data) should be constructors. The API should more be concerned with validating entries, logging, and serving stuff fast than about parsing what it receives.

Basically, if changing the model means you also necessarily need to change the API (and vice-versa), you have such a strong coupling that the module split doesn't make sense.

openfisca_core/entities.py
fpagnoux: 👌

fpagnoux:
Ok, let's keep the `to_json` method in the core, and move on with this PR, thanks @MattiSG !

openfisca_core/entities.py
cbenz: I agree but we may not re-design the API in this PR. It would be a breaking change. Especially given we plan to refactor the web API.

PR #426, opened 2017-01-04
Combine tax scale less verbose

benjello:


benjello:
You do not use this but I assigned you a review for sanity check.

fpagnoux:
I don't understand what this does, and I don't think there is any documentation on the tax scale. 

What is the goal of the PR ? Could you give more explanation ?

benjello:
`combine_tax_scales` combines many tax scales into a unique one that can be used to inverse some revenus as done in a reform. I use the reform extensively with survey data. 

This PR limits the verbosity of `combine_tax_scales` that warns the user when provided with a list (a legislation node) of supposed tax_scales do not contain only tax scales. Since now  the legislation node do contain instant and a name, it gets too much verbose so I want to silence these warnings

PR #427, opened 2017-01-09
Fix occasionnal `NaN` creation in `MarginalRateTaxScale.calc`

### LINKED ISSUE #410: bareme.calc returns NaN sometimes
Hi @benjello, following https://github.com/openfisca/openfisca-france/pull/573 I've found that `apply_bareme_for_relevant_type_sal` in `openfisca_france/model/prelevements_obligatoires/prelevements_sociaux/cotisations_sociales/base.py` calls `iter_cotisations` which calls `openfisca_core.taxscales.MarginalRateTaxScale.calc`.

This last function returns sometimes the `NaN`, in particular [this line](https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/taxscales.py#L198)

Should we add a check here? or is the problem upstream?

Here is the context:

``` python
     18     def iter_cotisations():
     19         for type_sal_name, type_sal_index in CAT:
     20             if type_sal_name not in bareme_by_type_sal_name:  # to deal with public_titulaire_militaire
     21                 continue
     22             bareme = bareme_by_type_sal_name[type_sal_name].get(bareme_name)  # TODO; should have better warnings
     23             if bareme is not None:
     24                 yield bareme.calc(
     25                     base * (categorie_salarie == type_sal_index),
     26                     factor = plafond_securite_sociale,
---> 27                     round_base_decimals = round_base_decimals,
     28                     )
     29     return - sum(iter_cotisations())

ipdb> plafond_securite_sociale
array([ 0.], dtype=float32)
```

``` python
    194     def calc(self, base, factor = 1, round_base_decimals = None):
    195         base1 = np.tile(base, (len(self.thresholds), 1)).T
    196         if isinstance(factor, (float, int)):
    197             factor = np.ones(len(base)) * factor
    198         thresholds1 = np.outer(factor, np.array(self.thresholds + [np.inf]))
    199         if round_base_decimals is not None:
    200             thresholds1 = np.round(thresholds1, round_base_decimals)
    201         a = max_(min_(base1, thresholds1[:, 1:]) - thresholds1[:, :-1], 0)
--> 202         if round_base_decimals is None:
    203             return np.dot(self.rates, a.T)
    204         else:
    205             r = np.tile(self.rates, (len(base), 1))
    206             b = np.round(a, round_base_decimals)
    207             return np.round(r * b, round_base_decimals).sum(axis = 1)
    208 

ipdb> base
array([ 0.], dtype=float32)
ipdb> factor
array([ 0.], dtype=float32)
ipdb> round_base_decimals
2
```

benjello:
Fixes #410


benjello:
@cbenz : ca you test this ? So I know if I can definitively merge.

openfisca_core/taxscales.py
```diff
@@ -195,7 +195,9 @@ def calc(self, base, factor = 1, round_base_decimals = None):
         base1 = np.tile(base, (len(self.thresholds), 1)).T
         if isinstance(factor, (float, int)):
             factor = np.ones(len(base)) * factor
-        thresholds1 = np.outer(factor, np.array(self.thresholds + [np.inf]))
+        # thresholds1 = np.outer(factor, np.array(self.thresholds + [np.inf]))
+        # changed to below to avoind NaN creation
```
cbenz: Can you remove those old lines?

openfisca_core/taxscales.py
```diff
@@ -195,7 +195,9 @@ def calc(self, base, factor = 1, round_base_decimals = None):
...
+        thresholds1 = np.outer(factor + np.finfo(np.float).eps, np.array(self.thresholds + [np.inf]))
```
cbenz: That's high-level NumPy usage!

I'm afraid I don't understand the technique here, and if I had to touch this part of the code I'd be embarrassed.

You're here the NumPy expert :)

openfisca_core/tests/test_tax_scales.py
```diff
@@ -25,7 +25,7 @@ def test_linear_average_rate_tax_scale():
     marginal_tax_scale.add_bracket(0, 0)
     marginal_tax_scale.add_bracket(1, 0.1)
     marginal_tax_scale.add_bracket(2, 0.2)
-    assert_near(marginal_tax_scale.calc(base), [0, .05, .1, .2], absolute_error_margin = 0)
+    assert_near(marginal_tax_scale.calc(base), [0, .05, .1, .2], absolute_error_margin = 1e-16)
```
cbenz: Does `assert_near` work here without the `absolute_error_margin` kwarg?

benjello: I don't think so.

openfisca_core/taxscales.py
benjello: If I recall well we wanted to avoid a 0 * inf; So I use the epsilon.
```
np.finfo(np.float).eps * np.inf
 ->  inf

0 * np.inf
 ->  nan
```

PR #438, opened 2017-01-26
Create class PeriodSizeIndependentFloatCol

benjello:
I need this in the taxipp branch of openfisca-france

### DIFF
--- a/openfisca_core/columns.py
+++ b/openfisca_core/columns.py
@@ -485,5 +485,9 @@ class EnumCol(IntCol):
         return value
 
 
+class PeriodSizeIndependentFloatCol(FloatCol):
+    is_period_size_independent = True
+
+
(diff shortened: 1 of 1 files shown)


fpagnoux:
Could we have more info about why this is needed ? It seems like we are building really complex types, which doesn't seem great to me.

benjello:
It is suitable for base_ressources type variable.
I understand that we need to clarify the types and calculate options but right now I need this.
You can leave it open, and we will come back to it when we have time (i do not have much time right now) to think hard on the multi-period computation and natural frequency of variables.

MattiSG:
In order to increase the readability of the current active work in progress, I close this PR since there is no active development on it. You are very welcome to reopen it with the details needed for the reviewers when you have more time 🙂

PR #440, opened 2017-01-30
Fix to_json for dated variables with extra params

fpagnoux:
There is a bug when computing intermediate variables with the API.

`DatedVariables.to_value_json ` method was not compatible with extra parameter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 4.3.2
+
+* Fix `to_value_json` for `DatedVariable` with extra parameters.
+
+This was causing a crash when calculating intermediate variables with the API.
+
+Unlike simple formulas, a `DatedVariable` have several functions. We thus need to select the right one according to the period before doing parameters introspection.
+
 ## 4.3.1
 
 * Fix `set_input` and `default` setting in `new_filled_column`
--- a/openfisca_core/holders.py
+++ b/openfisca_core/holders.py
@@ -360,17 +360,27 @@ class Holder(object):
     def get_from_cache(self, period, extra_params = None):
         return self if self.column.is_permanent else DatedHolder(self, period, extra_params)
 
-    def get_extra_param_names(self):
-        return self.formula.function.__func__.func_code.co_varnames[3:]
+    def get_extra_param_names(self, period):
+        from .formulas import DatedFormula
+        if isinstance(self.formula, DatedFormula):
+            # Get the function that matches the period
+            formula = [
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,8 +1,14 @@
...
+## 4.2.2
+
+* Fix `to_value_json` for `DatedVariable` with extra parameters.
+
+This was causing a crash when calculating intermediate variables with the API.
+
 ## 4.2.1
 
-  Fix permanent and period size independent variables neutralization
+* Fix permanent and period size independent variables neutralization
```
cbenz: Not related to this PR ;-) That's fun, today I was said the same thing ^^

openfisca_core/holders.py
```diff
@@ -360,17 +360,27 @@ def put_in_cache(self, value, period, extra_params = None):
     def get_from_cache(self, period, extra_params = None):
         return self if self.column.is_permanent else DatedHolder(self, period, extra_params)
 
-    def get_extra_param_names(self):
-        return self.formula.function.__func__.func_code.co_varnames[3:]
+    def get_extra_param_names(self, period = None):
```
cbenz: Why set a default value of `None` for `period`? Why not simply force the parameter to exist? This would be more predictable.

fpagnoux:
>Just wondering from a high-level point of view, why a DatedVariable would be a different case.
>
>I guess that's because DatedVariables have many formulas functions? And we select the one corresponding to the period of the computation?

Yes, exactly. We use introspection for extra parameters name, and the way to access the function differs whether it is a simple `Variable` (only one function) or a `DatedVariable` (several functions).

CHANGELOG.md
fpagnoux: I think this is the 3rd time I fix this in different PRs ;)

openfisca_core/holders.py
fpagnoux: :ok

CHANGELOG.md
cbenz: But it works with both `-` and `*`, so we will always use the wrong one... Why not accept both?

GitHub [mastering Markdown](https://guides.github.com/features/mastering-markdown/) howto shows we can use both (in examples/list section).

On the other hand, having both creates an impression of low quality :-)

@MattiSG Any thoughts?

fpagnoux: There was not _any_ bullet before, the `-` you see is just the diff mark.

But anyway, I think in general we should stay consistent.

PR #444, opened 2017-02-02
Internationalize scenarios

fpagnoux:
Connected to openfisca/openfisca-web-api#76

A lot of the logic to build an openfisca `scenario` from a `json` is duplicated [in](https://github.com/openfisca/openfisca-france/blob/13.0.0/openfisca_france/scenarios.py#L82) [all](https://github.com/openfisca/openfisca-tunisia/blob/master/openfisca_tunisia/scenarios.py#L72) [countries](https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/tests/dummy_country/scenarios.py#L52). 

This is mainly used to serve the tax and benefits system with `web-api`, but it has other applications.

Inside this duplicated code, we can find... a lot of duplication, as the entities are usually hardcoded, and similar logic is repeated for `Familles`, `FoyerFiscal`...

This PR generalise a big chunk of the code, and moves it from `france` to `core`.

To be honest, yes, this is still ugly and full of debt. The `scenarios` need a huge refactoring, which would probably be done the best by throwing everything away.

However, we have no visibility about when this refactoring will come, and having a turnkey working api will be useful for `tunisia`.


fpagnoux:
@cbenz It would be nice if you could have a look at this during the upcoming week, so that we can move forward when I come back on the 13th ;)

openfisca_core/json_to_test.py
```diff
@@ -0,0 +1,150 @@
+# -*- coding: utf-8 -*-
```
cbenz: If I understand correctly, you chose to extract *biryani converters* to a dedicated module, which is a good choice! This lightens the code of the Core.

But I'm not 100% satisfied with the name `json_to_test.py`. `test` makes me think about a unit test, here it is a test case.

fpagnoux: 👍

fpagnoux: You're right, this was confusing. 

I indeed chose to separate some logic specific to turning a JSON object into a test case in a separated flag to clear the `scenario` code.

This is renamed, tell me if it's ok for you.

benjello:
@cbenz : did you try it on openfisca-senegal ? openfisca-tunisia ?

fpagnoux:
I tried in with `openfisca-senegal`. Not with `tunisia`.

fpagnoux:
It works with `senegal`.
I can help you migrate `tunisia`: the main taks is to distinguish in [scenarios](https://github.com/openfisca/openfisca-tunisia/blob/master/openfisca_tunisia/scenarios.py) which part of the JSON processing are specific to Tunisia and which part are commons.

PR #447, opened 2017-02-11
Extend run_test to read country.country_tax_benefit_system too

benjello:
As user of a derived country package like [this one](https://github.com/openfisca/openfisca-france-data/blob/taxipp/openfisca_france_data/__init__.py#L172-L173) (`taxipp` branch), i need to be able to run tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 4.3.5
+
+* Refactor decomposition TaxBenefitSystem attributes. Reform inherit the decomposition_file_path from the reference TaxBenefitSystem.
+  This does not require changing anything from the caller, which should use the `decompositions.get_decomposition_json` function instead of those attributes.
+
 ## 4.3.4
 
 * Fix occasionnal `NaN` creation in `MarginalRateTaxScale.calc` resulting from `0 * np.inf`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 4.3.4
+
+* Fix occasionnal `NaN` creation in `MarginalRateTaxScale.calc` resulting from `0 * np.inf`
+
 ## 4.3.3
 
 * Use the actual TaxBenefitSystem and not its reference when neutralizing a column.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## 4.3.6
 
-* Extend `run_test` to use `CountryTaxBenefiSystem` _or_ `country_tax_benefit_system` from country openfisca package
+* Extend `run_test` to use `CountryTaxBenefiSystem` _or_ `country_tax_benefit_system` from country openfisca package.
+This is useful when your country pakcage doesn't provide a `CountryTaxBenefiSystem` class but a `country_tax_benefit_system` instance.
 
 ## 4.3.5
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 4.3.6
+
+* Extend `run_test` to use `CountryTaxBenefiSystem` _or_ `country_tax_benefit_system` from country openfisca package
+
 ## 4.3.5
 
 * Refactor decomposition TaxBenefitSystem attributes. Reform inherit the decomposition_file_path from the reference TaxBenefitSystem.
--- a/openfisca_core/scripts/run_test.py
+++ b/openfisca_core/scripts/run_test.py
@@ -53,7 +53,7 @@ def main():
         tax_benefit_system = country_package.country_tax_benefit_system
     else:
         print('ERROR: Cannot find CountryTaxBenefitSystem nor country_tax_benefit_system in `{}` __init__ file.\n\
-            On of them should be present for this package to be a valid Openfisca country package.'.format(args.country_package))
+            One of them should be present for this package to be a valid OpenFisca country package.'.format(args.country_package))
         sys.exit(1)
 
     if args.extensions:
--- a/openfisca_core/scripts/run_test.py
+++ b/openfisca_core/scripts/run_test.py
@@ -47,7 +47,13 @@ def main():
         if len(installed_country_packages) > 1:
             print('WARNING: Several country packages detected : `{}`. Using `{}` by default. To use another package, please use the --country_package option.'.format(', '.join(installed_country_packages), country_package_name))
 
-    tax_benefit_system = country_package.CountryTaxBenefitSystem()
+    try:
+        tax_benefit_system = country_package.CountryTaxBenefitSystem()
+    except AttributeError:
+        print('{}_tax_benefit_system'.format(country_package.__name__[10:]))
+        tax_benefit_system = getattr(
+            country_package, '{}_tax_benefit_system'.format(country_package.__name__[10:])
+            )
 
     if args.extensions:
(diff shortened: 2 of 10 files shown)


openfisca_core/scripts/run_test.py
```diff
@@ -47,7 +47,10 @@ def main():
         if len(installed_country_packages) > 1:
             print('WARNING: Several country packages detected : `{}`. Using `{}` by default. To use another package, please use the --country_package option.'.format(', '.join(installed_country_packages), country_package_name))
 
-    tax_benefit_system = country_package.CountryTaxBenefitSystem()
+    try:
```
cbenz: That's not a good practice to rely on exception when there can be a check like:

```python
if country_package.CountryTaxBenefitSystem
# or, like in ` openfisca_core/tools/__init__.py `
if hasattr(module, 'CountryTaxBenefitSystem'):
```

openfisca_core/scripts/run_test.py
```diff
@@ -47,7 +47,10 @@ def main():
...
+        tax_benefit_system = country_package.CountryTaxBenefitSystem()
+    except AttributeError:
+        tax_benefit_system = country_package.country_tax_benefit_system
```
cbenz: What is `country_package.country_tax_benefit_system`? I don't see it in [`senegal_taxbenefitsystem.py`](https://github.com/openfisca/senegal/blob/master/openfisca_senegal/senegal_taxbenefitsystem.py) nor [`france_taxbenefitsystem.py`](https://github.com/openfisca/openfisca-france/blob/master/openfisca_france/france_taxbenefitsystem.py)

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 4.3.6
+
+* Extend `run_test` to use `CountryTaxBenefiSystem` _or_ `country_tax_benefit_system` from country openfisca package
```
cbenz: I don't understand the purpose of this change: in which situation is `country_tax_benefit_system` necessary?

Can you add something in the changelog line? :)

openfisca_core/scripts/run_test.py
cbenz: > As user of a derived country package like this one, i need to be able to run tests.

Sorry! I didn't see it.

CHANGELOG.md
benjello: See the motivation of the PR.
Sometimes your country_tax_benefit_system is not an instance of a CountryTaxBenefitSystem but a reform

cbenz: > As user of a derived country package like this one, i need to be able to run tests.

Can you mention france-data in the changelog?

openfisca_core/scripts/run_test.py
cbenz: Tell me if I'm wrong but I don't see `country_tax_benefit_system` in https://github.com/openfisca/openfisca-france-data/blob/master/openfisca_france_data/__init__.py, but `france_data_tax_benefit_system` 🤔

openfisca_core/scripts/run_test.py
cbenz: I think the same check than openfisca_core/tools/__init__.py should be fine (?)

openfisca_core/scripts/run_test.py
```diff
@@ -47,10 +47,14 @@ def main():
...
 
-    try:
+    if hasattr(country_package, 'CountryTaxBenefitSystem'):
         tax_benefit_system = country_package.CountryTaxBenefitSystem()
-    except AttributeError:
+    elif hasattr(country_package, 'country_tax_benefit_system'):
         tax_benefit_system = country_package.country_tax_benefit_system
+    else:
+        print('ERROR: Cannot find CountryTaxBenefitSystem nor country_tax_benefit_system in `{}` __init__ file.\n\
+            On of them should be present for this package to be a valid Openfisca country package.'.format(args.country_package))
```
cbenz: Typos: s/on/one, s/Openfisca/OpenFisca :)

openfisca_core/scripts/run_test.py
```diff
@@ -47,10 +47,14 @@ def main():
...
             print('WARNING: Several country packages detected : `{}`. Using `{}` by default. To use another package, please use the --country_package option.'.format(', '.join(installed_country_packages), country_package_name))
 
-    try:
+    if hasattr(country_package, 'CountryTaxBenefitSystem'):
         tax_benefit_system = country_package.CountryTaxBenefitSystem()
-    except AttributeError:
+    elif hasattr(country_package, 'country_tax_benefit_system'):
         tax_benefit_system = country_package.country_tax_benefit_system
+    else:
+        print('ERROR: Cannot find CountryTaxBenefitSystem nor country_tax_benefit_system in `{}` __init__ file.\n\
```
cbenz: In Python 2, should be `print(u'xxx'.encode('utf-8'))` but since it's not the case everywhere let's keep this PR focused on only your changes.

fpagnoux:
The second options would be something like:

```sh
openfisca-run-test -c openfisca_france -r openfisca_france_data.openfisca_france_data
```

cbenz:
@fpagnoux To be clear what's the point with `Reform` vs `TaxBenefitSystem`? A `Reform` is an instance whereas a `TaxBenefitSystem` is a class?

We could generalize by defining the interface to be a function with no arguments.

Moreover I would vote to remove the camel-case of the class, because it could be a function.

@benjello you could do in france-data:
```python
country_tax_benefit_system = lambda: CountryTaxBenefitSystem()
```

fpagnoux:
The first one, maybe more complex, would be to replace the reform by something like :

```py
class FranceDataTaxBenefitSystem(FranceTaxBenefitSystem):
    def __init__(self):
        FranceTaxBenefitSystem.__init__(self)
        for variable in variables:
            if variable == Variable:
                continue
            try:
                self.add_variable(variable)
            except AttributeError:
                self.update_variable(variable)
```

fpagnoux:
>@fpagnoux To be clear what's the point with Reform vs TaxBenefitSystem? A Reform is an instance whereas a TaxBenefitSystem is a class?

Not even. A `Reform` is a particular case of a `TaxBenefitSystem` that is built by slightly modifying an existing `TaxBenefitSystem`.

In this case, they are functionally equivalent. We can consider that `france_data` is a reformed version of `france`, or just that it's another country package. Making it a reform allows to do comparative  simulations with and without the reform, but I doubt it is our usecase here.

fpagnoux:
>We could generalize by defining the interface to be a function with no arguments.
>
>Moreover I would vote to remove the camel-case of the class, because it could be a function.

I agree this seems more legit. However changing the case is a breaking change for many packages, not sure it's worth it.

cbenz:
OK, so now I'm pretty sure the fix has to be *only* on OpenFisca-France-Data side.

The interface with Core is a function with 0 arguments, which in OpenFisca-France/Senegal/Tunisia is a `TaxBenefitSystem` sub-class. In OpenFisca-France-Data the trick is to return a lambda like this:

```python
CountryTaxBenefitSystem = lambda: france_data_tax_benefit_system
```

(keeping camel-case)

The Core will be happy to execute this function. We're mocking a class.

This makes this PR unnecessary ^^

benjello:
I will implement @cbenz solution if everybody is ok, test it and come back to you soon. Leave this PR open and I will close it, if everything is ok.

benjello:
Fixed. Thanks @cbenz @fpagnoux

cbenz:
@benjello Can you delete the branch (using the button just above) after having double-checked?

PR #449, opened 2017-02-13
Fix occasionnal `NaN` creation in `MarginalRateTaxScale.calc`

fpagnoux:
Clean version of #427


cbenz:
It's a clean version concerning the git history.

PR #452, opened 2017-02-16
Simplify periods

### LINKED ISSUE #442: Simplify period management
## Observations

Looking at `openfisca-france`, it seems that most of the variables have a natural period granularity. For instance, the `rsa` is defined for a month, while `irpp` defined for a year.

## Current situation

The way this natural period is set for a given formula is today complex and confusing. For instance : 
```py
class rsa(Variable):
    [...]

    def function(famille, period):
        period = period.this_month
        [...]
        return period, result

```
The current implementation allows a formula to return a result for a different period than the one it has been called for.

For instance, if I call `rsa` for `2015`, the first line`period = period.this_month` will change `period` to `2015-01`. This period will be returned with the result. It is then the openfisca middleware that will check that the returned period is the same than the required one. In our case, it will throw an error except if a certain option has been set for the calculation.

### Problems caused
- Returning the period with the result is not natural, and often forgotten.
- The first line, which defines the natural period of the formula, is really unclear.
- The first line is repeated in almost all formulas.

## Proposition
If each variable has a natural period, then a **declarative approach** seems more simple.

### Example
Our previous formula becomes : 
```py
class rsa(Variable):
    period = MONTH # keyword to be discussed
    [...]

    def function(famille, period):
        [...]
        return result

```

### Behaviour
- If I call a variable for a period that is **not its natural one**, **an error is raised** before entering the formula.
    - For instance if I call `rsa` for `2015`, or `irpp` for `2015-01`.
- If a formula has no natural period, we do not check anything (some variables are period agnostics, for instance inputs or basis input combinations)

michelbl:
Fixes #442


openfisca_core/columns.py
```diff
@@ -32,6 +37,7 @@ class Column(object):
     formula_class = None
     is_period_size_independent = False  # When True, value of column doesn't depend from size of period (example: age)
     is_permanent = False  # When True, value of column doesn't depend from time (example: ID, birth)
+    period_behavior = None
```
cbenz: I'm still not comfortable with the term `period_behavior`.

What about keeping the same name `period_unit` like it is used in `periods.py`? I know this is less "searchable" but this *is* the same concept.

Moreover the `periods.py` should use the same `Enum` (cf my previous comment).

openfisca_core/formulas.py
```diff
@@ -28,6 +29,11 @@
 ADD = 'add'
 DIVIDE = 'divide'
 
+SET_INPUT_FILL = u'set_input_fill'
+SET_INPUT_REPLACE = u'set_input_replace'
+SET_INPUT_ADD = u'set_input_add'
```
cbenz: Idem, use an `Enum`

openfisca_core/formulas.py
```diff
@@ -79,7 +85,7 @@ def default_values(self):
     def real_formula(self):
         return self
 
-    def set_input(self, period, array):
+    def set_input(self, period, array, behavior=None):
```
cbenz: Why is this parameter unused? If that's voluntary could you add a comment ?

openfisca_core/formulas.py
```diff
@@ -925,7 +939,7 @@ def new_filled_column(__doc__ = None, __module__ = None,
     return column
 
 
-def set_input_dispatch_by_period(formula, period, array):
+def set_input_dispatch_by_period(formula, period, array, behavior=None):
```
cbenz: Idem, `behavior` param unused

openfisca_core/formulas.py
```diff
@@ -954,48 +968,84 @@ def set_input_dispatch_by_period(formula, period, array):
...
+        while sub_period.start < after_instant:
+            old_array = holder.get_array(sub_period)
+            if old_array is None:
+                holder.put_in_cache(divided_array, sub_period)
+            else:
+                holder.put_in_cache(divided_array + old_array, sub_period)
+            sub_period = sub_period.offset(1)
+
+    else:
+        raise ValueError('Unknown behavior {}'.format(behavior))
```
cbenz: Using an `Enum` would avoid having to handle invalid values.

openfisca_core/formulas.py
```diff
@@ -954,48 +968,84 @@ def set_input_dispatch_by_period(formula, period, array):
                 month = month.offset(1)
 
 
-def set_input_divide_by_period(formula, period, array):
+def set_input_divide_by_period(formula, period, array, behavior=None):
```
cbenz: Idiom: use
```python
def set_input_divide_by_period(formula, period, array, behavior=SET_INPUT_FILL):
```

openfisca_core/formulas.py
```diff
@@ -954,48 +968,84 @@ def set_input_dispatch_by_period(formula, period, array):
...
+    if formula.holder.column.period_behavior == MONTH:
+        cached_period_unit = periods.MONTH
+    elif formula.holder.column.period_behavior == YEAR:
+        cached_period_unit = periods.YEAR
+    else:
+        ValueError('set_input_divide_by_period can be used only for yearly or monthly variables.')
+
+    after_instant = period.start.offset(period_size, period_unit)
+
+    if behavior is None:
```
cbenz: Remove this `if` – cf my previous comment

openfisca_core/formulas.py
```diff
@@ -954,48 +968,84 @@ def set_input_dispatch_by_period(formula, period, array):
...
+            else:
+                holder.put_in_cache(divided_array + old_array, sub_period)
+            sub_period = sub_period.offset(1)
+
+    else:
+        raise ValueError('Unknown behavior {}'.format(behavior))
 
 
-def set_input_neutralized(formula, period, array):
+def set_input_neutralized(formula, period, array, behavior=None):
```
cbenz: Idem, unused kwarg

openfisca_core/holders.py
```diff
@@ -7,31 +7,29 @@
 
 from . import periods
 from .commons import empty_clone
+from .columns import MONTH, YEAR, PERMANENT
 
 
 class DatedHolder(object):
-    """A view of an holder, for a given period (and possibly a given set of extra parameters).
-    If the variable is not cached, it also contains the value of the variable for the given date."""
+    """The value of a variable for given a period (and possibly a given set of extra parameters).
```
cbenz: You replaced "for a given period" by "for given a period" but I think the original was the right way.

openfisca_core/base_functions.py
```diff
@@ -101,11 +101,12 @@ def last_duration_last_value(formula, simulation, period, *extra_params):
...
-                    return output_period, last_result
+                    return last_result
                 elif last_result.get(extra_params):
-                    return output_period, last_result.get(extra_params)
+                    return last_result.get(extra_params)
     if formula.function is not None:
         return formula.exec_function(simulation, period, *extra_params)
     array = holder.default_array()
-    return period, array
+    return array
```
fpagnoux: @benjello @cbenz have we ever used `last_duration_last_value` anywhere ?

fpagnoux: I think this base function is not compatible with our new rule that says that a formula always give an answer for the period it has been asked for.

If it is confirmed that this base function is not used, we should remove it in this PR.

openfisca_core/columns.py
```diff
@@ -19,6 +19,11 @@ def N_(message):
 
 year_or_month_or_day_re = re.compile(ur'(18|19|20)\d{2}(-(0?[1-9]|1[0-2])(-([0-2]?\d|3[0-1]))?)?$')
 
+# Possible values for period_behavior
+MONTH = u'month'
+YEAR = u'year'
+PERMANENT = u'permanent'
```
fpagnoux: How does an enum works ?

What happen if I do `print(QUIFAM['chef'])` ? Does it just print a number ?

openfisca_core/columns.py
fpagnoux: I'm also more leaning towards `period_unit`, which describes the concept better.

We can keep using `period_behavior` for development while we are not sure about the name and may need to do a global research and replace. But before merging, we'll have to come back to a better name.

openfisca_core/formulas.py
```diff
@@ -458,15 +454,22 @@ def compute(self, period = None, **parameters):
...
             function = self.function.im_func
             legislation = simulation.legislation_at
             if self.function.im_func.func_code.co_argcount == 2:
-                return function(entity, period)
+                result = function(entity, period)
             else:
-                return function(entity, period, legislation, *extra_params)
+                result = function(entity, period, legislation, *extra_params)
+
+        output_period, array = result
```
fpagnoux: Is this temporary code for migrating existing formulas ?

openfisca_core/formulas.py
fpagnoux: That's probably because all the `set_input***` must have the same signature, and the extra argument is only needed by `set_input_divide_by_period `.

openfisca_core/formulas.py
fpagnoux: 👍

openfisca_core/formulas.py
```diff
@@ -954,48 +965,84 @@ def set_input_dispatch_by_period(formula, period, array):
...
-                divided_array = remaining_array / months_count
-                month = period.start.period(u'month')
-                while month.start < after_instant:
-                    if holder.get_array(month) is None:
-                        holder.put_in_cache(divided_array, month)
-                    month = month.offset(1)
+        else:
+            raise ValueError('Values for variable {} are already set on period {}.'.format(holder.column.name, period))
+
+    elif behavior == SET_INPUT_REPLACE:
```
fpagnoux: Is there any use case where we use `SET_INPUT_REPLACE` ?

openfisca_core/formulas.py
fpagnoux: Even if I use an enum, as python is not strongly typed, I can still give a bad argument to the function no ?

openfisca_core/holders.py
fpagnoux: >A wrapper of the value of a variable for a given period (and possibly a given set of extra parameters).

?

openfisca_core/holders.py
```diff
@@ -112,86 +110,72 @@ def clone(self):
...
+        # Request a computation
         column_start_instant = periods.instant(column.start)
         column_stop_instant = periods.instant(column.end)
         if (column_start_instant is None or column_start_instant <= period.start) \
                 and (column_stop_instant is None or period.start <= column_stop_instant):
             formula_dated_holder = self.formula.compute(period = period, **parameters)
             assert formula_dated_holder is not None
             if not column.is_permanent:
-                assert accept_other_period or formula_dated_holder.period == period, \
+                assert formula_dated_holder.period == period, \
```
fpagnoux: Is it actually possible that this is not the case ?

It seems to me that we make sure at a lower level that `formula.compute` always return a `DatedHolder` with the right period, no ?

If so, this assertion can be kept if it helps development, but should be removed before merging.

openfisca_core/holders.py
```diff
@@ -112,86 +110,72 @@ def clone(self):
...
+        # Check that the requested period matches period_behavior
+        if self.column.period_behavior == YEAR and period.unit == periods.MONTH:
+            raise ValueError('Computation on period {} impossible on yearly variable {}'.format(
+                period, self.column.name
+                ))
+
+        if self.column.period_behavior == MONTH:
+            variable_period_unit = periods.MONTH
+        elif self.column.period_behavior == YEAR:
+            variable_period_unit = periods.YEAR
```
fpagnoux: When I read this can of code, I wonder like @cbenz if we shouldn't have by design `MONTH == periods.MONTH`

openfisca_core/scenarios.py
```diff
@@ -230,10 +222,7 @@ def fill_simulation(self, simulation, use_set_input_hooks = True, variables_name
                                 array = holder.default_array()
                             array[axis['index']:: axis_entity.step_size] = axis['min'] \
                                 + mesh.reshape(steps_count) * (axis['max'] - axis['min']) / (axis_count - 1)
-                            if use_set_input_hooks:
-                                holder.set_input(axis_period, array)
-                            else:
-                                holder.put_in_cache(array, axis_period)
+                            holder.set_input(axis_period, array, behavior=SET_INPUT_ADD)
```
fpagnoux: Introducing this extra `behavior` parameter to resolve this buggy usecase is not great, but I understand we didn't find much better.

Let's keep moving forward, but I may be worth rethinking this before merging.

fpagnoux:
Careful, you accidentally commited a submodule : 
image

openfisca_core/columns.py
michelbl: I agree

openfisca_core/formulas.py
michelbl: idem, I agree

openfisca_core/formulas.py
michelbl: I think this is intended :
```python
def f(a=None):
     return g(a)
def g(a=1):
     return a
f()
```
returns None

openfisca_core/formulas.py
michelbl: yes !

openfisca_core/formulas.py
michelbl: @cbenz mentionned the use from notebooks

openfisca_core/holders.py
michelbl: you're right : this check is done in exec_function, and both will be removed before merging

openfisca_core/holders.py
michelbl: I agree that this is a desirable long term goal, but this would require an 'infinite' period to reflect the PERMANENT behavior. Such modification of periods.py will have lots of impacts everywhere. Do you want to include that change in this PR ?

openfisca_core/columns.py
cbenz: I don't know exactly 🤔

openfisca_core/formulas.py
cbenz: If I'm not wrong, the Enum won't let an invalid value occur, as long as you use Enum.

But if you pass say a string, it won't be checked.

```python
color = Enum(['red', 'blue'])
color['red'] == 0 # OK
color['blue'] == 1 # OK
color['green'] # Exception
```

openfisca_core/formulas.py
cbenz: Think about the users of a notebook.

openfisca_core/holders.py
cbenz: @michelbl told me that, by design, there is no `PERMANENT` period in `periods.py`.

But this should be allowed (to be discussed)

openfisca_core/formulas.py
michelbl: Let's be careful : 
```
def f(a=None):
     return g(a)
def g(a=1):
     return a
f()
```
returns None

michelbl: A more 'elegant' solution : replace None by SET_INPUT_DEFAULT

MattiSG:
> Try to keep your first commit comment line under 50 characters.
[Here](https://chris.beams.io/posts/git-commit/#seven-rules) are a few other tips on what makes a great commit message  :)

> call it "Fixup : remove print".
Rather call it `fixup! remove print`, this way you can use [`--autosquash`](https://coderwall.com/p/hh-4ea/git-rebase-autosquash)  😃

openfisca_core/columns.py
michelbl:
```python
from openfisca_core.enumerations import Enum
ENUM = Enum(['a', 'b'])
print(ENUM['a'])
```
prints `0`

openfisca_core/holders.py
michelbl: Changed to the proposition of @fpagnoux

openfisca_core/base_functions.py
michelbl: Removed

openfisca_core/holders.py
michelbl: Removed

michelbl:
Reviews are welcome !

Some points that were raised and that still need to be discussed : 
* Change `period_behavior` to a better name
* Use enums or bare constants for `period_behavior` and set_input `behavior` ?
* set_input parameter `behavior` : keep it, change the implementation or find another solution ?

openfisca_core/holders.py
```diff
@@ -311,34 +232,35 @@ def graph(self, edges, get_input_variables_and_parameters, nodes, visited):
...
-    def set_input(self, period, array):
-        self.formula.set_input(period, array)
+    def set_input(self, period, array, behavior=None):
+        self.formula.set_input(period, array, behavior=behavior)
 
     def put_in_cache(self, value, period, extra_params = None):
         simulation = self.simulation
 
+        if not self.column.period_behavior == PERMANENT:
+            assert periods is not None
```
cbenz: `assert period` (and not `periods`) ;-)

openfisca_core/decompositions.py
```diff
@@ -20,16 +27,21 @@ def calculate(simulations, decomposition_json):
         else:
             node['values'] = values = []
             for simulation_index, simulation in enumerate(simulations):
+                variable_beriod_behavior = simulation.get_or_new_holder(node['code']).column.period_behavior
                 try:
-                    simulation.calculate_output(node['code'])
+                    if variable_beriod_behavior == MONTH and simulation.period.unit == u'year':
```
cbenz: Use `calculate_output`

openfisca_core/holders.py
```diff
@@ -112,164 +110,87 @@ def clone(self):
...
+                    (self.column.period_behavior == YEAR and period.unit != periods.YEAR)):
+                raise ValueError('Computation requested with wrong period unit for variable {} ({} instead of {})'.format(
+                    self.column.name,
+                    period.unit, self.column.period_behavior))
+        if period.size != 1:
+            raise ValueError('Computation requested for complex period {} for variable {}'.format(
+                period, self.column.name))
+
+        # First look for a value already cached
         dated_holder = self.get_from_cache(period, parameters.get('extra_params'))
```
cbenz: If `column` is permanent (`depcom` in dummy country), this is a holder. Name the variable: `holder`

openfisca_core/holders.py
```diff
@@ -112,164 +110,87 @@ def clone(self):
...
         The returned dated holder is always of the requested period and this method never returns None.
         """
         if period is None:
             period = self.simulation.period
         column = self.column
-        accept_other_period = parameters.get('accept_other_period', False)
 
-        # First look for dated_holders covering the whole period (without hole).
+        # Check that the requested period matches period_behavior
+        if not self.column.period_behavior == PERMANENT:
```
cbenz: Use `column` since `column = self.column` is declared above

openfisca_core/holders.py
michelbl: nice shot! :smiley:

openfisca_core/decompositions.py
michelbl: ok

openfisca_core/holders.py
michelbl: done

openfisca_core/holders.py
michelbl: done

MattiSG:
If at some point you feel it useful to get some non-expert dev's opinion on API and concepts naming, feel free to direct me at some specific elements 🙂

openfisca_core/columns.py
fpagnoux:
```py
from openfisca_core.enumerations import Enum
ENUM = Enum(['month', 'year'])
raise ValueError('Computation requested with wrong period unit for variable {} ({} instead of {})'.format('rsa', 'month', ENUM['year']))

>>> ValueError: Computation requested with wrong period unit for variable rsa (month instead of 1)
```

This is for me a huge drawback to the use of the current polyfill of `Enum`. Also, writing `ENUM['year']` is super heavy.

If we can use a better polyfill, ok, but the current one is not good enough.

fpagnoux: Maybe we can use https://pypi.python.org/pypi/enum34.

michelbl: Are we that far from moving to python3 ?

fpagnoux: No idea. @cbenz ?

openfisca_core/columns.py
fpagnoux: If we don't have a consensus on a better name that `period_unit`, let switch back to this. Search-friendliness matters much less than name meaningfulness.

openfisca_core/holders.py
fpagnoux: What kind of impacts are we thinking about ? If we just offer a constant named  period.INFINITY, which is an instance of period (or at least offer the same API), we should be fine, no ?

openfisca_core/columns.py
fpagnoux: @MattiSG: we could use some advice on naming. We are wondering how to name the "native" (bad candidate) period of a variable. For instance, `rsa` can only be called for a month, `impot_revenu` can only be called for a year.

MattiSG: Can it **only** be called for that period, or is it its **definition** period but the engine will do its best to fit whatever you request from it?

openfisca_core/holders.py
```diff
@@ -112,164 +110,87 @@ def clone(self):
...
+            raise ValueError('Computation on period {} impossible on yearly variable {}'.format(
+                period, self.column.name
+                ))
+
+        if self.column.period_behavior == MONTH:
+            variable_period_unit = periods.MONTH
+        elif self.column.period_behavior == YEAR:
+            variable_period_unit = periods.YEAR
+        else:
+            ValueError('compute_add can be used only for yearly or monthly variables.')
```
fpagnoux: `compute_add` is not a method we want the formula writer to know about.

`Unable to sum date_de_naissance over 2015 : only variables defined monthly or yearly can be summed over time`

?

openfisca_core/holders.py
```diff
@@ -112,164 +110,87 @@ def clone(self):
...
-                    "Requested a monthly or yearly period. Got {} returned by variable {}.".format(
-                        dated_holder.period, self.column.name)
-                array = dated_holder.array * period.size / (12 * dated_holder.period.size)
-            return self.put_in_cache(array, period, parameters.get('extra_params'))
-        else:
-            assert unit == u'year', unit
-            return self.compute(period = period)
+        # Check that the requested period matches period_behavior
+        if self.column.period_behavior != YEAR:
+            raise ValueError('Holder.compute_divide must be used on yearly variables. This is not the case for {}'.format(
```
fpagnoux: Same, we want to spare our users the worry of wondering what a `Holder` could be.

`Unable to divide value of rsa over time : only variables defined yearly can be divided over time`

?

openfisca_core/holders.py
```diff
@@ -112,164 +110,87 @@ def clone(self):
...
+
+        if period.unit == periods.MONTH:
+            computation_period = period.this_year
+            dated_holder = self.compute(period = computation_period, **parameters)
+            array = dated_holder.array / 12.
+            return DatedHolder(self, period, array, parameters.get('extra_params'))
+        elif period.unit == periods.YEAR:
+            return self.compute(period, **parameters)
+
+        raise ValueError('Holder.compute_divide must be used on a month or a year. This is not the case for {} for variable {}'.format(
```
fpagnoux: Same here: rewrite error message following the same pattern.

openfisca_core/holders.py
MattiSG: I recommend that the constant (be it here or introduced at some later point) is called —as you probably heard me repeat orally  ;) — `period.ETERNITY`. `INFINITY` is strongly implicitly typed as a `Number`, as practically all languages provide some kind of `INFINITY` constant. `ETERNITY` is strongly implicitly typed as a time value.

michelbl:
I suggest to remove the attribute `is_permanent` since it is redundant with `period_unit = ETERNITY`.

openfisca_core/columns.py
michelbl: *definition period* is a good description, I think. The engine will optionally add or divide, depending on which function was called and the hook `calculate_output`.

benjello:
ETERNITY can have a have a lower bound. For the eternity is implicitly from now on. I suggest ALWAYS.

openfisca_core/periods.py
```diff
@@ -358,6 +358,41 @@ def __str__(self):
...
+                u'year': 3,
+                }
+            return cmp(unit_weights[unit_a], unit_weights[unit_b])
+
+        if size_a != size_b:
+            return cmp(size_a, size_b)
+
+        return cmp(start_a, start_b)
+
+    def __lt__(self, other):
```
fpagnoux: Not an expert about these magic functions, but I thought that when you defined `__cmp__`, all the other ones would by default use  `__cmp__`. It's not the case ?

openfisca_core/periods.py
```diff
@@ -358,6 +358,41 @@ def __str__(self):
             if fragment is not None
             )
 
+    def __cmp__(self, other):
+        unit_a, start_a, size_a = self
+        unit_b, start_b, size_b = other
+
+        if unit_a != unit_b:
+            unit_weights = {
+                u'day': 1,
```
fpagnoux: Do we really have periods with `day` as a unit ?

openfisca_core/periods.py
```diff
@@ -358,6 +358,41 @@ def __str__(self):
             if fragment is not None
             )
 
+    def __cmp__(self, other):
```
fpagnoux: @When do we need to compare periods based on their size ?

If we need this, we must document it : defining an order on periods is nothing but obvious (here `period('2014') > period('2015-01')`, which some users may not expect.`

openfisca_core/tests/test_countries.py
```diff
@@ -256,3 +262,20 @@ def test_non_existing_variable():
...
+        parent1 = dict(
+            salaire_brut = 4000,
+            ),
+        ).new_simulation()
+
+    with assert_raises(ValueError) as error:
+        simulation.calculate('rsa', 2013)
+
+    error_message = str(error.exception)
+    expected_error_message = 'Computation requested with wrong period unit for variable rsa (year instead of month)'
```
fpagnoux: Auto-review : testing the exact message of the error is too much. Maybe check that it contains some key words, or just check for the type of expression.

openfisca_core/formulas.py
```diff
@@ -639,14 +635,15 @@ def calculate_output_add(formula, period):
...
 
-def calculate_output_add_divide(formula, period):
-    return formula.holder.compute_add_divide(period).array
-
-
 def calculate_output_divide(formula, period):
     return formula.holder.compute_divide(period).array
 
 
+def calculate_output_first_month(formula, period):
```
fpagnoux: Did we need this to pass the `france` tests ?

openfisca_core/holders.py
fpagnoux: Possible alternatives could be period.ALWAYS (suggested by @benjello) or period.PERMANENT (closest to current name).

MattiSG: Even though @benjello's point about a lower point is totally valid from an analytical standpoint, I would still recommend `ETERNITY` for consistency with other enum names.

If the other types are `MONTH` and `YEAR`, for example, you want the last one to also be a noun, not an adjective (`PERMANENT`) or an adverb (`ALWAYS`).

This might sound like arguing over nothing at first, but I can tell you it will otherwise always have that little weird feeling when you type it, even though you might not realise why, and your brain will always be slightly longer to retrieve the word (until you're so used to it it doesn't carry any meaning anymore —but API design should target newcomers too :wink:).

benjello: I think that the naming problem comes because we use overlapping concepts (and I do not necessarily think that it is bad):
 - a quantitative/extensive meaning: for example the allocations familiales are for whole month
 - a date/time interval: at that date/during that period the amount of the allocation familiale was some value or your age was some other value.

No real clue though

openfisca_core/periods.py
michelbl: Not when you subclass tuple...

openfisca_core/formulas.py
michelbl: This is needed to compute a monthly variable for a year.

The yaml tests compute variables for the (implicit) simulation period. This period may not match the variable period unit, requiring to add a hook calculate_output. The need to remove the simulation period is already stated in #460 .

openfisca_core/periods.py
michelbl: An order on periods is used here : https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/scenarios.py#L55 https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/scenarios.py#L162

My opinion is that only instants should have an order, not periods. I may move the comparison operators from periods.py to scenarios.py and forbid implicit period ordering. Sounds cleaner to you ?

openfisca_core/holders.py
```diff
@@ -112,164 +110,87 @@ def clone(self):
...
 
-        # First look for dated_holders covering the whole period (without hole).
-        dated_holder = self.get_from_cache(period, parameters.get('extra_params'))
-        if dated_holder.array is not None:
-            return dated_holder
+        # Check that the requested period matches period_behavior
+        if not column.period_behavior == PERMANENT:
+            if ((column.period_behavior == MONTH and period.unit != periods.MONTH) or
+                    (column.period_behavior == YEAR and period.unit != periods.YEAR)):
+                raise ValueError('Computation requested with wrong period unit for variable {} ({} instead of {})'.format(
```
michelbl: A link to the documentation wouldn't be even better ?

`Unable to compute monthly variable {} for year-long period {}. See https://doc.openfisca.fr/coding-the-legislation/35_periods.html'`

openfisca_core/periods.py
fpagnoux: You may want to define the `compare_period_size(x,y)` function is the `periods` module though, for better encapsulation.

openfisca_core/formulas.py
```diff
@@ -925,77 +923,111 @@ def new_filled_column(__doc__ = None, __module__ = None,
...
-                month = period.start.period(u'month')
-                while month.start < after_instant:
-                    if holder.get_array(month) is None:
-                        holder.put_in_cache(divided_array, month)
-                    month = month.offset(1)
-
-
-def set_input_neutralized(formula, period, array):
+        else:
+            pass  # raise ValueError('Values for variable {} are already set on period {}.'.format(holder.column.name, period))
```
fpagnoux: Either raise the error, or remove the else block.

If I understand well, we are in this case when in the input there is both the year value and the 12 months ones. Raising an error protects us from being inconsistant, but may be surprising for a user providing an overdefined but consistent input.

openfisca_core/periods.py
michelbl: https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/periods.py#L118

The question is "Do we really _use_ periods with `day` as a unit ?" I dare not check...

openfisca_core/tests/test_countries.py
michelbl: done

openfisca_core/formulas.py
fpagnoux: The `calculate_output` mechanism is a dirty hook, let's try to not invest more in it.
I think we can avoid it by slightly modyfing the tests in france (see the `france` PR)

openfisca_core/scenarios.py
```diff
@@ -15,6 +15,10 @@ def N_(message):
     return message
 
 
+def compare_periods(a, b):
+    return periods.compare_period_size(a, b) or periods.compare_period_start(a, b)
```
fpagnoux: Is there any value to sort the periods by start ? If no, we can just use `periods.compare_period_start` as `cmp` and avoid defining a local function.

fpagnoux: If this helper is actually needed, please define it near where it's needed too.

openfisca_core/base_functions.py
```diff
@@ -71,41 +79,23 @@ def requested_period_last_value(formula, simulation, period, *extra_params, **kw
     accept_future_value = kwargs.pop('accept_future_value', False)
     holder = formula.holder
     if holder._array_by_period is not None:
-        known_values = sorted(holder._array_by_period.iteritems(), reverse = True)
+        known_values = sorted(holder._array_by_period.iteritems(), cmp = compare_periods, reverse = True)
```
michelbl: I renamed it to `compare_start_instant` and I added a comment.

openfisca_core/scenarios.py
michelbl: removed

openfisca_core/tests/test_countries.py
```diff
@@ -154,26 +157,28 @@ def test_2_parallel_axes_same_values():
...
     year = 2013
+    month = '2013-01'
     simulation = tax_benefit_system.new_scenario().init_single_entity(
         period = year,
         parent1 = dict(
             birth = datetime.date(year - 40, 1, 1),
             ),
         ).new_simulation(debug = True)
-    assert_near(simulation.calculate('age'), [40], absolute_error_margin = 0.005)
+    assert_near(simulation.calculate('age', '2013-01'), [40], absolute_error_margin = 0.005)
```
cbenz: Use `month`

openfisca_core/tests/test_countries.py
```diff
@@ -154,26 +157,28 @@ def test_2_parallel_axes_same_values():
...
 
     simulation = tax_benefit_system.new_scenario().init_single_entity(
-        period = year,
+        period = month,
         parent1 = dict(
             age_en_mois = 40 * 12 + 11,
             ),
         ).new_simulation(debug = True)
-    assert_near(simulation.calculate('age'), [40], absolute_error_margin = 0.005)
+    assert_near(simulation.calculate('age', '2013-01'), [40], absolute_error_margin = 0.005)
```
cbenz: Use `month`

openfisca_core/tests/test_simulations.py
```diff
@@ -14,25 +11,5 @@
 simulation = scenario.new_simulation(trace=True)
```
cbenz: Move `simulation` in test, now that there is only one test left.

PR #436, opened 2017-01-25
Name compact node

fpagnoux:
* Always name a `CompactNode`, as otherwise traced simulations crash if a `BAREME` is directly under the root node of the xml parameteres file.

This bug was noticed working on `senegal`.

Connected to openfisca/openfisca-web-api#76.

PR #454, opened 2017-02-16
Handle case when CompactNode.name is None

cbenz:
This bug was noticed working on senegal.

Connected to openfisca/openfisca-web-api#76.

Replaces #436


fpagnoux:
Thanks for digging into this and suggesting a fix ! I'll double check it works on Senegal and come back to you.

fpagnoux:
I checked, it works, GTM

PR #455, opened 2017-02-17
Improve test runner

fpagnoux:
* Improve `openfisca-run-test` script
  - Make country package detection more robust (it only worked for packages installed in editable mode)
  - Use spaces instead of commas as separator in the script arguments when loading several extensions or reforms (this is more standard)
* Refactor the `scripts` module to seperate the logic specific to yaml test running from the one that can be re-used by any script which needs to build a tax and benefit system.

### DIFF
--- a/openfisca_core/scripts/__init__.py
+++ b/openfisca_core/scripts/__init__.py
@@ -1,7 +1,11 @@
 # -*- coding: utf-8 -*-
 
-import sys
 import importlib
+import logging
+import sys
+
+log = logging.getLogger(__name__)
+logging.basicConfig(format='%(levelname)s: %(message)s')
 
 
 def add_tax_benefit_system_arguments(parser):
@@ -18,7 +22,7 @@ def build_tax_benefit_sytem(country_package_name, extensions, reforms):
             country_package = importlib.import_module(country_package_name)
             assert hasattr(country_package, 'CountryTaxBenefitSystem')
         except (ImportError, AssertionError):
-            print('ERROR: `{}` does not seem to be a valid Openfisca country package.'.format(country_package_name))
+            log.error('`{}` does not seem to be a valid Openfisca country package.'.format(country_package_name))
             sys.exit(1)
     else:
         country_package_name = detect_country_package()
@@ -38,7 +42,7 @@ def build_tax_benefit_sytem(country_package_name, extensions, reforms):
                 reform = getattr(reform_module, reform_name)
                 tax_benefit_system = reform(tax_benefit_system)
             except:
-                print('ERROR: `{}` does not seem to be a valid Openfisca reform for `{}`.'.format(reform_path, country_package.__name__))
+                log.error('`{}` does not seem to be a valid Openfisca reform for `{}`.'.format(reform_path, country_package.__name__))
                 raise
 
     return tax_benefit_system
@@ -58,8 +62,8 @@ def detect_country_package():
                 installed_country_packages.append(module_name)
 
     if len(installed_country_packages) == 0:
-        print('ERROR: No country package has been detected on your environment. If your country package is installed but not detected, please use the --country_package option.')
+        log.error('No country package has been detected on your environment. If your country package is installed but not detected, please use the --country_package option.')
         sys.exit(1)
     if len(installed_country_packages) > 1:
-        print('WARNING: Several country packages detected : `{}`. Using `{}` by default. To use another package, please use the --country_package option.'.format(', '.join(installed_country_packages), installed_country_packages[0]))
+        log.warning('Several country packages detected : `{}`. Using `{}` by default. To use another package, please use the --country_package option.'.format(', '.join(installed_country_packages), installed_country_packages[0]))
     return installed_country_packages[0]
--- a/openfisca_core/scripts/__init__.py
+++ b/openfisca_core/scripts/__init__.py
@@ -12,12 +12,13 @@ def add_tax_benefit_system_arguments(parser):
     return parser
 
 
-def build_tax_benefit_sytem(country_package, extensions, reforms):
-    if country_package:
+def build_tax_benefit_sytem(country_package_name, extensions, reforms):
+    if country_package_name:
         try:
-            country_package = importlib.import_module(country_package)
-        except:
-            print('ERROR: `{}` does not seem to be a valid Openfisca country package.'.format(country_package))
+            country_package = importlib.import_module(country_package_name)
+            assert hasattr(country_package, 'CountryTaxBenefitSystem')
+        except (ImportError, AssertionError):
+            print('ERROR: `{}` does not seem to be a valid Openfisca country package.'.format(country_package_name))
             sys.exit(1)
     else:
         country_package_name = detect_country_package()
--- a/openfisca_core/scripts/__init__.py
+++ b/openfisca_core/scripts/__init__.py
@@ -2,6 +2,7 @@
 
 import importlib
 import logging
+import pkgutil
 import sys
 
 from openfisca_core.reforms import Reform
@@ -64,15 +65,11 @@ def build_tax_benefit_sytem(country_package_name, extensions, reforms):
 
 
 def detect_country_package():
-    import pkgutil
-    from importlib import import_module
-
     installed_country_packages = []
-
     for module_description in pkgutil.iter_modules():
         module_name = module_description[1]
         if 'openfisca' in module_name.lower():
-            module = import_module(module_name)
+            module = importlib.import_module(module_name)
             if hasattr(module, 'CountryTaxBenefitSystem'):
                 installed_country_packages.append(module_name)
 
--- a/openfisca_core/scripts/__init__.py
+++ b/openfisca_core/scripts/__init__.py
@@ -4,8 +4,15 @@ import importlib
 import logging
 import sys
 
+from openfisca_core.reforms import Reform
+
 log = logging.getLogger(__name__)
 logging.basicConfig(format='%(levelname)s: %(message)s')
+
+
+def handle_error(error_message):
+    log.error(error_message)
+    sys.exit(1)
 
 
 def add_tax_benefit_system_arguments(parser):
@@ -17,13 +24,14 @@ def add_tax_benefit_system_arguments(parser):
 
 
 def build_tax_benefit_sytem(country_package_name, extensions, reforms):
+
     if country_package_name:
         try:
             country_package = importlib.import_module(country_package_name)
-            assert hasattr(country_package, 'CountryTaxBenefitSystem')
-        except (ImportError, AssertionError):
-            log.error('`{}` does not seem to be a valid Openfisca country package.'.format(country_package_name))
-            sys.exit(1)
+        except ImportError:
+            handle_error('Could not import module `{}`. Make sure it is installed in your environment.'.format(country_package_name))
+        if not hasattr(country_package, 'CountryTaxBenefitSystem'):
+            handle_error('`{}` does not seem to be a valid Openfisca country package.'.format(country_package_name))
     else:
         country_package_name = detect_country_package()
         country_package = importlib.import_module(country_package_name)
@@ -37,13 +45,20 @@ def build_tax_benefit_sytem(country_package_name, extensions, reforms):
     if reforms:
         for reform_path in reforms:
             try:
-                [reform_package, reform_name] = reform_path.rsplit('.', 1)
+                reform_package, reform_name = reform_path.rsplit('.', 1)
+            except ValueError:
+                handle_error('`{}` does not seem to be a path pointing to a reform. A path looks like `some_country_package.reforms.some_reform.`'.format(reform_path))
+            try:
                 reform_module = importlib.import_module(reform_package)
+            except ImportError:
+                handle_error('Could not import `{}`.'.format(reform_package))
+            try:
                 reform = getattr(reform_module, reform_name)
-                tax_benefit_system = reform(tax_benefit_system)
-            except:
-                log.error('`{}` does not seem to be a valid Openfisca reform for `{}`.'.format(reform_path, country_package.__name__))
-                raise
+            except AttributeError:
+                handle_error('{} has no attribute {}'.format(reform_package, reform_name))
+            if not isinstance(reform, Reform):
+                handle_error('`{}` does not seem to be a valid Openfisca reform for `{}`.'.format(reform_path, country_package.__name__))
+            tax_benefit_system = reform(tax_benefit_system)
 
     return tax_benefit_system
 
@@ -62,8 +77,7 @@ def detect_country_package():
                 installed_country_packages.append(module_name)
 
     if len(installed_country_packages) == 0:
-        log.error('No country package has been detected on your environment. If your country package is installed but not detected, please use the --country_package option.')
-        sys.exit(1)
+        handle_error('No country package has been detected on your environment. If your country package is installed but not detected, please use the --country_package option.')
     if len(installed_country_packages) > 1:
         log.warning('Several country packages detected : `{}`. Using `{}` by default. To use another package, please use the --country_package option.'.format(', '.join(installed_country_packages), installed_country_packages[0]))
     return installed_country_packages[0]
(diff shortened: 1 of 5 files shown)


openfisca_core/scripts/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+
+def detect_country_package():
+    import pkgutil
+    from importlib import import_module
+
+    installed_country_packages = []
+
+    for module_description in pkgutil.iter_modules():
+        module_name = module_description[1]
+        if module_name.lower().find('openfisca') >= 0:
```
cbenz: Python idiom: use `module_name.lower().startswith('openfisca')`

openfisca_core/scripts/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+        'short': 'e',
+        'help': 'extensions to load',
+        'nargs': '*'
+        },
+    'reforms': {
+        'short': 'r',
+        'help': 'reforms to apply to the country package',
+        'nargs': '*'
+        }
+    }
```
cbenz: Relying on such a `dict` seems too much code factorization, I'd prefer having many `parser.add_argument` like in previous version before this patch, but factorized in `add_tax_benefit_system_arguments`.

openfisca_core/scripts/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+    return parser
+
+
+def build_tax_benefit_sytem(country_package_name, extensions, reforms):
+    if country_package_name:
+        try:
+            country_package = importlib.import_module(country_package_name)
+            assert hasattr(country_package, 'CountryTaxBenefitSystem')
+        except:
+            print('ERROR: `{}` does not seem to be a valid Openfisca country package.'.format(country_package_name))
```
cbenz: The strongest way to use `print`, which leads to no runtime exception (`UnicodeDecodeError`) is:

```python
person_name = u'Félix' # Note the accent
print(u'Hello {}!'.format(person_name).encode('utf-8'))
```

(that's heavy, yeah)

In this case `country_package_name` can be only ASCII so it's OK, but it's implicit.

openfisca_core/scripts/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+
+    return parser
+
+
+def build_tax_benefit_sytem(country_package_name, extensions, reforms):
+    if country_package_name:
+        try:
+            country_package = importlib.import_module(country_package_name)
+            assert hasattr(country_package, 'CountryTaxBenefitSystem')
+        except:
```
cbenz: Bad practice not to specify which exception to catch.

This could lead to catch an exception which should have bubbled up.

Moreover I think it could catch the exception raised by `assert`.

In this specific case I understand that everything must be catched, so I'd use `except Exception` (see [Python doc](https://docs.python.org/2/library/exceptions.html#exceptions.Exception))

openfisca_core/scripts/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+            tax_benefit_system.load_extension(extension)
+
+    if reforms:
+        for reform_path in reforms:
+            try:
+                [reform_package, reform_name] = reform_path.rsplit('.', 1)
+                reform_module = importlib.import_module(reform_package)
+                reform = getattr(reform_module, reform_name)
+                tax_benefit_system = reform(tax_benefit_system)
+            except:
```
cbenz: Idem about catching everything.

In this case I'm pretty sure we can end up testing things with `if`s rather than relying on `except`.

Second, having so much lines in the `try` block is bad code smell.

openfisca_core/scripts/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+                print('ERROR: `{}` does not seem to be a valid Openfisca reform for `{}`.'.format(reform_path, country_package.__name__))
```
cbenz: Idem

In this case, I anticipate a `UnicodeDecodeError` on runtime if `reform_path` is not UTF-8 encoded (which is the good practice in Python: all strings are `u` strings).

The canonical form should be:

```python
print(u'ERROR: `{}` does not seem to be a valid Openfisca reform for `{}`.'.format(reform_path, country_package.__name__).encode('utf-8'))
```

But that's a real pain! So I recommend you to use the Python `logging` module since it solves the encoding-to-the-terminal problem + adding the `ERROR` prefix.

---

```
a = u'è'
print('A: {}'.format(a))
---------------------------------------------------------------------------
UnicodeEncodeError                        Traceback (most recent call last)
<ipython-input-2-21d6f94fb722> in <module>()
----> 1 print('A: {}'.format(a))

UnicodeEncodeError: 'ascii' codec can't encode character u'\xe8' in position 0: ordinal not in range(128)

```

openfisca_core/scripts/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+    installed_country_packages = []
+
+    for module_description in pkgutil.iter_modules():
+        module_name = module_description[1]
+        if module_name.lower().find('openfisca') >= 0:
+            module = import_module(module_name)
+            if hasattr(module, 'CountryTaxBenefitSystem'):
+                installed_country_packages.append(module_name)
+
+    if len(installed_country_packages) == 0:
```
cbenz: Python idiom: use `if not installed_country_packages`

💭 Personally I'm against this idiom, since it relies on automatic casting from lists to booleans!

What to do when there is a bad idiom? 🤔

openfisca_core/scripts/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+        print('ERROR: No country package has been detected on your environment. If your country package is installed but not detected, please use the --country_package option.')
+        sys.exit(1)
+    if len(installed_country_packages) > 1:
```
cbenz: Use `elif`?

openfisca_core/scripts/__init__.py
fpagnoux: Each line can fail, and respectively raise a `ValueError`, `AttributeError`, `ImportError` or `TypeError`.
Each line could have a `if` before, but that would double the amount of code, and bring some heavy logics.

Do you see a better way of implementing this ? The only goal is to get the python objects corresponding to the path, such as `openfisca_france.reforms.aides_cd93.aides_cd93`. Is there some more idiomatic way ?

openfisca_core/scripts/__init__.py
fpagnoux: I explicited the exceptions we want to catch

openfisca_core/scripts/__init__.py
fpagnoux: 👍

openfisca_core/scripts/__init__.py
fpagnoux:
```ssh
openfisca-run-test test_mes_aides_53d21ee5159e330200810b87.yaml  -c éàêù
ERROR: `éàêù` does not seem to be a valid Openfisca country package.
```

openfisca_core/scripts/__init__.py
fpagnoux: It's exactly the same behaviour. I just want `openfisca` to be in the module name, not necessary in the middle. 

Is there a more idiomatic way to do it ? `'openfisca' in module_name.lower()`?

openfisca_core/scripts/__init__.py
fpagnoux: This idiom is very much JS-like and I like its lightness in some cases, but I feel like here we understand better the intention with the explicit condition ("There is zero packages that seems to be an openfisca country package").

openfisca_core/scripts/__init__.py
fpagnoux: It's not needed, as the previous condition ends with a `sys.exit()`. Is it a good practise to explicit the `elif` anyway ? 

In JS we usually don't do it with early exits.

openfisca_core/scripts/__init__.py
fpagnoux: But let's use the logging module if it's more standard.

openfisca_core/scripts/__init__.py
fpagnoux: Err I just ran into an unexpected `TypeError` giving a `.` as the country input.

In a specific script such as this one, where many things can go wrong with, it seems to me that catching everything while processing the user input is the most natural.

openfisca_core/scripts/__init__.py
cbenz: The catch-all is very sneaky, could catch any error you don't want to catch, and I would always insist to avoid it!

> Each line can fail, and respectively raise a ValueError, AttributeError, ImportError or TypeError.

```python
invalid_error_message = u'ERROR: ...'.format(...)

tax_benefit_system = None

try:
    [reform_package, reform_name] = reform_path.rsplit('.', 1)
except ValueError:
    log.exception(invalid_reform_message)
else:
    try:
        reform_module = importlib.import_module(reform_package)
    except ImportError:
        log.exception(invalid_reform_message)
    else:
        reform = getattr(reform_module, reform_name, None)
        if reform is not None:
            try:
                tax_benefit_system = reform(tax_benefit_system)
            except Exception:
                log.exception(invalid_reform_message)

return tax_benefit_system
```

This is absolutely not tested and I agree it's longer, but more secure, and can be perhaps simplified.

openfisca_core/scripts/__init__.py
cbenz: Yes, it's the `in` operator.

openfisca_core/scripts/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+    tax_benefit_system = country_package.CountryTaxBenefitSystem()
+
+    if extensions:
+        for extension in extensions:
+            tax_benefit_system.load_extension(extension)
+
+    if reforms:
+        for reform_path in reforms:
+            try:
+                [reform_package, reform_name] = reform_path.rsplit('.', 1)
```
l-vincent-l: This should be outside the the try.
Are you sure reform_path contains a '.' ?
It's more pythonic to write : reform_package, reform_name = reform_path.rsplit('.', 1)

fpagnoux: Not sure of anything, reform_path is an argument of the CLI script.

openfisca_core/scripts/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+                reform_module = importlib.import_module(reform_package)
```
l-vincent-l: I think what you want to do is :
try:
   reform_module = importlib.import_module(reform_package)
except ImportError:
   handle_error...
else:
  ....

openfisca_core/scripts/__init__.py
l-vincent-l: ok, so you want to test it before.

openfisca_core/scripts/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+                reform = getattr(reform_module, reform_name)
```
l-vincent-l: Here: 
try:
    reform = getattr
except AttributeError:
   handle_error...

openfisca_core/scripts/__init__.py
l-vincent-l: Here you will also catch the AssertionError of the assert.

openfisca_core/scripts/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+    if country_package_name:
+        try:
+            country_package = importlib.import_module(country_package_name)
+            assert hasattr(country_package, 'CountryTaxBenefitSystem')
+        except:
+            log.error('`{}` does not seem to be a valid Openfisca country package.'.format(country_package_name))
+            sys.exit(1)
+    else:
+        country_package_name = detect_country_package()
+        country_package = importlib.import_module(country_package_name)
```
l-vincent-l: I think you want to catch a ImportError here.

openfisca_core/scripts/__init__.py
cbenz: > Here you will also catch the AssertionError of the assert.

or use directly

```python
if hasattr(country_package, 'CountryTaxBenefitSystem')
```

;-)

fpagnoux:
Here you are, it still seems over-quality to me given the scope, but let's go !

openfisca_core/scripts/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+def build_tax_benefit_sytem(country_package_name, extensions, reforms):
+
+    if country_package_name:
+        try:
+            country_package = importlib.import_module(country_package_name)
+        except ImportError:
+            handle_error('Could not import module `{}`. Make sure it is installed in your environment.'.format(country_package_name))
+        if not hasattr(country_package, 'CountryTaxBenefitSystem'):
+            handle_error('`{}` does not seem to be a valid Openfisca country package.'.format(country_package_name))
+    else:
```
l-vincent-l: I think it would be better to write:
`
def build_tax_benefit_sytem(country_package_name, extensions, reforms):
if country_package_name is None:
    country_package_name = detect_country_package()
try:
            country_package = importlib.import_module(country_package_name)
        except ImportError:
            handle_error('Could not import module `{}`. Make sure it is installed in your environment.'.format(country_package_name))
        if not hasattr(country_package, 'CountryTaxBenefitSystem'):
           handle_error('`{}` does not seem to be a valid Openfisca country package.'.format(country_package_name))
`

openfisca_core/scripts/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+logging.basicConfig(format='%(levelname)s: %(message)s')
+
+
+def handle_error(error_message):
+    log.error(error_message)
+    sys.exit(1)
+
+
+def add_tax_benefit_system_arguments(parser):
+    parser.add_argument('-c', '--country_package', action = 'store', help = 'country package to use. If not provided, an automatic detection will be attempted by scanning the python packages installed in your environment which name contains the word "openfisca".')
```
cbenz: The `help = ` strings should have `u''`

openfisca_core/scripts/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+    return parser
+
+
+def build_tax_benefit_sytem(country_package_name, extensions, reforms):
+
+    if country_package_name:
+        try:
+            country_package = importlib.import_module(country_package_name)
+        except ImportError:
+            handle_error('Could not import module `{}`. Make sure it is installed in your environment.'.format(country_package_name))
```
cbenz: All the `handle_error` strings should be unicode-strings (have a `u''`)

Otherwise, any unicode-string mentioned in `.format()` arguments will certainly provoke a `UnicodeDecodeError`

Remember that in Python2 using unicode-strings is the default. Encoded-string (UTF-8) happens only when outputting the string to the external world (terminal, network request, etc.)

openfisca_core/scripts/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+                reform_package, reform_name = reform_path.rsplit('.', 1)
+            except ValueError:
+                handle_error('`{}` does not seem to be a path pointing to a reform. A path looks like `some_country_package.reforms.some_reform.`'.format(reform_path))
+            try:
+                reform_module = importlib.import_module(reform_package)
+            except ImportError:
+                handle_error('Could not import `{}`.'.format(reform_package))
+            try:
+                reform = getattr(reform_module, reform_name)
+            except AttributeError:
```
cbenz: Could be

```python
reform = getattr(reform_module, reform_name, None)
if reform is None:
    handle_error(u'Could not import `{}`.'.format(reform_package))```

(Prefer conditions over exceptions)

File: openfisca_core/scripts/__init__.py [hunk 17]
```diff
@@ -0,0 +1,83 @@
...
+            if not isinstance(reform, Reform):
+                handle_error('`{}` does not seem to be a valid Openfisca reform for `{}`.'.format(reform_path, country_package.__name__))
+            tax_benefit_system = reform(tax_benefit_system)
+
+    return tax_benefit_system
+
+
+def detect_country_package():
+    import pkgutil
+    from importlib import import_module
```
cbenz: Can those packages be imported at module level?

PR #462, opened 2017-03-02
Simplify periods amendment

fpagnoux:
@michelbl 

Here is my solution to avoid using options for `set_input`.

We keep the current behaviour of `set_input`, which is to "fill" the wholes. The underlying assumption is that when you build a scenario, `set_input` should only be called once per variable. 

This was the case by conception for regular simulations, but axes were calling `set_input`  potentially several times on the same variable, and potentially and variables that had already been `set_input`ed before in the regular initialisation.

My fix is, when they are axes, to build a buffer containing the inputs, and only in the end to call the `set_input`.

It would be simpler to just use the buffer in all cases, but I'm not sure about the performance impact for axes-less big simulations.

I opened the PR to ease the dialog, but if it's GTM we should not merge it and commit the content in the regular branch.

### DIFF
--- a/openfisca_core/formulas.py
+++ b/openfisca_core/formulas.py
@@ -29,10 +29,6 @@ log = logging.getLogger(__name__)
 ADD = 'add'
 DIVIDE = 'divide'
 
-SET_INPUT_FILL = u'set_input_fill'
-SET_INPUT_REPLACE = u'set_input_replace'
-SET_INPUT_ADD = u'set_input_add'
-
 
 # Exceptions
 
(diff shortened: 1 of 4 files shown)


fpagnoux:
Commits integrated in #452.

I added these commits to the main branch without a pull request merge. Not sure what is the best practise, but I like linear git histories :)

MattiSG:
> I added these commits to the main branch without a pull request merge

You can use the online “rebase and merge” option, or rebase, force push on the origin branch and fast-forward the destination branch  :)

Issue #468, opened 2017-03-07
Make explicit output variables requested periods


MattiSG:
To avoid mistakes where the period is not specified and the “simulation period” is used, make it mandatory to specify periods in `calculate` / `compute` calls.

MattiSG:
@benjello do you see any risk / difficulty in doing this?

benjello:
I agree that it is the right thing to do. But it will definitively impact many packages downstream.
But the fix should be easy by using if needed a period at a Scenario (or SurveyScenario) level instead of a simulation level.

PR #469, opened 2017-03-08
Make it mandatory to specify periods in variable calls

michelbl:
Connected to #468

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 7.0.0
+
+* Make it mandatory to provide a period when calculating a variable.
+  - When a computation is requested (with *.calculate_output, entities.__call__, *.calculate[_add|_divide], *.compute[_add|_divide]), the argument `period` is no longer optional.
+  - Previously, the period of the simulation was used instead of a missing period. It was error-prone, as values would be returned for the wrong period, without any error or warning to alert the formula writer about a likely coding error.
+
 ## 6.1.0
 
 * Move `base.py` content (file usually located in country packages) to core module `formula_toolbox` so that it can be reused by all countries
--- a/openfisca_core/tests/test_reforms.py
+++ b/openfisca_core/tests/test_reforms.py
@@ -36,13 +36,13 @@ def test_formula_neutralization():
     simulation = scenario.new_simulation(debug = True, reference = True)
     rsa = simulation.calculate('rsa', period = '2013-01')
     assert_near(rsa, 300, absolute_error_margin = 0)
-    revenu_disponible = simulation.calculate('revenu_disponible')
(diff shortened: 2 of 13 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 7.0.0
+
+* Make it mandatory to provide a period when calling a variable.
```
fpagnoux: `calling` -> `calculating`

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
...
+
+When a computation is requested (with *.calculate_output, entities.__call__, *.calculate[_add|_divide], *.compute[_add|_divide]), the argument `period` is no longer optional. Previously, the period of the simulation was used instead of a missing period. It could introduce silently computation errors.
```
fpagnoux: For consistency with the rest of the changelog, make it a sub-bullet (and maybe split lines before "Previously")

fpagnoux: `It could introduce silently computation errors.` -> `It was error-prone, as values would be returned for the wrong period, without any error or warning to alert the formula writer about a likely coding error.`

openfisca_core/tests/test_reforms.py
```diff
@@ -101,8 +101,8 @@ def apply(self):
         )
     simulation = scenario.new_simulation(reference = True)
     reform_simulation = scenario.new_simulation()
-    assert str(simulation.calculate('birth')[0]) == '1980-01-01'
-    assert str(reform_simulation.calculate('birth')[0]) == '1970-01-01'
+    assert str(simulation.calculate('birth', None)[0]) == '1980-01-01'
+    assert str(reform_simulation.calculate('birth', None)[0]) == '1970-01-01'
```
fpagnoux: This not great, but adding a specific treatment for permanent variables is a lot of complexity, given the amount of code repetition 🤔 .

There are not so many permanent columns, so it's good enough for me now.

CHANGELOG.md
michelbl: done

openfisca_core/tests/test_reforms.py
michelbl: To carry on the implementation of the period, the cleanest way would be to define a new "eternity period" as a triplet(u"eternity", None, None). I confirm that this would be a lot of work, requiring to update everything in the core.

fpagnoux:
Careful though : force pushing on a branch after asking for reviews is a bad practise. Here nothing happened, but it can break links, and make the reviewer work harder. 

Usually, the workflow is :
- Get a review
- Fix what is needed by adding new commits (potentially fixups)
- Get all the approvals you need
- Then only before merging, rebase and squash the fixups

PR #476, opened 2017-03-20
Extract dummy country from core

fpagnoux:
Connected to openfisca/openfisca-web-api#100

We need it to test the api

It's also a first step towards merging the dummy country and the country package boilerplate.


setup.py
```diff
@@ -35,6 +35,7 @@
             ],
         'test': [
             'nose',
+            'openfisca-dummy-country'
```
cbenz: OK for no version constraint for `openfisca-dummy-country` if `pip` is able to find a valid version of `openfisca-dummy-country` when doing `pip install openfisca-core[test]`.

fpagnoux: Excactly, the version constraint is probably going to come from the other direction (`dummy_country` references a `core`)

fpagnoux:
@cbenz you reviewed without approving, is this ok to merge or do you plan to spend more time on it ?

PR #483, opened 2017-03-31
Document neutralize_variable

fpagnoux:
Connected to openfisca/openfisca-doc#43

I tried something new as an experiment: deprecating without removing a function.

My motivation is to avoid the fixed cost of publishing a major (updating all the dependencies) for something small where retro-compatibility is easy to implement.

The workflow I imagine would be wait for the next major bumps to remove it for good. Either we:
- Remove it in the next PR that brings a major bump. Doesn't respect the one PR = one goal principle.
- Publish the next major bump as a pre-release. Then all deprecated stuff can be removed right after.

To be discussed later with @MattiSG @cbenz, this is not a priority.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,4 +1,4 @@
 # pip doesn't handle well dependencies with a relative path inside a requirements.txt file.
-# If we put .. instead of 'openfisca_core >= 4.2.0, < 5.0', it will look for a setup.py in the parent directory relatively to where the instruction is executed from, and not where the requirements file is.
-openfisca_core >= 4.2.0, < 6.0
+# If we put .. instead of the first line, it will look for a setup.py in the parent directory relatively to where the instruction is executed from, and not where the requirements file is.
+--editable git+https://github.com/openfisca/openfisca-core.git@master#egg=OpenFisca-Core
 sphinx-argparse
--- a/openfisca_core/taxbenefitsystems.py
+++ b/openfisca_core/taxbenefitsystems.py
@@ -10,6 +10,7 @@ import importlib
 import logging
 import inspect
 import pkg_resources
+import warnings
 
 from setuptools import find_packages
 
@@ -250,15 +251,16 @@ class TaxBenefitSystem(object):
 
         A neutralized variable always returns its default value when computed.
 
-        Trying to set inputs for a neutralized variable has no effect.
+        Trying to set inputs for a neutralized variable has no effect except raising a warning.
         """
         self.update_column(variable_name, neutralize_column(self.get_column(variable_name)))
 
     def neutralize_column(self, column_name):
-        print(
-            u"Warning: the neutralize_column method has been renamed to neutralize_variable. "
+        warnings.warn(
+            u"The neutralize_column method has been renamed to neutralize_variable. "
             u"neutralize_column has thus been deprecated and will be removed in the next major version. "
-            u"Please update your code."
+            u"Please update your code.",
+            Warning
             )
         self.neutralize_variable(column_name)
 
--- a/openfisca_core/taxbenefitsystems.py
+++ b/openfisca_core/taxbenefitsystems.py
@@ -244,8 +244,16 @@ class TaxBenefitSystem(object):
     def update_column(self, column_name, new_column):
         self.column_by_name[column_name] = new_column
 
+    def neutralize_variable(self, variable_name):
+        self.update_column(variable_name, neutralize_column(self.get_column(variable_name)))
+
     def neutralize_column(self, column_name):
-        self.update_column(column_name, neutralize_column(self.get_column(column_name)))
+        print(
+            u"Warning: the neutralize_column method has been renamed to neutralize_variable. "
+            u"neutralize_column has thus been deprecated and will be removed soon. "
+            u"Please update your code."
+            )
+        self.neutralize_variable(column_name)
 
     def add_legislation_params(self, path_to_xml_file, path_in_legislation_tree = None):
         """
--- a/openfisca_core/taxbenefitsystems.py
+++ b/openfisca_core/taxbenefitsystems.py
@@ -17,7 +17,7 @@ from setuptools import find_packages
 from . import conv, legislations, legislationsxml
 from variables import AbstractVariable
 from scenarios import AbstractScenario
-from formulas import neutralize_column
+from formulas import get_neutralized_column
 
 log = logging.getLogger(__name__)
 
@@ -253,7 +253,7 @@ class TaxBenefitSystem(object):
 
         Trying to set inputs for a neutralized variable has no effect except raising a warning.
         """
-        self.update_column(variable_name, neutralize_column(self.get_column(variable_name)))
+        self.update_column(variable_name, get_neutralized_column(self.get_column(variable_name)))
 
     def neutralize_column(self, column_name):
         warnings.warn(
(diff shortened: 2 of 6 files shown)


openfisca_core/taxbenefitsystems.py
```diff
@@ -244,8 +244,23 @@ def get_column(self, column_name, check_existence = False):
...
+        A neutralized variable always returns its default value when being computed.
+
+        Trying to set inputs for a neutralized variable has no effect.
+        """
+        self.update_column(variable_name, neutralize_column(self.get_column(variable_name)))
+
     def neutralize_column(self, column_name):
-        self.update_column(column_name, neutralize_column(self.get_column(column_name)))
+        print(
+            u"Warning: the neutralize_column method has been renamed to neutralize_variable. "
```
fpagnoux: I didn't use `log.warning` as it does't work well because there is no `handlers`:

```
No handlers could be found for logger "openfisca_core.taxbenefitsystems"
```

MattiSG:
> The workflow I imagine would be wait for the next major bumps to remove it for good.

Perfect.

> Remove it in the next PR that brings a major bump. Doesn't respect the one PR = one goal principle.

Indeed. But the cost is of adding one commit that removes all stuff that's been marked as deprecated. It's not good, but we're getting closer to having internalised version constraints and discipline on publishing, at which point we will be able to design a proper releasing process. I like this intermediary step.

openfisca_core/taxbenefitsystems.py
```diff
@@ -244,8 +244,23 @@ def get_column(self, column_name, check_existence = False):
...
+            u"neutralize_column has thus been deprecated and will be removed soon. "
```
MattiSG: soon → in the next major version.

openfisca_core/taxbenefitsystems.py
```diff
@@ -244,8 +244,23 @@ def get_column(self, column_name, check_existence = False):
     def update_column(self, column_name, new_column):
         self.column_by_name[column_name] = new_column
 
+    def neutralize_variable(self, variable_name):
+        """
+        Neutralizes an OpenFisca variable existing in the tax and benefit system.
+
+        A neutralized variable always returns its default value when being computed.
```
MattiSG: > when being computed → when computed

openfisca_core/taxbenefitsystems.py
```diff
@@ -244,8 +244,23 @@ def get_column(self, column_name, check_existence = False):
...
+
+        Trying to set inputs for a neutralized variable has no effect.
```
MattiSG: Shouldn't it raise a warning? 😕

openfisca_core/taxbenefitsystems.py
MattiSG: It might be worth setting log handlers, because this should definitely be printed to `stderr`.

doc/requirements.txt
```diff
@@ -1,4 +1,4 @@
 # pip doesn't handle well dependencies with a relative path inside a requirements.txt file.
 # If we put .. instead of 'openfisca_core >= 4.2.0, < 5.0', it will look for a setup.py in the parent directory relatively to where the instruction is executed from, and not where the requirements file is.
-openfisca_core >= 4.2.0, < 6.0
+openfisca_core >= 9.0.0
```
MattiSG: I don't understand why this is changed.

openfisca_core/taxbenefitsystems.py
cbenz: I can help setting those logging handlers.

Handlers are set at the application / script (OpenFisca-Web-API) level, not the library (OpenFisca-Core) level. But the library can provide default settings.

fpagnoux: The typical user of `neutralize_column ` does't use the web API, so we cannot rely on the Web API handlers. 
If you know how to set relevant default handlers, feel free to add a commit to this PR ;)

doc/requirements.txt
fpagnoux: This is the version to be used in the [reference doc](http://openfisca.readthedocs.io/en/latest/). We want to take the last published.

fpagnoux:
We should probably use [`warnings.warn`](https://docs.python.org/2/library/warnings.html#warnings.warn) and not bother with logging for this PR.

openfisca_core/taxbenefitsystems.py
```diff
@@ -244,8 +245,24 @@ def get_column(self, column_name, check_existence = False):
     def update_column(self, column_name, new_column):
         self.column_by_name[column_name] = new_column
 
+    def neutralize_variable(self, variable_name):
+        """
+        Neutralizes an OpenFisca variable existing in the tax and benefit system.
+
+        A neutralized variable always returns its default value when computed.
+
+        Trying to set inputs for a neutralized variable has no effect.
```
cbenz: > `Trying to set inputs` => `Setting inputs` ?

openfisca_core/taxbenefitsystems.py
```diff
@@ -244,8 +245,24 @@ def get_column(self, column_name, check_existence = False):
...
+        """
+        self.update_column(variable_name, neutralize_column(self.get_column(variable_name)))
```
cbenz: Should this helper `neutralize_column` be renamed to `neutralized_variable` (I insist on *neutralized*, like in `new_filled_column`)?

Because `TaxBenefitSystem.neutralize_variable` is an action, whereas the helper in `formulas.py` is a function which creates a new `Variable`.

cbenz:
I added a commit to use [Python warning framework (PEP 230)](https://www.python.org/dev/peps/pep-0230/).

By default since Python 2.7, deprecation warnings are hidden by default ([explanation here](https://docs.python.org/2/library/warnings.html#updating-code-for-new-versions-of-python)).

The recommended way to display them is either:
- using the [`python -W` option](https://docs.python.org/2/using/cmdline.html#cmdoption-W)
- using the [`PYTHONWARNINGS`](https://docs.python.org/2/using/cmdline.html#envvar-PYTHONWARNINGS) environment variable

In short it is possible to convert deprecation warnings into exceptions (`python -W error`) or just display them (`python -W default`). This will have to be documented in a section of the doc with a title like "How to upgrade my OpenFisca user code" (any thought @fpagnoux @MattiSG)?

See also the "Command Line Syntax " section of the [PEP 230](https://www.python.org/dev/peps/pep-0230/) (sorry there is no direct link to an anchor).

fpagnoux:
Thanks @cbenz for all the references, this is interesting. 

In our case, I don't think a `Warning` that is hidden by default is much useful. First, we cannot convince our users to always run `python` with the `-w` option.  Second, a "How to upgrade my OpenFisca user code" tutorial would definitely have some value, but it's a project on its own, and if it only contains: "Use python with -W", it's not fulfilling its promise.

As we are still in the experimental phase about deprecation, I would just add general `Warning` (visible by default). This is less semantic, but closer to the concrete behaviour we expect.

openfisca_core/taxbenefitsystems.py
fpagnoux: As we _cannot_ set an input for a neutralized variable, I prefer using `try`.

openfisca_core/taxbenefitsystems.py
fpagnoux: 👍

Issue #437, opened 2017-01-26
Do not crash when asked for a parameter after end date


fpagnoux:
As someone who use openfisca as a backend of a simulation application,
My application doesn't crash overnight without me having done anything,
Because we just passed the end date of one legislative parameter

----------------

The changes introduced with the fuzzy parameters are a catastrophe for anyone using openfisca in prod. It introduces a major production risk : if we pass the end date of a parameter we use, then the application cannot do any calculation anymore.

The PNDS (institutional re-user) application was down for some time because of this, as well as some of our testing instances.

Openfisca should offer a "production" mode where it is flexible and extrapolates parameters values, and a "strict" mode where it crashed when a value is missing if it is necessary.

cbenz:
Anything to reproduce the bug or do post-mortem?

fpagnoux:
You can try to compute `ppa` or `2017-01` with `openfisca-france#9.0.0`

fpagnoux:
I don't think a post-mortem would bring much, the dynamics inside the team are different now that when this behaviour was introduced. Let's just mention it at the next planification and see how we can fix that.

fpagnoux:
See old discussions in https://github.com/openfisca/openfisca-core/issues/368

MattiSG:
> should offer a "production" mode where it is flexible and extrapolates parameters values, and a "strict" mode

Or just print warnings all the time. Indeed, let's discuss that in the next planning session  :)

fpagnoux:
>Or just print warnings all the time. Indeed, let's discuss that in the next planning session :)

That what I meant by:
>if it is necessary
;)

MattiSG:
The current implementation:
- raises questions from contributors: https://github.com/openfisca/openfisca-france/issues/539
- implies last-minute maintenance: https://github.com/openfisca/openfisca-france/pull/645

MattiSG:
Implement and document the following:

- [ ] Remove the `fuzzy` attribute.
- [ ] Remove the `fin` date attribute.
- [ ] Always extend values towards the future.
- [ ] Add the possibility to state a null value at some start date, to show that a parameter has been removed from law.
[...]

PR #488, opened 2017-04-14
Remove fuzzy

michelbl:
Connected to #437


openfisca_core/legislations.py
```diff
@@ -305,9 +303,12 @@ def generate_dated_bracket_json(bracket_json, instant_str):
 
 def generate_dated_json_value(values_json, instant_str):
     for value_json in values_json:
-        value_stop_str = value_json.get('stop')
-        if value_json['start'] <= instant_str and (value_stop_str is None or instant_str <= value_stop_str):
-            return value_json['value']
+        value_start_str = value_json.get('start')
```
fpagnoux: These `_str` are useless and make it hard to read this code.

michelbl:
I found a (dirty) way to ckeck the tests with travis : 
- Add the flag `--process-dependency-links` in `.travis.yml` : `- pip install --process-dependency-links --editable .[test]` This flag is deprecated, however no replacement is available yet.
- In `setup.py`, add `dependency_links = ['git+https://github.com/michelbl/openfisca-dummy-country.git@remove-fuzzy#egg=OpenFisca-Dummy-Country-0.2.0'],` The version number (0.2.0) can be anything but it is required.

fpagnoux:
To test:

- For a basic parameter with no end date or placeholder:
    - Check that requesting it before its first start period raises an error
    - Check that its value change at the expected time

- For a basic parameter with a placeholder:
    - Check that requesting after the placeholder date does not raise an error

- For a basic parameter with an end date:
    - Check that requesting after the end date does raise an error

openfisca_core/reforms.py
```diff
@@ -126,58 +126,43 @@ def build_node(node, path_index):
...
-    if not_contained_items and not_contained_items[-1].get('stop') is None \
-            and str(start) > not_contained_items[-1]['start']:
-        last_not_contained_item = not_contained_items[-1].copy()  # Copy to avoid modifying `items` argument.
-        last_not_contained_item['stop'] = str(start.offset(-1, 'day'))
-        not_contained_items[-1] = last_not_contained_item
-    new_items = sorted(not_contained_items + [new_item], key=lambda item: item['start'])
-    return new_items
-
-
-def create_item(start, stop, value):
```
fpagnoux: If you are deprecating and removing this method, you must mention it in the changelog.

openfisca_core/reforms.py
```diff
@@ -126,58 +126,43 @@ def build_node(node, path_index):
...
-    dict_items = []
-    if start is not None:
-        dict_items.append(('start', str(start)))
-    if stop is not None:
-        dict_items.append(('stop', str(stop)))
-    dict_items.append(('value', value))
-    return collections.OrderedDict(dict_items)
-
-
-def split_item_containing_instant(instant, items):
```
fpagnoux: Same here

openfisca_core/reforms.py
fpagnoux: @benjello is `create_item` still used to build reforms ?

openfisca_core/reforms.py
fpagnoux: @benjello is `split_item_containing_instant ` still used to build reforms ?

fpagnoux:
I rebased the branch on the latest master, you may have to `git reset --hard origin/remove-fuzzy` instead of `git pull` (careful, you can lose local changes on your local branch `remove-fuzzy` if you have not yet pushed them).

openfisca_core/reforms.py
michelbl: ok

openfisca_core/reforms.py
michelbl: ok

michelbl:
@fpagnoux good point! The exception was not raised because undefined brackets at a given instant where not discared.

This is now fixed. However, the test whether a bracket is defined for a given instant is : https://github.com/openfisca/openfisca-core/pull/488/files#diff-3de96604c5f1d166b9a78364d4584288R357
If, for example, the threshold is correctly ended but not the rate, the bracket will be invalid but not ignored.

I was not able to find a list of rules that must be verified for scales validate_brackets_json_dates, except the code of `validate_brackets_json_dates` https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/legislations.py#L879

I certainly should dig into that code and reimplement it without the fuzzy system. However that is a consequent amount of work. Is it worth it ?

fpagnoux:
The goal will be to replace the current way of writing them by something like [this](https://api-test.openfisca.fr/parameter/impot_revenu.bareme), so let's not spend too much time validating a syntax we want to depreciate.

fpagnoux:
Waiting for the tests of https://github.com/openfisca/openfisca-france/pull/731/files to pass to be convinced this is fully functional.

PR #496, opened 2017-05-03
Switch all URLs to HTTPS

MattiSG:


fpagnoux:
Not super convinced the value of these changes are worth the administrative cost publishing a new version.
I'm secretly hiding this into another PR.

PR #502, opened 2017-05-04
Make location computation more robust

fpagnoux:
Connected to openfisca/legislation-explorer#82

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+### 12.0.3
+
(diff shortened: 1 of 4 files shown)


fpagnoux:
Tested in [prod](https://api-test.openfisca.fr/variable/cmu_c) with a dev version.

fpagnoux:
It didn't work, as variables don't have modules in the context within which we import them.

cbenz:
> It didn't work, as variables don't have modules in the context within which we import them.

Now I remember it's the exact reason I implemented it with a sub-string.

Issue #495, opened 2017-04-26
Improve the error when the period argument is forgotten


fpagnoux:
Within a formula, if I write :

```py
  salary = person('salary') # instead of  person('salary', period)
```

I get the error :
```
 File "/path/to/file.py", line 157, in function
     salary = person('salary')
  File "/path/to/openfisca_core/entities.py", line 277, in __call__
    result = self.reference_entity(*args, **kwargs)
TypeError: __call__() takes at least 3 arguments (2 given)
```

This is not really explicit. We should catch this error and display a better message.

MattiSG:
The error message should be something along the lines of:

```
You requested computation of variable “salary”, but you did not specify on which period in "/path/to/file.py:157".
You must always specify the period as the second parameter. The convention is to call this parameter “period”: person('salary', period).
See more information at <https://doc.openfisca.fr/coding-the-legislation/index.html>.
```

(the documentation URL should be more precise, just given here as an example)

Implementation suggestion: add a default value that triggers the error.

PR #506, opened 2017-05-12
Improve the error when the period argument is forgotten

michelbl:
Connected to #495

### DIFF
--- a/openfisca_core/entities.py
+++ b/openfisca_core/entities.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
 
-import numpy as np
+import traceback
 import warnings
+
+import numpy as np
 
 from formulas import ADD, DIVIDE
 
@@ -51,8 +53,22 @@ class Entity(object):
         if role is not None and not type(role) == Role:
             raise Exception("{} is not a valid role".format(role))
 
-    def __call__(self, variable_name, period, options = [], **parameters):
+    def check_period_validity(self, variable_name, period):
+        if period is None:
+            stack = traceback.extract_stack()
+            filename, line_number, function_name, line_of_code = stack[-3]
+            raise ValueError(u'''
+You requested computation of variable "{}", but you did not specify on which period in "{}:{}":
+    {}
+When you request the computation of a variable within a formula, you must always specify the period as the second parameter. The convention is to call this parameter "period". For example:
+    computed_salary = person('salary', period).
+See more information at <https://doc.openfisca.fr/coding-the-legislation/35_periods.html#periods-for-variable>.
+'''.format(variable_name, filename, line_number, line_of_code).encode('utf-8'))
+
+    def __call__(self, variable_name, period = None, options = [], **parameters):
         self.check_variable_defined_for_entity(variable_name)
+
+        self.check_period_validity(variable_name, period)
 
         if ADD in options and DIVIDE in options:
             raise ValueError(u'Options ADD and DIVIDE are incompatible (trying to compute variable {})'.format(variable_name).encode('utf-8'))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 12.1.2
+
+* Improve the error when the period argument is forgotten in entity call
+
 ## 12.1.1
 
 * When replacing a variable by another one in a reform, assume the new variable has the same metadata than the reference one.
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ from setuptools import setup, find_packages
 
 setup(
     name = 'OpenFisca-Core',
-    version = '12.1.1',
+    version = '12.1.2',
     author = 'OpenFisca Team',
     author_email = 'contact@openfisca.fr',
     classifiers = [
--- a/tests/core/test_countries.py
+++ b/tests/core/test_countries.py
@@ -6,15 +6,46 @@ import numpy as np
 from nose.tools import raises, assert_raises
 
 from openfisca_core.variables import Variable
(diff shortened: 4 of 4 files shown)


openfisca_core/entities.py
```diff
@@ -51,9 +51,21 @@ def check_role_validity(self, role):
...
+        if period is None:
+            import traceback
+            stack = traceback.extract_stack()
+            filename, line_number, function_name, line_of_code = stack[-2]
+            raise ValueError(u'''
+You requested computation of variable "{}", but you did not specify on which period in "{}:{}":
+    {}
+You must always specify the period as the second parameter. The convention is to call this parameter "period". For example:
+    computed_salary = person('salary', period).
+See more information at <https://doc.openfisca.fr/coding-the-legislation/10_basic_example.html#basic-example>.
```
fpagnoux: As a legislation writer, this page doesn't tell me much about why I need to add a period as an argument. 
I'd reference [this page](https://doc.openfisca.fr/coding-the-legislation/35_periods.html), and maybe add a sentence there saying that within a formula, you must *always* specify a period when you compute another variable.

openfisca_core/entities.py
```diff
@@ -51,9 +51,21 @@ def check_role_validity(self, role):
         if role is not None and not type(role) == Role:
             raise Exception("{} is not a valid role".format(role))
 
-    def __call__(self, variable_name, period, options = [], **parameters):
+    def __call__(self, variable_name, period=None, options = [], **parameters):
         self.check_variable_defined_for_entity(variable_name)
 
+        if period is None:
+            import traceback
```
fpagnoux: We usually import everything on the top of the module, except when you really can't do it (circular import).

If you're worried about performances, see [these](http://stackoverflow.com/questions/1188640/good-or-bad-practice-in-python-import-in-the-middle-of-a-file) [discussions](http://stackoverflow.com/questions/128478/should-python-import-statements-always-be-at-the-top-of-a-module).

openfisca_core/entities.py
```diff
@@ -51,9 +51,21 @@ def check_role_validity(self, role):
         if role is not None and not type(role) == Role:
             raise Exception("{} is not a valid role".format(role))
 
-    def __call__(self, variable_name, period, options = [], **parameters):
+    def __call__(self, variable_name, period=None, options = [], **parameters):
```
fpagnoux: While PEP8 recommandes `x=y` in functions kwargs, we prefer `x = y` in OpenFisca for readability.

openfisca_core/entities.py
fpagnoux: For a better separation of concerns and to keep `__call__` readable, I'd handle the error in a separate function (`check_something`, or `handle_something`).

openfisca_core/entities.py
michelbl: ok

openfisca_core/entities.py
michelbl: ok

michelbl: ok

openfisca_core/entities.py
michelbl: I tried to take into account your remarks.

michelbl:
As we discussed, not showing the traceback is a choice that must be made globally, so this is too big for the current issue.

openfisca_core/entities.py
```diff
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+import traceback
+
```
fpagnoux: > No need for this line break

michelbl: https://www.python.org/dev/peps/pep-0008/#imports

> Imports should be grouped in the following order:
> 
> * standard library imports
> * related third party imports
> * local application/library specific imports
>
> You should put a blank line between each group of imports.

This rule should not be observed for openfisca-core ?

michelbl:
The documentation seems already quite clear to me : 
> Most of the quantities calculated in openfisca can change over time. Therefore, each formula calculates a variable for a person (or a family, etc.) for a given period.
>
> This period is always the second argument of the formulas:

I added the anchor `periods-for-variable` https://doc.openfisca.fr/coding-the-legislation/35_periods.html#periods-for-variable

openfisca_core/entities.py
cbenz: Yes, this rule is observed in the OpenFisca-Core codebase, at least.

openfisca_core/entities.py
```diff
@@ -51,9 +53,23 @@ def check_role_validity(self, role):
         if role is not None and not type(role) == Role:
             raise Exception("{} is not a valid role".format(role))
 
-    def __call__(self, variable_name, period, options = [], **parameters):
+    def check_period_validity(self, variable_name, period):
+        if period is None:
+            stack = traceback.extract_stack()
+            filename, line_number, function_name, line_of_code = stack[-3]
+            raise ValueError(u'''
```
cbenz: In Python 2 exception message must be encoded (in UTF-8)

`raise ValueError(u'xxx {}'.format(yyy).encode('utf-8'))`

(If one of the variables mentioned in `.format` contains unicode characters, Python 2 won't take the liberty to choose UTF-8 encoding, and this will raise a `UnicodeEncodeError` which has nothing to do with the original `ValueError`)

michelbl: ok

openfisca_core/entities.py
fpagnoux: I may have mis-read the code, I thought there was a line break between two standard library imports 👌.
Yep, we should follow these rules.

fpagnoux:
@michelbl i'll let you rebase and squash before merging :)

PR #509, opened 2017-05-17
Document preview API routes

fpagnoux:
Connected to openfisca/openfisca-web-api#129

### DIFF
--- a/doc/source/parameter.rst
+++ b/doc/source/parameter.rst
@@ -1,0 +1,90 @@
+=============
+``/parameter``
+=============
+
+.. http:get:: /parameter/<parameter_id>
+
+    Get data about the ``<parameter_id>`` parameter
+
+    :>json description: Parameter description.
+    :>json id: Parameter id.
+    :>json values: Parameter values history. Only present if the parameter is a simple one. A ``null`` value means that the parameter does not exist anymore.
+    :>json brackets: Marginal scale history. Only present if the parameter is a marginal scale. A ``null`` value means that the scale does not exist anymore.
+    :resheader Country-Package: The name of the country package served by the API.
+    :resheader Country-Package-Version: The version of the country package served by the API.
+    :statuscode 200: no error.
+    :statuscode 404: the parameter ``<parameter_id>`` doesn't exist.
+
+    **Example: Simple parameter**
+
+        **Request**
+
+            .. sourcecode:: http
+
+                GET /parameter/taxes.income_tax_rate HTTP/1.1
+
+        **Response**:
+
+            .. sourcecode:: http
+
+                HTTP/1.1 200 OK
+                Country-Package: openfisca-france
+                Country-Package-Version: 18.2.1
+                Content-Type: application/json
+
+                {
+                  "description": "Income tax rate",
+                  "id": "taxes.income_tax_rate",
+                  "values": {
+                    "2012-01-01": 0.16,
+                    "2013-01-01": 0.13,
+                    "2014-01-01": 0.14,
+                    "2015-01-01": 0.15
+                  }
+                }
+
+    **Example: Marginal scale**
+
+        **Request**
+
+            .. sourcecode:: http
+
+                GET /parameter/taxes.social_security_contribution HTTP/1.1
+
+        **Response**:
+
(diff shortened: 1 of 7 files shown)


fpagnoux:
Preview visible [here](http://openfisca.readthedocs.io/en/add-new-api-doc/)

doc/source/parameter.rst
```diff
@@ -0,0 +1,90 @@
...
+=============
+
+.. http:get:: /parameter/<parameter_id>
+
+    Get data about the ``<parameter_id>`` parameter
+
+    :>json description: Parameter description.
+    :>json id: Parameter id.
+    :>json values: Parameter values history. Only present if the parameter is a simple one. A ``null`` value means that the parameter does not exist anymore.
+    :>json brackets: Marginal scale history. Only present if the parameter is a marginal scale. A ``null`` value means that the scale does not exist anymore.
```
MattiSG: > A ``null`` value means that the parameter does not exist anymore.

This is misleading: `values` can never be null.

Also, I see nowhere documented the types of these properties: `description` and `id` are `String`, while `values` and `brackets` are `Object`…

doc/source/parameter.rst
```diff
@@ -0,0 +1,90 @@
...
+    :resheader Country-Package: The name of the country package served by the API.
+    :resheader Country-Package-Version: The version of the country package served by the API.
+    :statuscode 200: no error.
```
MattiSG: “The requested parameter description is sent back in the response body.”

doc/source/parameters.rst
```diff
@@ -0,0 +1,44 @@
...
+``/parameters``
+==============
+
+.. http:get:: /parameters
+
+    All the parameters contained in the tax and benefit system
+
+    :resheader Country-Package: The name of the country package served by the API
+    :resheader Country-Package-Version: The version of the country package served by the API
+    :statuscode 200: no error
```
MattiSG: “The list of all parameters is sent back in the response body.”

doc/source/parameters.rst
```diff
@@ -0,0 +1,44 @@
+==============
+``/parameters``
+==============
+
+.. http:get:: /parameters
+
+    All the parameters contained in the tax and benefit system
```
MattiSG: Ok, but I could mostly infer this with the name. What I'm looking for in the doc is the response format. I need to know that I will get an object whose keys will be parameters IDs and values objects with a single property, “description”, which is a String, and that I can use the ID to navigate to more information with the `parameter` endpoint.

doc/source/parameter.rst
```diff
@@ -0,0 +1,90 @@
+=============
+``/parameter``
+=============
+
+.. http:get:: /parameter/<parameter_id>
+
+    Get data about the ``<parameter_id>`` parameter
```
MattiSG: “data about” could be inferred with the name of the endpoint. “all information contained in the loaded tax and benefit system about” at least tells me that I will get a full dump. Maybe “the type, description and values”?

doc/source/variable.rst
```diff
@@ -0,0 +1,59 @@
...
+=============
+
+.. http:get:: /variable/<variable_id>
+
+    Get data about the ``<variable_id>`` variable
+
+    :>json defaultValue: Variable default value.
+    :>json definitionPeriod: Definition period of the variable. Possible values : ``YEAR``, ``MONTH``, ``ETERNITY``.
+    :>json description: Variable description.
+    :>json entity: Entity for which the variable is defined. For instance, ``person``, ``household``…
```
MattiSG: Link to the reference manual.

doc/source/variable.rst
```diff
@@ -0,0 +1,59 @@
...
+``/variable``
+=============
+
+.. http:get:: /variable/<variable_id>
+
+    Get data about the ``<variable_id>`` variable
+
+    :>json defaultValue: Variable default value.
+    :>json definitionPeriod: Definition period of the variable. Possible values : ``YEAR``, ``MONTH``, ``ETERNITY``.
+    :>json description: Variable description.
```
MattiSG: All properties need types (see comments on `parameter`).

PR #511, opened 2017-05-17
Fix naming conflict between the official and preview API

fpagnoux:
Connected to openfisca/openfisca-web-api#129


fpagnoux:
This is already in prod as `12.1.4.dev2`.

PR #516, opened 2017-05-23
Handle parallel testing

fpagnoux:
Connected to openfisca/openfisca-france#734

#### New features

* Use `nose` in the `openfisca-run-test` script
  - This avoids boilerplate code on country packages, and makes parallelism easier to set on Circle CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 12.2.0
+
+* Improve the `openfisca-run-test` script
+  - Add a `--nose` option to run the tests with nose.
+    - This avoids boilerplate code on country packages, and makes parallelism easier to set on Circle CI.
+
 ## 12.1.4
 
 * Fix package naming conflict between the preview API and the official one.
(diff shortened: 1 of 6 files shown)


fpagnoux:
Good point 👍 .
It's indeed better to keep the parallelism business out of the core if we can.
I found an alternative implementation.

I still need the `--nose` option though. This options allows me to run OpenFisca YAML tests cases with nose (CI-friendly output, no failfast) without a boilerplate `test_yaml.py`. This has two advantages:
- Less boilerplate
- More important, I can do `openfisca-run-test --nose fileA.yaml fileB.yaml path/to/file.yaml`. This allows me to [rely on Circle CI native parallelism features](https://github.com/openfisca/openfisca-france/blob/circle-parallel/circle.yml#L22).

fpagnoux:
> But couldn't we keep this option as the default and only one? When is it useful not to use this option?

Good question. The only drawback I see is that if running with `nose` is the only option, then
     - Either `nosetests` becomes a dependency of `openfisca-core`
     - Or we keep it as a test dependency, but anyone who wants to run `YAML` tests must  `pip install openfisca_xxx[test]`

I guess the second option is ok.

Issue #514, opened 2017-05-18
Document the preview API using OpenAPI


fpagnoux:
As a developper using the web API, 
I can easily know what output format I should expect

MattiSG:
- [ ] Serve the API documentation from the Legislation Explorer.

MattiSG:
Use http://swagger.io.
Look in http://swagger.io/open-source-integrations/ for possible Flask integrations.
Once the description has been written, use http://swagger.io/swagger-ui/ to consume and visualise the description.

Anna-Livia:
To serve the specification from the flask app, you can try :
Changing :
`app = Flask(__name__)`
into 
`app = Flask(__name__, static_url_path='')`
Adding :
`@app.route('/openFiscaAPI.yml')
def root():
    return app.send_static_file('openFiscaAPI.yml')`
in `app.py`
Adding the `openFiscaAPI.yml`file in a directory called `static`

Display OpenAPI spec with Swagger UI :
I put the swaggerUI files (html, css, js) in a `openAPI_interface`branch  on the `legislation-explorer`repo. To use it, fill in the URL field in the`index.html`file.

PR #518, opened 2017-05-31
Add an OpenAPI route

fpagnoux:
Connected to #514.

Serve the OpenAPI specification under the API route `/swagger/`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 12.3.0
 
-* Serve the [OpenAPI specification](https://www.openapis.org/) under the API route `/swagger/`
+* Serve the [OpenAPI specification](https://www.openapis.org/) under the API route `/spec/`
 
 ## 12.2.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 12.3.0
+
+* Serve the [OpenAPI specification](https://www.openapis.org/) under the API route `/swagger/`
+
 ## 12.2.0
 
 * Use `nose` in the `openfisca-run-test` script
--- a/openfisca_web_api_preview/app.py
+++ b/openfisca_web_api_preview/app.py
@@ -44,8 +44,8 @@ def create_app(country_package = os.environ.get('COUNTRY_PACKAGE')):
             raise abort(404)
         return jsonify(variable)
 
-    @app.route('/swagger')
-    def get_swagger():
+    @app.route('/spec')
+    def get_spec():
         return jsonify(data['openAPI_spec'])
 
     @app.after_request
--- a/openfisca_web_api_preview/app.py
+++ b/openfisca_web_api_preview/app.py
@@ -1,10 +1,13 @@
 # -*- coding: utf-8 -*-
 
+import yaml
 import os
 from flask import Flask, jsonify, abort
 from flask_cors import CORS
 
 from loader import build_data
+
+OPEN_API_CONFIG_FILE = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'openAPI.yml')
 
 
 def create_app(country_package = os.environ.get('COUNTRY_PACKAGE')):
@@ -43,6 +46,16 @@ def create_app(country_package = os.environ.get('COUNTRY_PACKAGE')):
         if variable is None:
             raise abort(404)
         return jsonify(variable)
+
+    @app.route('/swagger')
+    def root():
+        file = open(OPEN_API_CONFIG_FILE, 'r')
+        spec = yaml.load(file)
+        country_package_name = data['country_package_metadata']['name'].title()
+        spec['info']['title'] = spec['info']['title'].replace("{CONTRY_PACKAGE_NAME}", country_package_name)
+        spec['info']['description'] = spec['info']['description'].replace("{CONTRY_PACKAGE_NAME}", country_package_name)
+        spec['host'] = os.environ.get('SERVER_NAME')
+        return jsonify(spec)
 
     @app.after_request
     def apply_headers(response):
--- a/openfisca_web_api_preview/app.py
+++ b/openfisca_web_api_preview/app.py
@@ -1,13 +1,10 @@
 # -*- coding: utf-8 -*-
 
-import yaml
 import os
 from flask import Flask, jsonify, abort
 from flask_cors import CORS
 
 from loader import build_data
-
-OPEN_API_CONFIG_FILE = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'openAPI.yml')
 
 
 def create_app(country_package = os.environ.get('COUNTRY_PACKAGE')):
@@ -48,14 +45,8 @@ def create_app(country_package = os.environ.get('COUNTRY_PACKAGE')):
         return jsonify(variable)
 
     @app.route('/swagger')
-    def root():
-        file = open(OPEN_API_CONFIG_FILE, 'r')
-        spec = yaml.load(file)
-        country_package_name = data['country_package_metadata']['name'].title()
-        spec['info']['title'] = spec['info']['title'].replace("{CONTRY_PACKAGE_NAME}", country_package_name)
-        spec['info']['description'] = spec['info']['description'].replace("{CONTRY_PACKAGE_NAME}", country_package_name)
-        spec['host'] = os.environ.get('SERVER_NAME')
-        return jsonify(spec)
+    def get_swagger():
+        return jsonify(data['openAPI_spec'])
 
     @app.after_request
     def apply_headers(response):
--- a/openfisca_web_api_preview/openAPI.yml
+++ b/openfisca_web_api_preview/openAPI.yml
@@ -1,7 +1,7 @@
 swagger: "2.0"
 info:
-  title: "OpenFisca France"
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 12.2.0
+
+* Serve the OpenAPI specification under the API route `/swagger/`
```
MattiSG: Add a link to the OpenAPI website.

openfisca_web_api_preview/app.py
```diff
@@ -44,6 +46,16 @@ def get_variable(id):
...
 
+    @app.route('/swagger')
+    def root():
+        file = open(OPEN_API_CONFIG_FILE, 'r')
+        spec = yaml.load(file)
+        country_package_name = data['country_package_metadata']['name'].title()
+        spec['info']['title'] = spec['info']['title'].replace("{CONTRY_PACKAGE_NAME}", country_package_name)
+        spec['info']['description'] = spec['info']['description'].replace("{CONTRY_PACKAGE_NAME}", country_package_name)
+        spec['host'] = os.environ.get('SERVER_NAME')
+        return jsonify(spec)
```
MattiSG: Shouldn't this be cached on startup? Here, it seems we're re-reading the file from disk on each request.

openfisca_web_api_preview/openAPI.yml
```diff
@@ -0,0 +1,253 @@
+swagger: "2.0"
+info:
+  title: "{CONTRY_PACKAGE_NAME} Web API"
```
MattiSG: Typo (everywhere): this is `COUNTRY_PACKAGE_NAME` (missing `U`).

openfisca_web_api_preview/openAPI.yml
```diff
@@ -0,0 +1,253 @@
...
+  description: "The OpenFisca Web API lets you get up-to-date information and formulas pertaining to the {CONTRY_PACKAGE_NAME} legislation e.g. minimum wage, state scholasrship, social benefits, income taxes…"
```
MattiSG: I don't think we should give these examples here: the country package name might not provide these elements. If we don't know how to explain better, then at least we should make it clearer that these are illustrative and might not be provided.

openfisca_web_api_preview/openAPI.yml
```diff
@@ -0,0 +1,253 @@
...
+    email: "contact@openfisca.fr"
+  license:
+    name: "AGPL"
+    url: "https://www.gnu.org/licenses/agpl-3.0.html"
+host: null
+schemes:
+  - "https"
+tags:
+- name: "Parameters"
+  description: "A parameter is a numeric property of the legislation which can evolve over time."
```
MattiSG: Shouldn't “which” be “that”? I read “which” as referring to the legislation, not to the parameter.

openfisca_web_api_preview/openAPI.yml
```diff
@@ -0,0 +1,253 @@
...
+  externalDocs:
+    description: "Find out more"
```
MattiSG: > SEO & clarity: “Parameter official documentation” or “Parameter reference” or “OpenFisca paramater reference”.

openfisca_web_api_preview/openAPI.yml
```diff
@@ -0,0 +1,253 @@
...
+            Country-Package-Version:
+              description: "The version of the country package served by the API"
+              type: "string"
+          schema:
+            $ref: '#/definitions/Parameters'
+  /parameter/{parameterID}:
+    get:
+      tags:
+      - "Parameters"
+      summary: "Gets information about a specific parameter"
```
MattiSG: Conjugation inconsistency: the previous `summary` uses infinitive. I'm in favour of infinitive, as it is not the endpoint that _gets information_, but the user.

openfisca_web_api_preview/openAPI.yml
```diff
@@ -0,0 +1,253 @@
...
+      operationId: "getParameter"
+      produces:
+      - "application/json"
+      parameters:
+      - name: "parameterID"
+        in: "path"
+        description: "ID of parameter (e.g. impot_revenu.bareme)"
```
MattiSG: This example is misleading, as this country package might not provide it, and the user who would try to use this example would face a 404 and not understand what went wrong. Explain that these IDs can be obtained by enumerating the `/parameters` endpoint.

openfisca_web_api_preview/openAPI.yml
```diff
@@ -0,0 +1,253 @@
...
+        required: true
+        type: "string"
+      responses:
+        200:
+          description: "The requested parameter's information is sent back in the response body"
+          headers :
+            Country-Package :
+              description : "The name of the country package served by the API"
```
MattiSG: > “served by the API” is very definitive. “currently loaded in this API server”?

openfisca_web_api_preview/openAPI.yml
```diff
@@ -0,0 +1,253 @@
...
+            Country-Package-Version:
+              description: "The version of the country package served by the API"
+              type: "string"
+          schema:
+            $ref: "#/definitions/Variables"
+  /variable/{variableID}:
+    get:
+      tags:
+      - "Variables"
+      summary: "Gets information about a specific variable"
```
MattiSG: Conjugation inconsistency, see same comment for `/parameter`.

openfisca_web_api_preview/openAPI.yml
```diff
@@ -0,0 +1,253 @@
...
+      operationId: "getVariable"
+      produces:
+      - "application/json"
+      parameters:
+      - name: "variableID"
+        in: "path"
+        description: "ID of a variable (e.g. aide_embauche_pme)"
```
MattiSG: This example is misleading, as this country package might not provide it, and the user who would try to use this example would face a 404 and not understand what went wrong. Explain that these IDs can be obtained by enumerating the `/variables` endpoint.

openfisca_web_api_preview/openAPI.yml
```diff
@@ -0,0 +1,253 @@
...
+        required: true
+        type: "string"
+      responses:
+        200:
+          description: "The requested parameter's information is sent back in the response body"
```
MattiSG: We're in the `variable` endpoint.

openfisca_web_api_preview/openAPI.yml
```diff
@@ -0,0 +1,253 @@
...
+            Country-Package :
+              description : "The name of the country package served by the API"
+              type: "string"
+            Country-Package-Version:
+              description: "The version of the country package served by the API"
+              type: "string"
+          schema:
+            $ref: "#/definitions/Variable"
+        404:
+          description: "The requested parameter does not exist"
```
MattiSG: Variable.

openfisca_web_api_preview/openAPI.yml
```diff
@@ -0,0 +1,253 @@
...
+        definition:
+          type: 'string'
+
+  Variable:
+    type: "object"
+    properties:
+      defaultValue:
+        type: "string"
+      definitionPeriod:
+        type: "string"
```
MattiSG: Is there no way to make the expected format more explicit?

openfisca_web_api_preview/openAPI.yml
```diff
@@ -0,0 +1,253 @@
...
+      reference :
+        type: "array"
+        items:
+          type : "string"
+      source :
+        type: "string"
+      valueType:
+         type: "string"
+    example:
+      defaultValue": 0,
```
MattiSG: Lost quote…?

openfisca_web_api_preview/openAPI.yml
```diff
@@ -0,0 +1,253 @@
...
+      definitionPeriod: "MONTH"
+      description: "Aide \u00e0 l'embauche d'un salari\u00e9 pour les PME"
```
MattiSG: Is this encoding normal?

fpagnoux:
Sorry there was an accidental force push since you reviewed.

openfisca_web_api_preview/app.py
```diff
@@ -44,6 +44,10 @@ def get_variable(id):
             raise abort(404)
         return jsonify(variable)
 
+    @app.route('/swagger')
```
MattiSG: What do you think about `/spec`, `/help` or `/manual`? `swagger` is a brand name, and it has since then be superseded by “OpenAPI”. This name still has the benefit of being the same as the current `/swagger` endpoint, but things have changed since then.

tests/web_api/test_openAPI.py
```diff
@@ -0,0 +1,19 @@
...
+from nose.tools import assert_equal, assert_items_equal
+from . import subject
+
+openAPI_response = subject.get('/swagger')
+
+
+def test_return_code():
+    assert_equal(openAPI_response.status_code, OK)
+
+
```
MattiSG: A test that just loads the JSON and thus checks it is valid would be welcome.

openfisca_web_api_preview/app.py
fpagnoux: 👍 for `/spec/`

tests/web_api/test_openAPI.py
fpagnoux: That's done in the next test ;)

PR #522, opened 2017-06-01
Merge Variables and DatedVariables

### LINKED ISSUE #371: Merge `Variable` and `DatedVariable`
See https://github.com/openfisca/openfisca-france/pull/376#issuecomment-172483898

When the decorator `dated_function` is used on a `Variable` like this:

``` python
class cehr(Variable):
    [...]

    @dated_function(start = date(2011, 1, 1), stop = date(2011, 12, 31))
        return period, [...]

    @dated_function(start = date(2012, 1, 1))
        return period, [...]
```

the Core should detect that it's a "dated variable" automatically.

sandcha:
#### Breaking changes

- In variables:
  - Remove SimpleVariable and DatedVariable. Any Variable acts like old DatedVariable
  - Remove start_date attribute
  - Rename stop_date attribute to end
- In formulas:
  - Remove SimpleFormula and DatedFormula. Any Formula acts like old DatedFormula
  - Remove @dated_function: start goes to formula name suffix, stop is deduced from other formulas start dates and variable end attribute 

#### New features

- Introduce `formula_YYYY[_MM[_DD]]`
  - Allows for formula start instant definition
- Introduce `end = 'YYYY-MM-DD'`
  - Allows for variable end date definition in string format
- Change ETERNITY period effect
  - Allows for dated Variable definition on ETERNITY period

#### Technical changes

- In variables:
  - Remove Abstract class
- In formulas:
  - Remove Abstract classes
- In tests:
  - Add test_variable

Fixes #371.


fpagnoux:
Tests are [passing in france](https://circleci.com/gh/openfisca/openfisca-france/134#tests/containers/3) 🎉

(Red only because the version has not been bumped)

openfisca_core/variables.py
```diff
@@ -34,7 +34,7 @@ def get_introspection_data(self, tax_benefit_system):
         return comments, source_file_path, source_code, start_line_number
 
     def to_column(self, tax_benefit_system):
-        formula_class = self.__class__.formula_class
+        formula_class = self.__class__.formula_class  # Formula
```
fpagnoux: Since you simplified everything, `formula_class` is always `Formula` !

We should: 
- remove the class attribute `formula_class = Formula`
- remove `formula_class` for `new_filled_column` arguments as it's always the same.

openfisca_core/variables.py
```diff
@@ -4,11 +4,11 @@
...
 import textwrap
 
-from openfisca_core.formulas import SimpleFormula, DatedFormula, new_filled_column
+from openfisca_core.formulas import Formula, new_filled_column
 
 
-class AbstractVariable(object):
-    formula_class = None
+class Variable(object):
+    formula_class = Formula
```
fpagnoux: Remove this class attribute (see comment below).

tests/core/test_variable.py
```diff
@@ -0,0 +1,771 @@
...
+
+def test_dates__dated_attribute__one_formula():
+    variable = tax_benefit_system.column_by_name['dated_attribute__one_formula']
+    assert variable is not None
+    assert variable.end == '1989-12-31'
+
+    assert variable.formula_class.dated_formulas_class.__len__() == 1
+    formula = variable.formula_class.dated_formulas_class[0]
+    assert formula is not None
+    assert formula['start_instant'].date == datetime.date(1, 1, 1)
```
fpagnoux: `datetime.date(1, 1, 1)` should be replaced by `datetime.date.min`.
If tomorrow `datetime` is adapted to handle the egyptian era , we don't want to impose ourselves to stop at `0001-01-01` :)

openfisca_core/variables.py
sandcha: formula_class removed from Variable but maintained as item of Formula.dated_formulas_class

tests/core/test_variable.py
```diff
@@ -0,0 +1,771 @@
...
+    definition_period = MONTH
+    label = u"Variable with dated attributes (including deprecated start_date), no function."
+    start_date = datetime.date(1980, 1, 1)  # Deprecated
+    end = '1989-12-31'
+
+
+add_variable_catch_assertion(tax_benefit_system, variable__deprecated_start_date, 'Deprecated "start_date" attribute in definition of variable')
+
+
+# 371 - end, no function
```
sandcha: As a lazy way to find history. Removed.

tests/core/test_variable.py
```diff
@@ -0,0 +1,771 @@
...
+
+
+tax_benefit_system.add_variable(variable__dated_attribute)
+
+
+def test_variable__dated_attribute():
+    variable = tax_benefit_system.column_by_name['variable__dated_attribute']
+    assert variable is not None
+    assert variable.end == '1989-12-31'
+    assert variable.formula_class.dated_formulas_class.__len__() == 0
```
sandcha: It's not about testing all combinations : an end date changes how the validity of a function is estimated. But, it's tested elsewhere so ok for removing it.

tests/core/test_variable.py
sandcha: Ok for tests. And it will show us the egyptian era case when it comes as `datetime.date.min` is not applied in `formulas.py` (to be consistent with formula name specifications).

openfisca_core/formulas.py
```diff
@@ -663,6 +620,50 @@ def get_neutralized_column(column):
...
+    formula_default_month = '01'
+    formula_default_day = '01'
+
+    if formula_name == formula_name_prefix:
+        formula_name += formula_name_separator + formula_default_year + formula_name_separator\
+            + formula_default_month + formula_name_separator\
+            + formula_default_day
+
+    else:
+        match = re.match(r'(formula_)(\d{4}(_\d{2}){0,2})', formula_name)  # YYYY or YYYY_MM or YYYY_MM_DD
```
fpagnoux: Also, are the parenthesis around `formula_` necessary ?

openfisca_core/formulas.py
```diff
@@ -663,6 +620,50 @@ def get_neutralized_column(column):
...
+
+
+def get_datetime_date(variable_name, date_str):
+    """
+    Return a datetime.date from a date string.
+    """
+    try:
+        time = datetime.datetime.strptime(date_str, '%Y-%m-%d')
+    except ValueError:
+        raise ValueError(u"Incorrect 'end' attribute format in '{}'. 'YYYY-MM-DD' expected where YYYY, MM and DD are year, month and day. Found: {}".format(variable_name, date_str).encode('utf-8'))
```
fpagnoux: Why does a function named `get_datetime_date` know about an `end` attribute of a variable ? This error should be thrown directly by `new_filled_column`.

openfisca_core/formulas.py
```diff
@@ -774,15 +766,11 @@ def new_filled_column(
...
-
-    if stop_date is UnboundLocalError:
-        stop_date = None if reference_column is None else reference_column.end
-    elif stop_date is not None:
-        assert isinstance(stop_date, datetime.date)
+    if end is UnboundLocalError:
+        end = None if reference_column is None else reference_column.end
+    elif end is not None:
+        assert isinstance(end, str)
+        get_datetime_date(name, end)  # Raises ValueError on format error.
```
fpagnoux: Calling `get_something` just to check if it throws an error is not great.

Why not storing the parsed date in the column ? 

```py
end = get_datetime_date(end)
```

fpagnoux: Or even, removing the helper `get_datetime_date` and validating the attribute here.

```py
try:
     end = datetime.datetime.strptime(end, '%Y-%m-%d')
except ValueError:
     raise ValueError(u"Incorrect 'end' attribute format in '{}'. 'YYYY-MM-DD' expected where YYYY, MM and DD are year, month and day. Found: {}".format(variable_name, date_str).encode('utf-8'))
```

openfisca_core/formulas.py
```diff
@@ -834,110 +823,71 @@ def new_filled_column(
...
+    formula_name_separator = '_'
+    dated_formulas_class = []
+    for function_name, function in specific_attributes.copy().iteritems():
+        if not function_name.startswith(formula_name_prefix):
+            # Current attribute isn't a formula
+            continue
+
+        formula_start_date = extract_formula_date_from_name(function_name, formula_name_prefix, formula_name_separator)
+        if end is not None:
+            end_date = get_datetime_date(name, end)
```
fpagnoux: You're parsing the string `n` times. See above the suggestion to parse it once and for all.

openfisca_core/formulas.py
```diff
@@ -834,110 +823,71 @@ def new_filled_column(
...
-                            'stop_instant']:
-                        break
-                else:
-                    dated_formulas_class.append(reference_dated_formula_class)
-            dated_formulas_class.sort(key = lambda dated_formula_class: dated_formula_class['start_instant'])
+    # Turn function into a dated function
+    formula_name_prefix = 'formula'
+    formula_name_separator = '_'
+    dated_formulas_class = []
+    for function_name, function in specific_attributes.copy().iteritems():
```
fpagnoux: Why are we using  `copy()` ?

openfisca_core/formulas.py
```diff
@@ -834,110 +823,71 @@ def new_filled_column(
...
+        del specific_attributes[function_name]
+        dated_formulas_class.append(dict(
+            formula_class = dated_formula_class,
+            start_instant = periods.instant(formula_start_date),
+            ))
 
-        formula_class_attributes['dated_formulas_class'] = dated_formulas_class
-    else:
-        assert issubclass(formula_class, SimpleFormula), formula_class
+    # Sort dated formulas by start instant.
```
fpagnoux: The code is pretty explicit. Why adding this extra comment ?

openfisca_core/formulas.py
```diff
@@ -834,110 +823,71 @@ def new_filled_column(
...
+    dated_formulas_class.sort(key = lambda dated_formula_class: dated_formula_class['start_instant'])
+
+    # Add dated formulas defined in (optional) reference column when they are not overridden by new dated
+    # formulas.
```
fpagnoux: Why splitting this line ? Most editors today can deal with line wrapping.

openfisca_core/formulas.py
```diff
@@ -834,110 +823,71 @@ def new_filled_column(
...
+    if reference_column is not None and issubclass(reference_column.formula_class, Formula):
```
fpagnoux: `issubclass(reference_column.formula_class, Formula)` is true by construction. Testing it is useless.

openfisca_core/formulas.py
```diff
@@ -834,110 +823,71 @@ def new_filled_column(
...
 
-        function = specific_attributes.pop('function', None)
-        if column.definition_period == ETERNITY:
-            assert function is None
-        if reference_column is not None and function is None:
-            function = reference_column.formula_class.function
-        formula_class_attributes['function'] = function
+    formula_class_attributes['dated_formulas_class'] = dated_formulas_class
 
     # Ensure that all attributes defined in ConversionColumn class are used.
```
fpagnoux: Nothing to do with your PR, but this line is outdated (`ConversionColumn` do not exist), we can remove it.

openfisca_core/formulas.py
```diff
@@ -834,110 +823,71 @@ def new_filled_column(
...
+    assert not specific_attributes.get('start_date'), \
+        'Deprecated "start_date" attribute in definition of variable "{}".'.format(name)
```
fpagnoux: > 👏

openfisca_core/formulas.py
```diff
@@ -834,110 +823,71 @@ def new_filled_column(
...
     column.entity = entity
     column.formula_class = formula_class
     column.definition_period = definition_period
     column.is_neutralized = is_neutralized
     column.label = label
     column.law_reference = law_reference
     column.name = name
-    if start_date is not None:
-        column.start = start_date
+
```
fpagnoux: This line break is not necessary.

tests/core/test_variable.py
```diff
@@ -0,0 +1,472 @@
...
+    definition_period = MONTH
+    label = u"Variable without date."
+
+
+tax_benefit_system.add_variable(variable__no_date)
+
+
+def test_variable__no_dated_attributes():
+    variable = tax_benefit_system.column_by_name['variable__no_date']
+    assert variable is not None
```
fpagnoux: This can't be `None`, otherwise `variable = tax_benefit_system.column_by_name['variable__no_date']` would have thrown an error.

tests/core/test_variable.py
```diff
@@ -0,0 +1,472 @@
...
+        else:
+            raise
+    except:
+        raise
+
+
+# TESTS
+
+
+# NO DATED ATTRIBUTE(S) - NO DATED FORMULA
```
fpagnoux: `No end attribute` would be more explicit.

tests/core/test_variable.py
```diff
@@ -0,0 +1,472 @@
...
+
+
+# DATED ATTRIBUTE(S) - NO DATED FORMULA
+
+
+class variable__deprecated_start_date(Variable):
+    column = IntCol
+    entity = Person
+    definition_period = MONTH
+    label = u"Variable with dated attributes (including deprecated start_date), no function."
```
fpagnoux: The concept of "dated attribute" is not necessary. Variables may have and `end` date, and `formulas`, that's enough.

tests/core/test_variable.py
```diff
@@ -0,0 +1,472 @@
...
+    start_date = datetime.date(1980, 1, 1)  # Deprecated
+    end = '1989-12-31'
+
+
+check_error_at_add_variable(tax_benefit_system, variable__deprecated_start_date, 'Deprecated "start_date" attribute in definition of variable')
+
+
+# end, no function
```
fpagnoux: no formula.

It's important to be precise about our domain language. When a new developper comes in OpenFisca, they may be confused : is a function the same than a formula ?

tests/core/test_variable.py
```diff
@@ -0,0 +1,472 @@
...
+
+class variable__strange_end_attribute(Variable):
+    column = IntCol
+    entity = Person
+    definition_period = MONTH
+    label = u"Variable with dubious end attribute, no function."
```
fpagnoux: no formula

tests/core/test_variable.py
```diff
@@ -0,0 +1,472 @@
...
+    except:
+        raise
+
+    # Check that Error at variable adding prevents it from registration in the taxbenefitsystem.
+    assert not tax_benefit_system.column_by_name.get('variable__strange_end_attribute')
+
+
+# end, no function
+
+class variable__dated_attribute(Variable):
```
fpagnoux: `variable__end_attribute`

tests/core/test_variable.py
```diff
@@ -0,0 +1,472 @@
...
+
+def test_variable__dated_attribute():
+    variable = tax_benefit_system.column_by_name['variable__dated_attribute']
+    assert variable is not None
+    assert variable.end == '1989-12-31'
+
+
+# end, one function without date
+
+class dated_attribute__one_simple_formula(Variable):
```
fpagnoux: `end_attribute...`

tests/core/test_variable.py
```diff
@@ -0,0 +1,472 @@
...
+    assert simulation.calculate('dated_attribute__one_simple_formula', month) == 100
+
+    month = '1990-01'
+    simulation = new_simulation(tax_benefit_system, month)
+    assert simulation.calculate('dated_attribute__one_simple_formula', month) == IntCol.default
+
+
+def test_dates__dated_attribute__one_simple_formula():
+    variable = tax_benefit_system.column_by_name['dated_attribute__one_simple_formula']
+    assert variable is not None
```
fpagnoux: Same than above, this cannot be `None`

tests/core/test_variable.py
```diff
@@ -0,0 +1,472 @@
...
+    assert variable.end == '1989-12-31'
+
+    assert len(variable.formula_class.dated_formulas_class) == 1
+    formula = variable.formula_class.dated_formulas_class[0]
+    assert formula is not None
```
fpagnoux: Could this really be `None` ?

tests/core/test_variable.py
```diff
@@ -0,0 +1,472 @@
...
+
+    i = 0
+    formula = variable.formula_class.dated_formulas_class[i]
+    assert formula is not None
+    assert formula['start_instant'] is not None
+    assert formula['start_instant'].date == datetime.date(2000, 1, 1)
+
+    i = 1
+    formula = variable.formula_class.dated_formulas_class[i]
+    assert formula is not None
```
fpagnoux: Same than above: can this ever be `None` ?

tests/core/test_variable.py
```diff
@@ -0,0 +1,472 @@
...
+    check_error_at_add_variable(tax_benefit_system, no_attribute__formulas__different_names__dates_overlap, "Dated formulas overlap")
+
+
+# function(start), different names, no date overlap
+
+class no_attribute__formulas__different_names__no_overlap(Variable):
+    column = IntCol
+    entity = Person
+    definition_period = MONTH
+    label = u"Variable, no and attribute, multiple dated functions with different names and no date overlap."
```
fpagnoux: Typo

CHANGELOG.md
```diff
@@ -1,5 +1,35 @@
 # Changelog
 
+# 13.0.0
```
fpagnoux: I copy-pasted what you wrote in the PR description to start discussing the Changelog :).

The general idea of the Changelog is to be written from the point of view of a OpenFisca Core user (ie someone who is writing their own legislation)

CHANGELOG.md
```diff
@@ -1,5 +1,35 @@
...
+
+#### Breaking changes
+
+- In variables:
+  - Remove SimpleVariable and DatedVariable. Any Variable acts like old DatedVariable
```
fpagnoux: I don't think there is currently anything named `SimpleVariable`. 

Suggestion: 

- Merge `Variable` and `DatedVariable`
    - `Variable` can now handle formula evolution over time.

CHANGELOG.md
```diff
@@ -1,5 +1,35 @@
...
+  - Remove start_date attribute
```
fpagnoux: Remove -> Deprecate and remove

CHANGELOG.md
```diff
@@ -1,5 +1,35 @@
...
+  - Rename stop_date attribute to end
```
fpagnoux: When you are quoting code (for instance `stop_date`), it's better to use backticks (`) to format it properly.

CHANGELOG.md
```diff
@@ -1,5 +1,35 @@
...
+- In formulas:
+  - Remove SimpleFormula and DatedFormula. Any Formula acts like old DatedFormula
+  - Remove @dated_function: start goes to formula name suffix, stop is deduced from other formulas start dates and variable end attribute
```
fpagnoux: I'dd add an example for one formula to explicit the change.

CHANGELOG.md
```diff
@@ -1,5 +1,35 @@
...
+- Change ETERNITY period effect
+  - Allows for dated Variable definition on ETERNITY period
+
+#### Technical changes
+
+- In variables:
+  - Remove Abstract class
+- In formulas:
+  - Remove Abstract classes
+- In tests:
```
fpagnoux: Adding tests is usually not something the users care about, so usually it's not necessary to mention it in the changelog.

CHANGELOG.md
```diff
@@ -1,5 +1,35 @@
...
+  - Allows for formula start instant definition
+- Introduce `end = 'YYYY-MM-DD'`
+  - Allows for variable end date definition in string format
+- Change ETERNITY period effect
+  - Allows for dated Variable definition on ETERNITY period
+
+#### Technical changes
+
+- In variables:
+  - Remove Abstract class
```
fpagnoux: This information is probably not needed in the Changelog:
- Users don't interact with these abstract classes.
- Developers can read in the diff that you removed these classes

CHANGELOG.md
```diff
@@ -1,5 +1,35 @@
...
+  - Remove @dated_function: start goes to formula name suffix, stop is deduced from other formulas start dates and variable end attribute 
+
+#### New features
+
+- Introduce `formula_YYYY[_MM[_DD]]`
+  - Allows for formula start instant definition
+- Introduce `end = 'YYYY-MM-DD'`
+  - Allows for variable end date definition in string format
+- Change ETERNITY period effect
+  - Allows for dated Variable definition on ETERNITY period
```
fpagnoux: Dated variables don't exist anymore, so mentioning them in the changelog is not great.

Suggestion : 

- Remove restriction that prevented to have formula changes over time for a variable with `ETERNITY` as a definition period.

openfisca_core/formulas.py
```diff
@@ -433,162 +481,61 @@ def compute(self, period, **parameters):
...
         self.clean_cycle_detection_data()
         if max_nb_cycles is not None:
             simulation.max_nb_cycles = None
 
-        return dated_holder
+        return holders.DatedHolder(self.holder, period, array, extra_params)
+
+    def find_function(self, period):
+        """
+        This function finds the last active formula for the time interval [period starting date, variable end attribute].
```
sandcha: Done (it was used to act like the python official documentation but it's not even consistent there).

CHANGELOG.md
```diff
@@ -1,5 +1,35 @@
...
+  - Remove SimpleVariable and DatedVariable. Any Variable acts like old DatedVariable
+  - Remove start_date attribute
+  - Rename stop_date attribute to end
+- In formulas:
+  - Remove SimpleFormula and DatedFormula. Any Formula acts like old DatedFormula
+  - Remove @dated_function: start goes to formula name suffix, stop is deduced from other formulas start dates and variable end attribute 
+
+#### New features
+
+- Introduce `formula_YYYY[_MM[_DD]]`
```
fpagnoux: This replaces the former `@dated_function` decorator, so it should be mentioned near the its deprecation.

As a user, it's better for me to find in the same place what syntax has disappeared, and what has replaced it.

CHANGELOG.md
```diff
@@ -1,5 +1,35 @@
...
+  - Allows for formula start instant definition
+- Introduce `end = 'YYYY-MM-DD'`
+  - Allows for variable end date definition in string format
```
fpagnoux: Same: should be put near `Rename stop_date attribute to end`

tests/core/test_variable.py
```diff
@@ -0,0 +1,472 @@
...
+class variable__deprecated_start_date(Variable):
+    column = IntCol
+    entity = Person
+    definition_period = MONTH
+    label = u"Variable with dated attributes (including deprecated start_date), no function."
+    start_date = datetime.date(1980, 1, 1)  # Deprecated
+    end = '1989-12-31'
+
+
+check_error_at_add_variable(tax_benefit_system, variable__deprecated_start_date, 'Deprecated "start_date" attribute in definition of variable')
```
fpagnoux: Should be in a test

fpagnoux: Fixed it

openfisca_core/formulas.py
```diff
@@ -433,162 +481,61 @@ def compute(self, period, **parameters):
...
+        """
+        end = self.holder.column.end
+        if end and period.start.date > get_datetime_date(self, end):
+            return None
+
+        # Assume that self.dated_formulas and self.dated_formulas_class
+        # contain same formulas (instance and class) in same order
+        assert len(self.dated_formulas) == len(self.dated_formulas_class)
```
sandcha: If later a developer changes only `self.dated_formulas` or only`self.dated_formulas_class`, an error should be thrown to express that this important assumption is broken. Besides, it's the minimal assertion.

openfisca_core/formulas.py
```diff
@@ -663,6 +620,50 @@ def get_neutralized_column(column):
...
+    Valid dated name formats are : 'formula', 'formula_YYYY', 'formula_YYYY_MM' and 'formula_YYYY_MM_DD' where YYYY, MM and DD are a year, month and day.
+
+    Default year is '0001'. Default month and day are '01'.
+    """
+    formula_default_year = '0001'
+    formula_default_month = '01'
+    formula_default_day = '01'
+
+    if formula_name == formula_name_prefix:
+        formula_name += formula_name_separator + formula_default_year + formula_name_separator\
```
sandcha: But flake8 says "W503 line break before binary operator". So, no brackets and backslashes won the contest.

openfisca_core/formulas.py
sandcha: Removed after next lines update.

openfisca_core/formulas.py
sandcha: `end` removed but ` get_datetime_date` called 2 times and some psychological issue with adding new lines to `new_filled_column`.

openfisca_core/formulas.py
sandcha: To avoid adding dependencies between `new_filled_column` sections it seems better to avoid declaring an end_date there. 
And, it is better to extract this code in a helper because the same instruction has to be called 2 times for 2 different things (checking the attribute format and giving a date time date from a string).

openfisca_core/formulas.py
sandcha: Because there is `del specific_attributes[function_name]` in the loop.

openfisca_core/formulas.py
sandcha: Not parsed for same reason: the expected output is different. 
And this imperfect solution helps `new_filled_column` diet.

openfisca_core/formulas.py
sandcha: To catch the eye :smile: . Removed.

tests/core/test_variable.py
sandcha: Done everywhere in the file but on this Variable as `variable__deprecated_start_date` has `start_date` and `end`.

tests/core/test_variable.py
sandcha: Not really. Removed in file.

CHANGELOG.md
sandcha: Ok for second line and the fanciful SimpleVariable.
In first line, what about AbstractVariable ? It was possible to create its own kind of variable before and we removed this service.

CHANGELOG.md
sandcha: Does a real deprecated attribute raise a warning and is still managed or does it raise a fatal error ?

CHANGELOG.md
sandcha: Or one example with variable class definition and formulas to show linked changes?

CHANGELOG.md
sandcha: It seems to me that it's more a service that disappears than a technical change. So, should we really silence this information ?

openfisca_core/formulas.py
```diff
@@ -433,162 +483,57 @@ def compute(self, period, **parameters):
...
+
+    def find_function(self, period):
+        """
+        Finds the last active formula for the time interval [period starting date, variable end attribute].
+        """
+        end = self.holder.column.end
+        if end and period.start.date > get_datetime_date(self, end):
+            return None
+
+        i = len(self.dated_formulas)
```
fpagnoux: This variable is not use anywhere anymore ;)

openfisca_core/formulas.py
```diff
@@ -663,6 +618,61 @@ def get_neutralized_column(column):
...
+    Valid dated name formats are : 'formula', 'formula_YYYY', 'formula_YYYY_MM' and 'formula_YYYY_MM_DD' where YYYY, MM and DD are a year, month and day.
+
+    Default year is '0001'. Default month and day are '01'.
+    """
+    formula_name_prefix = 'formula'
+    if not attribute_name.startswith(formula_name_prefix):
+        # Current attribute isn't a formula
+        return None
+
+    formula_name_separator = '_'
```
fpagnoux: Why not declare all these constants at the same place than `formula_name_prefix`?

openfisca_core/formulas.py
fpagnoux: `W503` is [highly controversial](https://gitlab.com/pycqa/flake8/issues/139) and [outdated](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator).

I'm black-listing it right now.

fpagnoux: Switched to implicit line continuation

openfisca_core/formulas.py
```diff
@@ -663,6 +618,63 @@ def get_neutralized_column(column):
...
+
+        if len(start_str) == 7:  # YYYY_MM
+            start_str += formula_name_separator + formula_default_day
+            formula_name += formula_name_separator + formula_default_day
+
+    return datetime.datetime.strptime(formula_name,
+    formula_name_prefix + formula_name_separator + '%Y_%m_%d').date()
+
+
+def get_datetime_date(src_name, date_str):
```
fpagnoux: Akward signature.

openfisca_core/formulas.py
fpagnoux: I don't see how you would need to call this helper several times if you just parsed the string once and for all.
Careful, here, we are parsing the `end` attribute _every single time_ we call `find_function`. 

It seems much more natural to parse the input only once. The fact that we end end using a method named `get_datetime_date` just to raise an error is a clear sign that our code structure is lopsided.

openfisca_core/formulas.py
fpagnoux: Good catch 👍

openfisca_core/formulas.py
fpagnoux: If you want to make `new_filled_column ` smaller, you can split it into functions that make sense 😉 .

This doesn't.

tests/core/test_variable.py
```diff
@@ -0,0 +1,548 @@
...
+
+
+# END ATTRIBUTE - NO DATED FORMULA
+
+
+class variable__deprecated_start_date(Variable):
+    column = IntCol
+    entity = Person
+    definition_period = MONTH
+    label = u"Variable with dated attributes (including deprecated start_date), no formula."
```
fpagnoux:
```py
u"Variable with and end attribute and a deprecated start_date, no formula."
```

CHANGELOG.md
fpagnoux: I don't think `AbstractVariable` was ever meant to be extended by child classes in country packages 🤔.
I don't think anyone ever did it anyway, so it's ok not to mention it.

CHANGELOG.md
fpagnoux: Good question.
If we just "deprecate" something, I'd assume that it's still available and raise a waring.
If we "deprecate and remove" something, I'd assume it will raise a fatal error.
If we "remove" something, I'm just afraid it's not crystal clear that what we removed is a _feature_.

@MattiSG @michelbl @Anna-Livia thoughts ?

CHANGELOG.md
```diff
@@ -1,5 +1,64 @@
...
+  - Rename `stop_date` attribute to `end`
+    - Introduce end string format `end = 'YYYY-MM-DD'`
+- In formulas:
+  - Merge `SimpleFormula` and `DatedFormula`.
+    - `Formula` evolves over time.
+  - Remove `@dated_function`
+    - start definition goes to formula name: `formula_YYYY[_MM[_DD]]`
+    - stop is deduced from next formula start
+
+  Before:
```
fpagnoux: Nice 👏

CHANGELOG.md
fpagnoux: I don't think `AbstractVariable` was ever meant to be extended by child classes in country packages 🤔.
I don't think anyone ever did it anyway, so it's ok not to mention it.

tests/core/test_variable.py
```diff
@@ -0,0 +1,548 @@
...
+    month = '2005-01'
+    simulation = new_simulation(tax_benefit_system, month)
+    assert simulation.calculate('end_attribute__formulas__different_names', month) == 200
+
+    month = '2010-12'
+    simulation = new_simulation(tax_benefit_system, month)
+    assert simulation.calculate('end_attribute__formulas__different_names', month) == 300
+
+
+def test_clone__end_attribute__formulas__different_names():
```
fpagnoux: > I'm putting this as an optional feedback as we need to move on, but **unit** tests means than these tests should test **one behaviour**. This way, if a test does not pass, you know what is broken.
>
> This tests is basically just inspecting the whole structure of the tax benefit system. There is _no_ connection between cloning and checking that `column_by_name` is indeed storing what it is supposed to store.

fpagnoux: >In the future, please try to make your tests more focused.
>
>It's ok do add all sorts of assertions in the code when developing, if it can help you understand the complicated and lopsided structure of OpenFisca.

>But keep in mind that _every line of code_ added to `master`, including tests, will have to be maintained and understood by OpenFisca developers.

> Also, we want to change deeply the architecture of `core`. Over-testing it will have a cost : if our tests tend to check more than they should do, _every time_ we will change something (for instance renaming`IntCol`), dozens of tests that have _nothing to do_ with what we are doing will break.

openfisca_core/formulas.py
sandcha: In case it exits with None, we would have unnecessary assignments execution.

fpagnoux: If you worry about unnecessary assignments, make these constants and put them out of this function ;)

tests/core/test_variable.py
sandcha: It's was misleading of me to speak about unit tests. It started like that but they are more validation/system tests.

openfisca_core/formulas.py
```diff
@@ -774,15 +770,14 @@ def new_filled_column(
...
-    if stop_date is UnboundLocalError:
-        stop_date = None if reference_column is None else reference_column.end
-    elif stop_date is not None:
-        assert isinstance(stop_date, datetime.date)
+    if end is UnboundLocalError:
+        end = None if reference_column is None else reference_column.end
+    elif end is not None:
+        assert isinstance(end, str), 'Type error on {}. String expected. Found: {}'.format(name + '.end', type(end))
+        try:
+            get_datetime_date(end)
```
fpagnoux: @MattiSG,

We need another opinion to cut short this discussion.
We're inside the `new_filled_column` function, which builds an OpenFisca column (=variable) based on what has been coded by the legislation writer (`class rsa(Variable)...`).

The function `get_datetime_date ` parses a `'YYYY-MM-DD'` string and returns the equivalent `datetime.date`.

The suggested implementation here is to:

- First call the `get_datetime_date` function once when building the column, just to check if it returns an error. We do not save the result.
- [Then](https://github.com/openfisca/openfisca-core/pull/522/files#diff-31e80408104c8f43c6b796a369a77585R842), a little further, call it for _each_ dated formula to check that its start date is consistent with the variable end date.
- [Finally](https://github.com/openfisca/openfisca-core/pull/522/files#diff-31e80408104c8f43c6b796a369a77585R502), we also call it in the `find_function` method, whose job is to find the right dated function to use in a calculation. This method is called _each time_ the variable is calculated.

I think there is no reason to parse a string `1 + k + n` times.
(where`k` = numbers of dated formulas, `n` = numbers of times the variable is calculated). I'm therefore in favour of parsing this string once and for all when we build the variable.

@sandcha thinks otherwise, and I'll let her expose her arguments.

@MattiSG wdyt ?

sandcha: I suggest to declare an `end` attribute of `date` type to Formula class. There is no desagreement about not parsing a string multiple times.

tests/core/test_variable.py
MattiSG: I do not think it is wise to add system tests to OpenFisca in its current state, especially in a file that is already over 500 lines. Please don't rigidify stuff that nobody understands properly. It is welcome to add guarantees to new behaviour, but not very useful to check that we don't change a behaviour that might be flawed.

openfisca_core/formulas.py
fpagnoux: Ok, great 🙂 . Tell me if you need help for this implementation!

MattiSG: I've been told you reached an agreement orally, glad to have been a good plastic duck 😄

I'm obviously in favour of limiting performance impact, if there is no good reason to multiply this parsing then let's not do it — duh.

I would like to bring your attention to the fact that it is more than time to ship, though. Too much energy and time is going into something that is not a priority. If it is too hard to reach a consensus on the implementation we'll have to abandon that story and learn from it. This is becoming too much of a drain on delivering value.
I'm sure you can get this done quickly. Please don't hesitate to take shortcuts, the quality of this code is already much above average. Better done than perfect.

fpagnoux:
# Migration guide

The following regular expressions can help you adapt your OpenFisca country package to the new version of core.

## Replacing variables `stop_date` attribute by `end`

Throughout your code:

Replace

```
stop_date = date\((\d{4}), (\d{1}), (\d{1})\)
```

by 

```
end = '\1-0\2-0\3'
```

Replace

```
stop_date = date\((\d{4}), (\d{2}), (\d{1})\)
```

by 

```
end = '\1-\2-0\3'
```

Replace

```
stop_date = date\((\d{4}), (\d{1}), (\d{1})\)
```

by 

```
end = '\1-0\2-0\3'
```

Replace

```
stop_date = date\((\d{4}), (\d{1}), (\d{2})\)
```

by 

```
end = '\1-0\2-\3'
```

Replace

```
stop_date = date\((\d{4}), (\d{2}), (\d{2})\)
```

by 

```
end = '\1-\2-\3'
```

Issue #521, opened 2017-05-31
Rename “url” to “reference”


MattiSG:
The `url` attribute is mostly, and should actually be, used for legislative references. Make this usage clearer, as pioneered in the parameter API.

michelbl:
I found that `reference` is already used as an attribute of a Variable in openfisca-france, for only 3 cases:
* https://github.com/openfisca/openfisca-france/blob/master/openfisca_france/reforms/plfr2014.py#L22
* https://github.com/openfisca/openfisca-france/blob/master/openfisca_france/reforms/plfr2014.py#L39
* https://github.com/openfisca/openfisca-france/blob/master/openfisca_france/reforms/inversion_revenus.py#L51

This is not useful as the reference column is already inferred.

I suggest to remove the attribute `reference`, as used to refer to a column.

fpagnoux:
@guillett If I tell you like a variable in OpenFisca, such as `rsa`, has a `reference`, what do spontaneously think this attribute refers to ? Is it ambiguous ? 

If I tell you it refers to a legislative reference, are you surprised or does it make sense ?

MattiSG:
> I suggest to remove the attribute reference, as used to refer to a column.

Agreed. Calling the original version of a variable in a reform “reference” doesn't sound very obvious to me. And I can't find a better alternative to “legislative reference”.

MattiSG:
We have found a consensus on names as discussed in #525 and can proceed with final renames & merging. We should have discussed the naming scheme before implementation.

- `reference` for variable external documentation references.
- `baseline` for tax and benefit system references.
- `baseline_variable` for baseline variable.
- `use_baseline` for the boolean option, that was before this PR reference, triggering the load of the reference tax and benefit system.

Thanks @benjello @fpagnoux @bfabre01 & Twitter followers! 🎊

PR #525, opened 2017-06-08
Url rename

michelbl:
Connected to #521

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,15 +4,18 @@
 
 #### Breaking changes - [#525](https://github.com/openfisca/openfisca-core/pull/525)
 
-* Rename Variable attribute `url` to `reference`
+* Rename `Variable` attribute `url` to `reference`
   - This attribute is the legislative reference of a variable.
   - As previously, this attribute can be a string, or a list of strings.
-* Rename Variable attribute `reference` to `baseline_variable`
+* Rename `Variable` attribute `reference` to `baseline_variable`
   - This attibute is, for a variable defined in a reform, the baseline variable the reform variable is replacing.
 * Remove variable attribute `law_reference`
-* `TaxBenefitSystem.reference` renamed to `TaxBenefitSystem.baseline`
-* Parameter `reference` of `AbstractScenario.new_simulation()` renamed to `use_baseline`
-* Parameter `reference` of `Simulation.legislation_at()` renamed to `use_baseline`
+* Rename `TaxBenefitSystem.reference` to `TaxBenefitSystem.baseline`
+* Rename `TaxBenefitSystem.get_reference_compact_legislation` to `TaxBenefitSystem.get_baseline_compact_legislation`
+* Rename `Simulation.reference_compact_legislation_by_instant_cache` to `Simulation.baseline_compact_legislation_by_instant_cache`
+* Rename `Simulation.get_reference_compact_legislation` to `Simulation.get_baseline_compact_legislation`
+* Rename parameter `reference` of `AbstractScenario.new_simulation()` to `use_baseline`
+* Rename parameter `reference` of `Simulation.legislation_at()` to `use_baseline`
 
 ## 14.1.4 - [#539](https://github.com/openfisca/openfisca-core/pull/539)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## 15.0.0
+
+#### Breaking changes - [#525](https://github.com/openfisca/openfisca-core/pull/525)
+
+* Rename Variable attribute `url` to `reference`
+  - As previously, this attribute can be a string, or a list of strings.
+* Remove variable attribute `law_reference`
+* `TaxBenefitSystem.reference` renamed to `TaxBenefitSystem.reference_tax_benefit_system`
+* Parameter `reference` of `AbstractScenario.new_simulation()` renamed to `use_reference_tax_benefit_system`
+* Parameter `reference` of `Simulation.legislation_at()` renamed to `use_reference_tax_benefit_system`
+
 ## 14.1.4 - [#539](https://github.com/openfisca/openfisca-core/pull/539)
 
 #### New features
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## 14.0.0
+
+#### Breaking changes - [#525](https://github.com/openfisca/openfisca-core/pull/525)
+
+* Rename Variable attribute `url` to `reference`
+  - As previously, this attribute can be a string, or a list of strings.
+  - Variable attribute `law_reference` is removed.
```
fpagnoux: This should be a different item of the list, at the same level than "Rename...", and at the imperative mode.

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+  - `TaxBenefitSystem.reference` renamed to `TaxBenefitSystem.reference_tbs`.
```
fpagnoux: Not sure how I feel about that. I understand how you came to do this renaming (to be consistent about what a `reference` means), but this is not directly related to our issue.


@MattiSG thoughts ?

fpagnoux: If this change is kept: 
    - we should avoid acronyms and call it reference_tax_benefit_system if need be.
    - This is a proper breaking change, and should be a first level item

openfisca_core/taxbenefitsystems.py
```diff
@@ -134,7 +134,7 @@ def load_variable(self, variable_class, update = False):
         existing_column = self.get_column(name)
         if existing_column:
             if update:
-                attributes['reference'] = existing_column
+                attributes['reference_column'] = existing_column
```
fpagnoux: >This implementation allows the formula writer to define a `reference_column` attribute, that will be... ignored in most cases (when we are indeed replacing an existing variable).
>
>This problem existed before this PR, so not sure whether this should be treated here. I'll let you decide :)

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+  - Parameter `reference` of `AbstractScenario.new_simulation()` renamed to `use_reference_tbs`.
```
fpagnoux: The two `reference` options you're renaming to `use_reference_tbs` have always sounded weird to me: `reform.legislation_at(x, reference=True)` returns the legislation of the tax benefit system the reform was built on 😕 .
I don't know if it's relevant to rename something we may just want to depreciate one of these days 🤔 .

fpagnoux: If this change is kept: 
    - we should avoid acronyms and call it reference_tax_benefit_system if need be.
    - This is a proper breaking change, and should be a first level item

CHANGELOG.md
benjello: I would definitively prefer to use legislative_reference` instead of  `reference` and keep `reference` for the dichotomy `reference`/`reform`.

MattiSG: Calling the original version of a variable in a reform “reference” doesn't sound very obvious to me. And I can't find a better alternative to “legislative reference”. Plus, if you consider how many times the current “reference” is used vs how many times “legislative reference” should be used (once for each variable), we should definitely prioritise the most common meaning.

I'm pretty sure we can find an acceptable alternate word for the “reference” tax benefit system in the context of a reform, which would solve both the `TaxBenefitSystem.reference` and `some_variable.reference` attribute in a reform.

What about the “base” tax benefit system? The “original”?

If none of these work out (please explain why), can we call the legislative reference `source`?

benjello: My point was to keep `reference` as an attribute for `TaxBenefitSystem` since their `reference` is natural for two reasons:
 - the reference is also a TaxBenefitSystem
 - `reference/reform` is very clear dichotomy to me (`base` calls for `extension` and `original` calls for `copy`) 
   
Extension of the `reference` to the Variable as `reference_variable` or `reference` is on the same vein.

I am ok to use `source` or `legislative_reference`  instead of  `url`.

michelbl: As you wish. All these variables are now disambiguated so a renaming is straightforward.

However I cannot work on this issue after tomorrow, so if a consensus is not reached yet, someone else will have to make it and merge to master.

benjello:
Quelques résultats de la consultation (encore en cours) à l'IPP  (@SophieIPP , @bfabre01 , @abozio )
 - pour le `TaxBenefitSystem` de référence, on peut utiliser `baseline`
 - par contre `reference` pour la référence législative n'est pas forcément aussi adaptée pour un anglophone que pour un francophone.  Pas de solution géniale trouvée mais `legal_reference`, `law` ont été proposées.

MattiSG:
I really like `baseline` for the tax benefit system of a reform, I find this quite explicit!

`reference` not being obvious for a native English speaker sounds more surprising to me. Academic papers use a “references” chapter to list official references. `law` or `legal` is good for most cases, but might be a bit restrictive as I'm not 100% sure we want to have only law. Sometimes we can refer to interpretation documents too, don't we?

MattiSG:
I've tried something: https://twitter.com/openfisca/status/877806466585042944 #LazyUX 😬

benjello:
Yes different types of document describes how a tax should be computed or the value of some legislation parameters.

bfabre01:
I think "baseline" is indeed more appropriate. It may be better for english speakers. Moreover, what we call "reference" may not correspond to the actual "law" or "legal" system. For instance, when we run simulations on social benefits reform, the "reference" can be the current system with full take-up, while the take-up is partial in the facts.

benjello:
@bfabre01 : be careful `reference` that could be `law` or `legal` is a potential substitue for the `url` attribute of `Variable` not the attribute `reference`  of `TaxBenefitSystem` !

MattiSG:
Let's proceed with:

- `reference` for variable legal references.
- `baseline` for tax and benefit system references.
- `variable_baseline` for variable baseline tax and benefit system references.

capture d ecran 2017-06-22 a 12 23 27

MattiSG:
This wasn't shipped in the current sprint, so we have more time to discuss the naming scheme. Let's think in whole name sets.

benjello:
I am ok for:
- `reference` for variable legal references.
- `baseline` for tax and benefit system references.
- `baseline_variable` for baseline variable

MattiSG:
There is also `use_reference_tbs` (a boolean option, that was before this PR `reference`, triggering the load of the `reference_tbs`). I would go with `use_baseline`.

PR #526, opened 2017-06-08
Require numpy < 1.13

michelbl:
numpy 1.13 breaks openfisca

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 13.0.1 - [#526](https://github.com/openfisca/openfisca-core/pull/526)
+
+### Bug fix
+
+* Require numpy < 1.13.
+  - Openfisca is not yet compatible with the new numpy version 1.13.
+
 ## 13.0.0
 
 #### Breaking changes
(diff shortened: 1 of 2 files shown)


benjello:
@michelbl @fpagnoux : do you recall what was the problem with numpy 1.13 ?
Thanks !

michelbl:
See https://docs.scipy.org/doc/numpy-1.13.0/release.html

I think it was about `negative()` or `substract()` : 
```
negative(bool_), TypeError when negative applied to booleans.
subtract(bool_, bool_), TypeError when subtracting boolean from boolean.
```

benjello:
@michelbl : Do you have tracked down where the problems appeared in openfisca ?

michelbl:
Can't recall exactly. It may have been hidden in the infix form `boolean_vector_a - boolean_vector_b`. Why are you interested in that issue ?

benjello:
Thanks @michelbl : I am interested in keep openfisca code in line with latest version of main packages because people usually install openfisca on the top of distribution like anaconda or debian that usually provide latest version of those packages. And I want to avoid the pain of dealing with this kind of dependencies. In that line we should think on moving to python 3 too.

Issue #464, opened 2017-03-05
Do not hide import errors in `openfisca-run-test`


benjello:
When I try to run `openfisca-run-test` using a country-package I am working on I cannot sort out the ImportError origin between an install/packaging problem from a coding error specific to the country-package.
The [following lines](https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/scripts/__init__.py#L30-L33):
````python
    try:
        country_package = importlib.import_module(country_package_name)
    except ImportError:
        handle_error(u'Could not import module `{}`. Make sure it is installed in your environment.'.format(country_package_name))
```` 
catches all the ImportError without showing me the actual error.

@fpagnoux @cbenz what about printing the catched error ?

**Comment by MattiSG** [comment]
Expected result:

```
{{ stack trace }}

Could not import module {}.
Are you sure it is installed in your environment? If so, look at the stack trace above to determine the origin of this error.
See more at <https://github.com/openfisca/country-template#installing>.
```

PR #527, opened 2017-06-14
Do not hide error at module import failure

sandcha:
Connected to #464 

#### New features

- Add stack trace when a module import fails at `openfisca-run-test`
   - Improve error message whether import is called with `-c` option or not

### DIFF
--- a/openfisca_core/scripts/__init__.py
+++ b/openfisca_core/scripts/__init__.py
@@ -5,6 +5,7 @@ import importlib
 import logging
 import pkgutil
 import sys
+from os import linesep
 
 log = logging.getLogger(__name__)
 logging.basicConfig(format='%(levelname)s: %(message)s')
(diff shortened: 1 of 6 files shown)


openfisca_core/scripts/__init__.py
```diff
@@ -28,7 +29,10 @@ def build_tax_benefit_sytem(country_package_name, extensions, reforms):
     try:
         country_package = importlib.import_module(country_package_name)
     except ImportError:
-        handle_error(u'Could not import module `{}`. Make sure it is installed in your environment.'.format(country_package_name))
+        handle_error(u'{}\nCould not import module `{}`.\n'.format(traceback.format_exc(), country_package_name)
```
fpagnoux: `\n` would not work on Windows.
To make your code cross-platfrom, you can use [`os.linesep`](https://docs.python.org/3/library/os.html#os.linesep) instead.

> You may want to use the `join` string method:
>```py
>message = os.linesep.join([
>    "First line",
>    "Second line"
>    ])
>print(message)
>>>> First line
>>>> Second line
>```

openfisca_core/scripts/__init__.py
```diff
@@ -51,7 +55,11 @@ def detect_country_package():
     for module_description in pkgutil.iter_modules():
         module_name = module_description[1]
         if 'openfisca' in module_name.lower():
-            module = importlib.import_module(module_name)
+            try:
+                module = importlib.import_module(module_name)
+            except ImportError:
+                handle_error(u'{}\nCould not import module `{}`.\n'.format(traceback.format_exc(), module_name)
```
fpagnoux: Same remark.

openfisca_core/scripts/__init__.py
sandcha: Thanks @fpagnoux. Done.

fpagnoux:
Also: we have the same issue [here](https://github.com/openfisca/openfisca-core/blob/master/openfisca_web_api_preview/loader/tax_benefit_system.py#L9). I think it would be relevant fix it in this PR 🙂.

sandcha:
@fpagnoux Done and updated for all ImportError messages in Core.

Linked to https://github.com/openfisca/openfisca-extension-template/pull/3
As I slightly updated the extension template readme in order to add the same reference for country and extension modules import errors.

PR #529, opened 2017-06-22
Welcome users to the API

Anna-Livia:
#### New features
This PR adds 2 routes to the OpenFisca Web API : 
- '/' --> now has a welcome message 
The root of the API used to return a 404, misleading users intu thinking there was an issue.
The welcome message clears up the situation.

-'/calculate' --> now has a welcome message (for GET method)
The HTTP request modules check if a route exists through GET before sending a POST request.
By adding a welcome message, we avoid a 404 being return on the GET check.


MattiSG:
I don't think that these messages are useful as such. “Welcome” is nice but useless, and can actually be misleading for autodiscovery since `200` status codes are sent. Additionally, if you want to open these endpoints, then you have to document them too.

Who are your users misled due to a `404` on `/`? What were they expecting?

If we go this way, I'm strongly in favour of using [HATEOAS](http://putaindecode.io/fr/articles/api/hateoas/).

fpagnoux:
>Who are your users misled due to a 404 on /? What were they expecting?
@Anna-Livia  thought the API didn't work because the landing page was an error message

>If we go this way, I'm strongly in favour of using HATEOAS.

👍. So it's probably not worth investing in a new welcome page, if it's not HATEOAS

MattiSG:
> thought the API didn't work because the landing page was an error message

I understand that, yet a `404` on an undocumented endpoint (`/` is one if we think in terms of API) is standard procedure… The API “not working” would have been the server giving no answer or serving a `500`. I see the point of having a generic HATEOAS help message served on all `404`, but not of having `/` being a “test” endpoint.

Closing this PR. Users, feel free to comment and ask for reopen if you felt this need! 🙂

PR #533, opened 2017-06-22
Fix simulation.clone and entity.clone

fpagnoux:
#### Bug fix

- Fix `simulation.clone` and `entity.clone` methods

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 14.1.1 - [#533](https://github.com/openfisca/openfisca-core/pull/533)
+
+#### Bug fix
+
+- Fix `simulation.clone` and `entity.clone` methods.
+  - A bug was introduced in `14.1.0`
+
 ## 14.1.0 - [#528](https://github.com/openfisca/openfisca-core/pull/528)
 
 #### New features
--- a/openfisca_core/holders.py
+++ b/openfisca_core/holders.py
@@ -100,7 +100,7 @@ class Holder(object):
     def calculate_output(self, period):
         return self.formula.calculate_output(period)
 
-    def clone(self):
+    def clone(self, entity):
         """Copy the holder just enough to be able to run a new simulation without modifying the original simulation."""
         new = empty_clone(self)
         new_dict = new.__dict__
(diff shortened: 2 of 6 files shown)


fpagnoux:
Fixes france's build issue : https://github.com/openfisca/openfisca-france/pull/786

openfisca_core/holders.py
```diff
@@ -100,7 +100,7 @@ def calculate(self, period, **parameters):
     def calculate_output(self, period):
         return self.formula.calculate_output(period)
 
-    def clone(self):
+    def clone(self, new_entity):
```
sandcha: Rename `new_entity` to `entity` ? `new_entity` name assumes something out of the bounds of this method.

openfisca_core/simulations.py
```diff
@@ -132,12 +132,13 @@ def clone(self, debug = False, debug_all = False, trace = False):
...
                 new_dict[key] = value
 
-        new.persons = self.persons.clone()
+        new.persons = self.persons.clone(new)
         setattr(new, new.persons.key, new.persons)
         new.entities = {new.persons.key: new.persons}
 
         for entity_class in self.tax_benefit_system.group_entities:
-            entity = self.entities[entity_class.key].clone()
+            entity = self.entities[entity_class.key].clone(new)
```
fpagnoux: No, it's not a class, but an instance ;)

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 14.1.1 - [#533](https://github.com/openfisca/openfisca-core/pull/533)
+
+#### Bug fix
+
+- Fix `simulation.clone` and `entity.clone` methods
```
fpagnoux: 👍

tests/core/test_simulations.py
```diff
@@ -25,3 +26,26 @@ def test_calculate__holder_attribute_content():
...
+                "bill": {"salary": {"2017-01": 3000}},
+                },
+            })
+
+    simulation_clone = simulation.clone()
+    assert simulation != simulation_clone
+
+    for entity_id, entity in simulation.entities.iteritems():
+        assert entity != simulation_clone.entities[entity_id]
+
```
fpagnoux: 👌

openfisca_core/holders.py
fpagnoux: 👌

openfisca_core/simulations.py
```diff
@@ -132,12 +132,13 @@ def clone(self, debug = False, debug_all = False, trace = False):
             if key not in ('debug', 'debug_all', 'trace', 'entities'):
```
fpagnoux: Not sure, but I think we don't want the `debug`, `debug_all` and `trace` attributes of the original simulation to be copied to the new one. Instead, we only take their value from the parameters of the `clone` method.

fpagnoux: >What about the traceback ? Add it to the keys to ignore ?

It was not ignored before this PR, nor before `14.1.0`. As we don't fully understand the subtlety of this code, I'd prefer to try adding things without exactly knowing why.

fpagnoux: > Add `persons` ?

We redefine `new.person` just further, so it's useless. I remove `entities` as well, for the same reason.

PR #538, opened 2017-06-28
Make missing_value base function compatible with v4 syntax and extra params

fpagnoux:
This `base_function` was hidden in `openfisca_core/formulas.py` instead of `openfisca_core/base_functions.py`.

Therefore several changes applied to base functions were not applied to it. This PR fixes it, and make this base function work [like the others](https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/base_functions.py#L11)

This bug was without consequences, as this base function is rarely used. We however need to fix that to keep migrating france code towards the `v4` syntax.

### DIFF
--- a/openfisca_core/formulas.py
+++ b/openfisca_core/formulas.py
@@ -610,10 +610,10 @@ def calculate_output_divide(formula, period):
     return formula.holder.compute_divide(period).array
 
 
-def missing_value(formula, simulation, period):
+def missing_value(formula, simulation, period, *extra_params):
     function = formula.find_function(period)
     if function is not None:
-        return function(simulation, period)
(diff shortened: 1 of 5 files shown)


openfisca_core/formulas.py
```diff
@@ -610,10 +610,10 @@ def calculate_output_divide(formula, period):
     return formula.holder.compute_divide(period).array
 
 
-def missing_value(formula, simulation, period):
+def missing_value(formula, simulation, period, *extra_params):
```
sandcha: Move it to `base_functions` file ?

fpagnoux:
@sandcha have your remarks been treated 🙂  ?

Issue #532, opened 2017-06-22
Document new simulation constructor


MattiSG:
The new version of the `/calculate` endpoint of the web API has implied a new constructor for simulations in the Python API.

This constructor is currently only visible in the changelog. It should be documented ~~and the previous one deprecated.~~

fpagnoux:
As discussed on slack, it seems early to deprecate the current way of building a simulation in favor of the new one, as we may need to iterate on the latter.

However, it should be clear in the code that the new constructor aims at replacing the former one, so that no developper wonder why there seems to be two ways of doing the same thing.

PR #539, opened 2017-06-28
Improve preview  api

fpagnoux:
Connected to openfisca/demonstrator#3
Connected to #532

A few problems emerged when using the new route `/calculate` with `openfisca-france`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
   - Force all persons to be allocated to all entities
   - Improve error messages
   - Detect unexpected entity errors first
+- Document new simulation and entities constructors.
 
 ## 14.1.3 - [#541](https://github.com/openfisca/openfisca-core/pull/541)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## 14.1.4 - [#539](https://github.com/openfisca/openfisca-core/pull/539)
+
+#### New features
+
+- In the preview API `/calculate` route
+  - Handle roles with no plural
+  - Force all persons to be allocated to all entities
+  - Improve error messages
+  - Detect unexpected entity errors first
+
 ## 14.1.3 - [#541](https://github.com/openfisca/openfisca-core/pull/541)
 
 #### Minor Change
--- a/openfisca_core/entities.py
+++ b/openfisca_core/entities.py
@@ -290,6 +290,9 @@ class GroupEntity(Entity):
             self.members_role[person_index] = self.flattened_roles[0]
             self.members_legacy_role[person_index] = 0
 
+        #  Deprecated attribute used by deprecated projection opertors, such as sum_by_entity
+        self.roles_count = self.members_legacy_role.max() + 1
+
     def init_members(self, roles_json, entity_id):
         for role_id, role_definition in roles_json.iteritems():
             check_type(role_definition, list, [self.plural, entity_id, role_id])
@@ -313,9 +316,6 @@ class GroupEntity(Entity):
             self.members_entity_id[person_index] = entity_index
             self.members_role[person_index] = person_role
             self.members_legacy_role[person_index] = person_legacy_role
-
-        #  Deprecated attribute used by deprecated projection opertors, such as sum_by_entity
-        self.roles_count = self.members_legacy_role.max() + 1
 
     @property
     def members_role(self):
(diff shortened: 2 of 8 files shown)


fpagnoux:
Connected to openfisca/demonstrator#3

openfisca_core/entities.py
```diff
@@ -281,15 +285,15 @@ def init_from_json(self, entities_json):
 
         Entity.init_from_json(self, entities_json)
```
sandcha: Choose between attributes like `entities_json` and calling the simulation object ?  Simulation is called before and it knows the json entities, doesn't it ?

openfisca_core/simulations.py
```diff
@@ -65,13 +72,19 @@ def __init__(
     def instantiate_entities(self, simulation_json):
         if simulation_json:
             check_type(simulation_json, dict, ['error'])
-            copied_simulation_json = simulation_json.copy()  # Avoid mutating the input
-
-            persons_json = copied_simulation_json.pop(self.tax_benefit_system.person_entity.plural, None)
+            allowed_entities = set(entity_class.plural for entity_class in self.tax_benefit_system.entities)
+            unexpected_entities = [entity for entity in simulation_json if entity not in allowed_entities]
```
sandcha: 👍

tests/core/test_entities.py
```diff
@@ -184,52 +184,6 @@ def test_set_input_with_constructor():
     assert_near(person('salary', "2017-10"), [2000, 3000, 1600, 0, 0])
 
 
-def test_person_without_household():
```
sandcha: Keep the tests subjects and check for error type and unknown entity ?

tests/web_api/test_calculate.py
```diff
@@ -47,8 +47,7 @@ def test_responses():
         ('{"persons": {"bob": {"salary": {"invalid period": 2000 }}}}', BAD_REQUEST, 'persons/bob/salary/invalid period', 'Invalid period',),
         ('{"persons": {"bob": {"salary": {"invalid period": null }}}}', BAD_REQUEST, 'persons/bob/salary/invalid period', 'Invalid period',),
         ('{"persons": {"bob": {"basic_income": {"2017": 2000 }}}}', BAD_REQUEST, 'persons/bob/basic_income/2017', 'basic_income is only defined for months',),
-        ('{"persons": {"bob": {}}, "households": null}', OK, None, None,),
-        ('{"persons": {"bob": {"birth": {"ETERNITY": "1980-01-01"} }}}', OK, None, None,),
+        ('{"persons": {"bob": {"birth": {"ETERNITY": "1980-01-01"} }}}', BAD_REQUEST, 'households', 'is not a member of any household',),
```
sandcha: Add test on a role with no plural ?

fpagnoux: There is no such a role in the country template (and adding it would be making the template unnecessarily complex).

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## 14.1.3 - [#539](https://github.com/openfisca/openfisca-core/pull/539)
```
fpagnoux: Thanks, I'll keep that in mind, but that is usually done while rebasing.

tests/core/test_entities.py
fpagnoux: We already test the content of the error message [here](https://github.com/openfisca/openfisca-core/blob/7414070d634bef4040ec6e4ed246695f77bbfc16/tests/web_api/test_calculate.py#L50). What exactly would you like to test 🤔  ?

openfisca_core/entities.py
fpagnoux: Sorry, but I'm not sure I understand your suggestion. Could you please rephrase it ?

fpagnoux: At this stage, the simulation is not yet fully initialised. We are in the process of building a "group" entity, and this is part of building the simulation.

We can (and need to) call `self.simulation.persons` because by construction, the "persons" entity is [always the first entity to be built](https://github.com/openfisca/openfisca-core/blob/7414070d634bef4040ec6e4ed246695f77bbfc16/openfisca_core/simulations.py#L88). We actually need to know data about the person entity to be able to build the group entities.

tests/core/test_entities.py
sandcha: Sorry ; nothing. It was about unknown entity but I see now that it was tested earlier.

PR #541, opened 2017-07-03
Rewrite /calculate example for france

fpagnoux:
Depends on #539
Connected to openfisca/legislation-explorer#100

### DIFF
--- a/openfisca_web_api_preview/openAPI.yml
+++ b/openfisca_web_api_preview/openAPI.yml
@@ -55,7 +55,7 @@ paths:
           schema:
             $ref: '#/definitions/SituationOutput'
         404:
-          description: "A variable mentioned in the input siutation does not exist in the loaded tax and benefit system. Details are sent back in the response body"
+          description: "A variable mentioned in the input situation does not exist in the loaded tax and benefit system. Details are sent back in the response body"
           headers:
             Country-Package:
               description: "The name of the country package currently loaded in this API server"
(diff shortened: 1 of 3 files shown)


openfisca_web_api_preview/openAPI.yml
```diff
@@ -306,78 +306,58 @@ definitions:
...
-          housing_tax:
-              "2017": null
-          accomodation_size:
-              "2017-01": 300
-              "2017-02": 300
-              "2017-03": 300
-              "2017-04": 300
-              "2017-05": 300
+          enfants: [Janet]
+      foyer_fiscaux:
```
Anna-Livia: foyers_fiscaux

openfisca_web_api_preview/openAPI.yml
```diff
@@ -306,78 +306,58 @@ definitions:
...
+          salaire_de_base:
+            "2017": 30000
+        Janet: {}
+      menages:
+        menage_1:
+          personne_de_reference: [Bill]
+          conjoint: [Bob]
+          enfants: [Janet]
+          revenu_disponible:
+            "2017": 44629
```
Anna-Livia: For calculation results, here is what the API returns
 "impots_directs": {
        "2017": -3733.7275390625
      },
      "revenu_disponible": {
        "2017": 36585.8984375
-->I think it could be interesting to give example results as Floats so that it is consistant with what the API returns.

openfisca_web_api_preview/openAPI.yml
fpagnoux: 👍

openfisca_web_api_preview/openAPI.yml
fpagnoux: 👌

Issue #500, opened 2017-05-04
Deprecate parsers


MattiSG:
- [x] Remove parsers optional dependency.
- [x] Remove any documentation on the topic.
- [x] Document that the parsers are deprecated and explain why (in the README).
- [x] Update and close obsolete issues in `openfisca-parsers`.

cbenz:
The script [`show_merged_parameters.py`](https://github.com/openfisca/openfisca-france/blob/baremes-ipp/openfisca_france/scripts/parameters/baremes_ipp/show_merged_parameters.py) uses the parsers.

In a discussion with @fpagnoux we proposed to downgrade the feature of the script which relies on the parsers, if it's not a core feature.

PR #545, opened 2017-07-10
Deprecate parsers

fpagnoux:
Connected to #500 

#### Breaking changes

* Deprecate `parsers`
  - They can no longer be installed through `pip install openfisca_core[parsers]`
  - They can still install them directly with `pip install openfisca_parsers`, but they won't be maintained.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+# 16.0.0
+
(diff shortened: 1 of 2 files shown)


fpagnoux:
No reviews the night before the demo. As this is non-critical and low-risk, I'm rebasing and merging.

PR #546, opened 2017-07-12
Piwik tracking

sandcha:
Connected to [openfisca-web-api/#138](https://github.com/openfisca/openfisca-web-api/issues/138)

#### New features

- Introduce `openfisca_tracker` module
  - Allows for piwik tracking of preview web api endpoints calls


tests/web_api/test_tracking.py
```diff
@@ -0,0 +1,95 @@
+# -*- coding: utf-8 -*-
```
fpagnoux: Unless this test _actually_ test that the running time is lower than a defined upper bound, this should not be executed each time we run the test.

tests/web_api/__init__.py
```diff
@@ -3,5 +3,14 @@
 from openfisca_web_api_preview.app import create_app
 
 TEST_COUNTRY_PACKAGE_NAME = 'openfisca_country_template'
+TRACKER_URL = 'https://openfisca.innocraft.cloud/piwik.php'
+TRACKER_IDSITE = 1
+
+ACTIVE_TRACKING = True
```
sandcha: To allow tests with and without tracking.

tests/web_api/__init__.py
```diff
@@ -3,5 +3,14 @@
...
+
 distribution = pkg_resources.get_distribution(TEST_COUNTRY_PACKAGE_NAME)
-subject = create_app(TEST_COUNTRY_PACKAGE_NAME).test_client()
+
+if ACTIVE_TRACKING:
+    subject = create_app(TEST_COUNTRY_PACKAGE_NAME, TRACKER_URL, TRACKER_IDSITE).test_client()
```
sandcha: Ok for removing it. It's a way to test the whole chain without writing requests. But, it's only interesting when we use a Piwik instance dedicated to tests.

tests/web_api/test_tracking.py
sandcha: Renamed to `measure_tracking.py` to avoid automatic execution with other tests.
I don't know what the upper bound should be and if it should be different on circle. Any suggestions?

README.md
```diff
@@ -35,3 +35,29 @@ COUNTRY_PACKAGE=openfisca_country_template gunicorn "openfisca_web_api_preview.a
...
+The tracker is not installed by default. To install it, run:
+
+```sh
+pip install openfisca_core[tracker]  # Or `pip install --editable ".[tracker]"` for an editable installation
+```
+
+
+#### Tracker configuration
+
+The tracker is activated when these two environment variables are set:
```
fpagnoux: > I wonder if it would not be simpler to just combine these two environment variable into one, for instance `TRACKER=https://stats.data.gouv.fr/piwik.php?idsite=2` 🤔

fpagnoux: Doesn't seem so obvious with bull tracking.

PR #547, opened 2017-07-26
Check version number change before deployment

### LINKED ISSUE #513: Prevent false negatives in Circle tests on master branch
If we rebuild a Circle CI master build, it [turns red](https://circleci.com/gh/openfisca/openfisca-core/90).

This is due to the fact that we try to tag the commit with a label that already exists.

~~This is due to the fact that we wrongfully check that the version has been bumped~~.

~~Suggested implementation: don't run the script `check-version-bump.sh` on branch master.~~

Anna-Livia:
Connected to #513 

#### Technical changes
When there is a new commit on the master branch, if the version number has not been changed, the deployment will not be triggered.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -17,10 +17,4 @@ deployment:
   master:
     branch: master
     commands:
-      if ! git rev-parse `python setup.py --version` ; then
-        - git tag `python setup.py --version`
-        - git push --tags
-        - python setup.py bdist_wheel
-        - twine upload dist/* --username openfisca-bot --password $PYPI_PASSWORD
-        - ssh deploy-new-api@api-test.openfisca.fr
-      fi
+      - ./deploy-if-version-bump.sh
--- a/circle.yml
+++ b/circle.yml
@@ -17,8 +17,10 @@ deployment:
   master:
     branch: master
     commands:
-      - git tag `python setup.py --version`
-      - git push --tags
-      - python setup.py bdist_wheel
-      - twine upload dist/* --username openfisca-bot --password $PYPI_PASSWORD
-      - ssh deploy-new-api@api-test.openfisca.fr
+      if ! git rev-parse `python setup.py --version` ; then
+        - git tag `python setup.py --version`
+        - git push --tags
+        - python setup.py bdist_wheel
+        - twine upload dist/* --username openfisca-bot --password $PYPI_PASSWORD
+        - ssh deploy-new-api@api-test.openfisca.fr
+      fi
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,9 @@
 machine:
   python:
-    version: 2.7.12
+    version: 2.7.12  # this specific version is preinstalled on the CI servers and allows faster builds
+  environment:
+    PYPI_USERNAME: openfisca-bot
+    # PYPI_PASSWORD: this value is set in CircleCI's web interface; do not set it here, it is a secret!
 
 dependencies:
   override:
--- a/deploy-if-version-bump.sh
+++ b/deploy-if-version-bump.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+set -e
+if ! git rev-parse `python setup.py --version` 2>/dev/null ; then
+    git tag `python setup.py --version`
+    git push --tags
+    python setup.py bdist_wheel
+    twine upload dist/* --username openfisca-bot --password fakepassword
+    ssh deploy-new-api@api-test.openfisca.fr
+else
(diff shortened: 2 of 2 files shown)


circle.yml
```diff
@@ -17,8 +17,4 @@ deployment:
   master:
     branch: master
     commands:
-      - git tag `python setup.py --version`
-      - git push --tags
-      - python setup.py bdist_wheel
-      - twine upload dist/* --username openfisca-bot --password $PYPI_PASSWORD
-      - ssh deploy-new-api@api-test.openfisca.fr
+      - sh ./deploy-if-version-bump.sh
```
fpagnoux: It's more usual to make the script executable with `chmod +x deploy-if-version-bump.sh`, and then just write:
```
- ./deploy-if-version-bump.sh
```

(See what's done [line 15](https://github.com/openfisca/openfisca-core/pull/547/files#diff-29944324a3cbf9f4bd0162dfe3975d88R15))

deploy-if-version-bump.sh
```diff
@@ -0,0 +1,13 @@
+# deploy-if-version-bump.sh
```
fpagnoux: Is this comment necessary ?

deploy-if-version-bump.sh
```diff
@@ -0,0 +1,13 @@
...
+#!/bin/sh
```
fpagnoux: This line states than your script should be using with the machine `sh`. This `sh` can be `zsh` on your machine, `bash` quite often, or some older and more basic shells in certain machines.

If your script uses `bash` features, and you declare `#!/bin/sh` here, it may not work on the latter machines.

To avoid this issue, it's more common to ask the machine to use `bash`. To do so, just replace this line by `#! /usr/bin/env bash`.

deploy-if-version-bump.sh
```diff
@@ -0,0 +1,13 @@
...
+
+set -e
+if ! git rev-parse `python setup.py --version` 2>/dev/null ; then
+    git tag `python setup.py --version`
+    git push --tags
+    python setup.py bdist_wheel
+    twine upload dist/* --username openfisca-bot --password fakepassword
```
fpagnoux: Careful: don't forget to switch it back to the real pasword !

deploy-if-version-bump.sh
fpagnoux: @MattiSG tell me if I'm saying nonsense, I'm not a shell expert ;).

fpagnoux:
Also: this doesnt fixes #513 ! The same changes should be applied to `openfisca-france` and `openfisca-country-template`

circle.yml
MattiSG: Slightly less secure too, but more usual indeed :wink:

deploy-if-version-bump.sh
MattiSG: If you do rely on `bash`-specific features, indeed then specify `bash`. However, for portability, it is better to stay with `sh`, because `sh` is a standard more than an implementation. Specifying something else means indeed you opt-in to more specific features.

deploy-if-version-bump.sh
Anna-Livia: 👍

deploy-if-version-bump.sh
Anna-Livia: 👍

circle.yml
Anna-Livia: 👍

deploy-if-version-bump.sh
Anna-Livia: So we keep #!/bin/sh ?

deploy-if-version-bump.sh
```diff
@@ -0,0 +1,12 @@
...
+
+set -e
+if ! git rev-parse `python setup.py --version` 2>/dev/null ; then
+    git tag `python setup.py --version`
+    git push --tags  # update the repository version
+    python setup.py bdist_wheel  # build this package in the dist directory
+    twine upload dist/* --username $PYPI_USERNAME --password $PYPI_PASSWORD  # publish
+    ssh deploy-new-api@api-test.openfisca.fr
+else
+    echo "No deployment - Existing release with `python setup.py --version` version number"
```
fpagnoux: > As an OpenFisca developper, it is not clear to me how `Existing release with 16.1.0 version number` is related to `No deployment`. The fact that a version already exists is the _technical_ criterion, not the _functional_ reason.
> What about something like : `No deployment - Only non-functional elements were modified in this change` ?

PR #550, opened 2017-08-02
Add Enum possible values to the web API

fpagnoux:
Connected to openfisca/legislation-explorer#95

Depends on https://github.com/openfisca/country-template/pull/16

In the preview web API, for variables of type `Enum`:

* Accept and recommand to use strings as simulation inputs, instead of the enum indices.
  - For instance, `{"housing_occupancy_status": {"2017-01": "Tenant"}}` is now accepted and prefered to `{"housing_occupancy_status": {"2017-01": 0}}`).
  - Using the enum indices as inputs is _still accepted_ for backward compatibility, but _should not_ be encouraged.
* Return strings instead of enum indices.
  - For instance, is `housing_occupancy_status` is calculated for `2017-01`, `{"housing_occupancy_status": {"2017-01": "Tenant"}}` is now returned, instead of `{"housing_occupancy_status": {"2017-01": 0}}`.
* In the `/variable/<variable_name>` route, document possible values. 
* In the Open API specification, document possible values following JSON schema.

### DIFF
--- a/openfisca_core/entities.py
+++ b/openfisca_core/entities.py
@@ -13,6 +13,7 @@ from simulations import check_type, SituationParsingError
 from holders import Holder, PeriodMismatchError
 from periods import compare_period_size, period as make_period
 from taxbenefitsystems import VariableNotFound
+from columns import EnumCol
 
 
 class Entity(object):
@@ -79,6 +80,8 @@ class Entity(object):
                     array = holder.buffer.get(period)
                     if array is None:
                         array = holder.default_array()
+                    if isinstance(holder.column, EnumCol) and isinstance(value, basestring):
+                        value = holder.column.enum[value]
 
                     try:
                         array[entity_index] = value
--- a/openfisca_core/entities.py
+++ b/openfisca_core/entities.py
@@ -59,34 +59,42 @@ class Entity(object):
     def init_variable_values(self, entity_object, entity_id):
         entity_index = self.ids.index(entity_id)
         for variable_name, variable_values in entity_object.iteritems():
+            path_in_json = [self.plural, entity_id, variable_name]
             try:
                 self.check_variable_defined_for_entity(variable_name)
             except ValueError as e:  # The variable is defined for another entity
-                raise SituationParsingError([self.plural, entity_id, variable_name], e.message)
+                raise SituationParsingError(path_in_json, e.message)
             except VariableNotFound as e:  # The variable doesn't exist
-                raise SituationParsingError([self.plural, entity_id, variable_name], e.message, code = 404)
+                raise SituationParsingError(path_in_json, e.message, code = 404)
 
             if not isinstance(variable_values, dict):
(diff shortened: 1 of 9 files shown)


fpagnoux:
### Example

```JSON
http://localhost:5000/variable/statut_marital
```

openfisca_core/enumerations.py
```diff
@@ -6,6 +6,7 @@ def __init__(self, varlist, start = 0):
         self._vars = {}
         self._nums = {}
         self._count = 0
+        self.list = varlist
```
Anna-Livia: If the user gives a value that is not in the list, would it be possible to catch it like this :

```py
class Enum(object):
    def __init__(self, varlist, start = 0):
        self._vars = {}
        self._nums = {}
        self._count = 0
        self.list = varlist
        for var in varlist:
            self._vars.update({self._count + start: var})
            self._nums.update({var: self._count + start})
            self._count += 1

    def __getitem__(self, var):
        try:
            return self._nums[var]
        except KeyError:
            valid_keys = ', '.join(self.list)
            log.error("' {} ' is not a valid entry for this variable. The available values are: {}".format(var, valid_keys))
            raise
```

openfisca_web_api_preview/loader/variables.py
```diff
@@ -60,8 +70,8 @@ def build_variable(variable, country_package_metadata):
     result = {
         'id': variable.name,
         'description': variable.label,
-        'valueType': variable.__class__.__name__.replace('Col', ''),
-        'defaultValue': format_value(variable.default),
+        'valueType': get_variable_type(variable),
```
Anna-Livia: 💯

openfisca_core/enumerations.py
fpagnoux: Good catch 💯 !

Indeed, if users send an invalid value, we must return a `400` response (and not a `500`), and give them a clear answer so that they can fix their request.
As it's a client mistake, and not an internal server error, we don't need to log, but we need to catch the error.

fpagnoux:
I fixed the error when a user sends an invalid value for an enum:

```JSON
{
  "households": {
    "_": {
      "housing_occupancy_status": {
        "2017-01": " 'Toto' is not a valid value for 'housing_occupancy_status'. Possible values are ['Tenant', 'Owner', 'Free logder', 'Homeless']."
      }
    }
  }
}
```

fpagnoux:
@Anna-Livia je te laisse reviewer le dernier changement 🙂

openfisca_core/entities.py
```diff
@@ -58,34 +58,42 @@ def init_from_json(self, entities_json):
     def init_variable_values(self, entity_object, entity_id):
         entity_index = self.ids.index(entity_id)
         for variable_name, variable_values in entity_object.iteritems():
+            path_in_json = [self.plural, entity_id, variable_name]
```
Anna-Livia: 💯

PR #551, opened 2017-08-02
Usa valid JSON types in spec

fpagnoux:
Return valid JSON types in the Open API specification

`Boolean`, `Int`, `Float`, `Date` are not valid types 😁 .

Note: this is not in the sprint, so no emergency.


fpagnoux:
There is more wrong this types: for instance the API returns exotic types such as `FixedStr`. Closing the PR for now, will re-open later with a more complete fix.

PR #553, opened 2017-08-04
Make deploy script executable

fpagnoux:
Deployment is broken since #547, as script is not executable (`chmod +x deploy-if-version-bump.sh`).


fpagnoux:
I checked `france` and `country_template`, this is only an issue for `core`.

PR #563, opened 2017-08-23
Check type reference

### LINKED ISSUE #556: Fix API references serialisation
Some references seem to be too escaped in the variables API, and to output the `u'` Python Unicode marker.

[Example](https://api-test.openfisca.fr/variable/apl) and rendering:

capture d ecran 2017-08-17 a 12 45 05

Anna-Livia:
Fixes #556 

#### Technical changes

- Allows for user to put a coma at the end of a variable reference, without consequence to the API output.
- All variable references are transformed into lists of strings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+### 17.0.1
+
+#### Minor Change
+
+- Handle the case where the user accidentally puts a coma at the end of a variable reference so it has no consequence on the API output.
+- All variable references are transformed into lists of strings.
+
 # 17.0.0 - [#552](https://github.com/openfisca/openfisca-core/pull/552)
 
 #### Breaking changes
@@ -192,7 +199,6 @@
 * In the API preview, update the internal transformation of the parameters.
 
 * In the directory `script`, add a subdirectory `migrations`.
-
 
 ## 16.3.0
 
--- a/openfisca_core/variables.py
+++ b/openfisca_core/variables.py
@@ -58,7 +58,7 @@ class Variable(object):
 
         reference = self.attributes.pop('reference', None)
         if reference:
-            if isinstance(reference, unicode) or isinstance(reference, str):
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+### 16.3.1
+
+#### Minor Change
+
+- Allows for user to put a coma at the end of a variable reference, without consequence to the API output.
```
fpagnoux: Formulated like this, this sounds like a feature, while it's just robustness to user mistakes. What about something like:

> Handle the case where the user accidentally puts a coma at the end of a variable reference...

openfisca_core/variables.py
```diff
@@ -56,6 +56,17 @@ def to_column(self, tax_benefit_system):
             except ValueError:
                 raise ValueError(u"Incorrect 'end' attribute format in '{}'. 'YYYY-MM-DD' expected where YYYY, MM and DD are year, month and day. Found: {}".format(self.name, end).encode('utf-8'))
 
+        reference = self.attributes.pop('reference', None)
+        if reference:
+            if isinstance(reference, unicode) or isinstance(reference, str):
```
fpagnoux: `isinstance(reference, basestring)` is equivalent to `isinstance(reference, unicode) or isinstance(reference, str)`

openfisca_core/variables.py
```diff
@@ -56,6 +56,17 @@ def to_column(self, tax_benefit_system):
...
+                reference = [reference]
+            elif isinstance(reference, list):
+                pass
+            elif isinstance(reference, tuple):
+                reference = list(reference)
+            else:
+                raise TypeError('The reference of the variable {} is a {} instead of a String or a List of Strings.'.format(self.name, type(reference)))
```
fpagnoux: We are not checking that our reference is a list of string.

If I write `reference = Household,` for instance, there will be no error raised.

Issue #567, opened 2017-08-31
Keep references when migrating XML params to YAML


fpagnoux:
The migration script currently does not keep the parameter references

See openfisca/openfisca-france#810

PR #569, opened 2017-09-05
Keep references when migrating XML params to YAML

michelbl:
Connected to #567.

### DIFF
--- a/openfisca_core/scripts/migrations/v16_2_to_v17/xml_to_yaml.py
+++ b/openfisca_core/scripts/migrations/v16_2_to_v17/xml_to_yaml.py
@@ -80,7 +80,13 @@ def transform_values_history(children, value_format):
         elif child.tag == 'END':
             values[date] = {'value': None}
         elif child.tag == 'PLACEHOLDER':
-            values[date] = "expected"
+            values[date] = 'expected'
+
+        if 'reference' in child.attrib:
+            if values[date] == 'expected':
+                values[date] = {'expected': None, 'reference': child.attrib['reference']}
+            else:
+                values[date]['reference'] = child.attrib['reference']
 
     return values
 
@@ -260,7 +266,7 @@ def write_yaml(node, path):
         raise ValueError('Unknown type {}'.format(node_type))
 
 
-def write_legislation(legislation_xml_info_list, path):
+def write_parameters(legislation_xml_info_list, path):
     params_tree = load_legislation(legislation_xml_info_list)
 
     write_yaml(params_tree, path)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+### 17.1.2 [#569](https://github.com/openfisca/openfisca-core/pull/569)
+
+- Fix migration script `xml_to_yaml.py`
+  - The attributes `reference` of nodes `END`, `VALUE`, `PLACEHOLDER`, `BAREME`, `TRANCHE`, `ASSIETTE`, `TAUX`, `MONTANT`, `SEUIL` were not parsed. Now they are parsed to YAML.
+
 ### 17.1.1
 
 - Reference the new address of the documentation
--- a/openfisca_core/scripts/migrations/v16_2_to_v17/legislation.xsd
+++ b/openfisca_core/scripts/migrations/v16_2_to_v17/legislation.xsd
@@ -29,7 +29,7 @@
 
     <xs:complexType name="codeType">
         <xs:group ref="valuesHistory"/>
-        <xs:attribute name="code"  type="xs:string" use="required"/>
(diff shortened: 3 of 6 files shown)


openfisca_core/scripts/migrations/v16_2_to_v17/xml_to_yaml.py
```diff
@@ -260,7 +266,7 @@ def write_yaml(node, path):
         raise ValueError('Unknown type {}'.format(node_type))
 
 
-def write_legislation(legislation_xml_info_list, path):
+def write_parameters(legislation_xml_info_list, path):
```
Anna-Livia: How is this function used ? It never seems to be called ...

openfisca_core/scripts/migrations/v16_2_to_v17/xml_to_yaml.py
```diff
@@ -80,7 +80,13 @@ def transform_values_history(children, value_format):
         elif child.tag == 'END':
             values[date] = {'value': None}
         elif child.tag == 'PLACEHOLDER':
-            values[date] = "expected"
+            values[date] = 'expected'
+
+        if 'reference' in child.attrib:
+            if values[date] == 'expected':
+                values[date] = {'expected': None, 'reference': child.attrib['reference']}
```
Anna-Livia: why is `'expected': None`? It seems to me that it would be more appropriate to have a syntax that says that the `value[date]`is **in fact** expected.

openfisca_core/scripts/migrations/v16_2_to_v17/xml_to_yaml.py
```diff
@@ -80,7 +80,13 @@ def transform_values_history(children, value_format):
         elif child.tag == 'END':
             values[date] = {'value': None}
```
Anna-Livia: Can't an `END` have a reference ?

openfisca_core/scripts/migrations/v16_2_to_v17/xml_to_yaml.py
```diff
@@ -80,7 +80,13 @@ def transform_values_history(children, value_format):
...
         elif child.tag == 'PLACEHOLDER':
-            values[date] = "expected"
+            values[date] = 'expected'
+
+        if 'reference' in child.attrib:
+            if values[date] == 'expected':
+                values[date] = {'expected': None, 'reference': child.attrib['reference']}
+            else:
+                values[date]['reference'] = child.attrib['reference']
```
Anna-Livia: Why is this not applied to all cases ? What is the different between this one and the one above ?

openfisca_core/scripts/migrations/v16_2_to_v17/xml_to_yaml.py
michelbl: A `END` node can have a reference atribute.
https://github.com/openfisca/openfisca-core/blob/9f08be7ee97848f0f8f177c946486b395e6aa18e/openfisca_core/scripts/migrations/v16_2_to_v17/xml_to_yaml.py#L85

openfisca_core/scripts/migrations/v16_2_to_v17/xml_to_yaml.py
michelbl: The admissible syntaxes for expected parameters were decided here : https://github.com/openfisca/openfisca-core/issues/523#issuecomment-313648250

openfisca_core/scripts/migrations/v16_2_to_v17/xml_to_yaml.py
michelbl: If tag is END or VALUE, values[date] is a dict. If tag is PLACEHOLDER, it is the string "expected".
This object is then serialized in YAML. The string "expected" is prefered over {"expected": None} because the following YAML is ugly : 
```yaml
    2017-01-01:
        expected: None
```

openfisca_core/scripts/migrations/v16_2_to_v17/xml_to_yaml.py
michelbl: It is used by country-specific migration scripts (openfisca-core/openfisca_core/scripts/migrations/v16_2_to_v17/xml_to_yaml_country_template.py for example)

Anna-Livia: Then shouldn't it be changed from `xml_to_yaml.write_legislation(legislation_xml_info_list, target_path)` to ` xml_to_yaml.write_parameters(legislation_xml_info_list, target_path)` in that file ?

michelbl: done

Issue #573, opened 2017-09-14
Allow legislation dynamic requesting


fpagnoux:
In france, some parameters are [not loaded the standard way](https://github.com/openfisca/openfisca-france/issues/818) because the current way of handling parameters doesn't scale . To use them the "proper" way, we would have to do this:

 :scream:

```py
code_insee = menage('code_insee', period)
taux = 
(code_insee == "75001") * p.tx_transport.code_75001 +
(code_insee == "75002") * p.tx_transport.code_75002 +
(code_insee == "75003") * p.tx_transport.code_75003 +
(code_insee == "75004") * p.tx_transport.code_75004 +
(code_insee == "75005") * p.tx_transport.code_75005 +
...
```

This can be solved allowing to dynamically request the parameters. For instance:

```py
code_insee = menage('code_insee', period)
taux = p.tx_transport[code_insee]
```

Keep in mind than both `code_insee` and `taux` are vectors.

Some exploratory work has been done earlier :

See [POC on core](https://github.com/openfisca/openfisca-core/compare/dynamic-leg-computing) et [impact on france](https://github.com/openfisca/openfisca-france/compare/master...simplify-al). This uses numpy `recarrays`

PR #575, opened 2017-09-21
Enable to calculate parameters according to a variable

fpagnoux:
Connected to #573 

Enable to calculate parameters according to a variable

For instance, if a parameter `rate` depends on a variable `zone` that can take the values `z1` or `z2`:

In `rate.yaml`:

```
z1:
  values:
    '2015-01-01':
      value: 100
z2:
  values:
    '2015-01-01':
      value: 200
```

Then it is now possible to use:

```py
zone = household('zone', period)
rate = parameters(period).rate[zone]
```

See example in france : https://github.com/openfisca/openfisca-france/pull/823

----

Todo:

- [x] Add the doc page and update the `xxx` link
- [x] Basic perf test for a big population


fpagnoux:
I'm not sure about how to name this feature (in the doc, the error messages, and the code)
- I used the expression "fancy indexing". This is a [current expression](http://www.scipy-lectures.org/intro/numpy/array_object.html#fancy-indexing) in the numpy world, but I'm not sure it's the best choice for OpenFisca users. 
- The more official version is ["advanced indexing"](https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html#advanced-indexing). 
- We also called this feature "dynamic parameter computation" at some point. 
- "Vectorial parameters computing" could be another option.

@Anna-Livia @sandcha @benjello @michelbl ideas ?

fpagnoux:
## Performance test summary

**TLDR** : Using fancy indexing costs around `30%` more time than the classical way _for one formula_. The overall impact on a simulation should be much lower, as only a few functions need fancy indexing. This cost seems reasonable compared to the benefits of the feature, which significantly reduces some formulas complexity, and could solve the recurring issue of combinatory explosion in OpenFisca.


### Could we go faster ?

Currently, if the user tries to compute `parameters.rate[zone]` with `zone = ['zone1', 'zone2', 'unexpected_key']`, where `parameters.rate.unexpected_key` doesn't exist, a `ParameterNotFound` error is raised, with an explicit error message.

We could reduce the overhead to around `10%` if we changed the behaviour to : no error is raised, but  `parameters.rate[zone] = [200, 250, nan]`. 

That would however be much less nice to users, as they would have to figure out how a `nan` happened to sneak in their result.

### Features we had to give up

Currently, if in `parameters.rate` we have three nodes (such as `zone1`, `zone2`, `zone3`) and a few leafs (such as `colocation`, `couple`), I will **not** be able to use `parameters.rate[zone]`, even if `zone = ['zone1', 'zone2']`.
The children of the node we want to use fancy indexing on **must be perfectly homogenous**. 

There was a way to go around this limitation, but it made the calculation 5 times slower for one formula, which seemed to much.

openfisca_core/parameters.py
```diff
@@ -559,17 +561,178 @@ def __getattr__(self, key):
...
+                for node, name in zip(nodes, names)[1:]:
+                    if not isinstance(node, ParameterNodeAtInstant):
+                        raise_type_inhomogeneity_error(first_name, name)
+                    first_node_keys = first_node._children.keys()
+                    node_keys = node._children.keys()
+                    if not first_node_keys == node_keys:
+                        missing_keys = set(first_node_keys).difference(node_keys)
+                        if missing_keys:
+                            raise_key_inhomogeneity_error(first_name, name, missing_keys.pop())
+                        missing_key = set(node_keys).difference(first_node_keys).pop()
```
Anna-Livia: Would it be helpful to explicitly put an `else`here ?

CHANGELOG.md
```diff
@@ -1,5 +1,31 @@
 # Changelog
 
+## 17.2.0 [#570](https://github.com/openfisca/openfisca-core/pull/570)
```
Anna-Livia: Why is it a breaking change ? It seams it only add new possibilities

fpagnoux: It's not. A breaking change would be `18.0.0`. `17.2.0` means a new feature.

openfisca_core/parameters.py
```diff
@@ -559,17 +561,178 @@ def __getattr__(self, key):
...
+            return VectorialParameterNodeAtInstant.build_from_node(self)[key]
+        return self._children[key]
 
     def __iter__(self):
         return iter(self._children)
 
 
+class VectorialParameterNodeAtInstant(object):
+    """
+        Parameter node of the legislation at a given instant which has been vectorized: it may differ for each entity.
```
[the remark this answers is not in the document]
fpagnoux: The way I see it (you can challenge the vision of course):

Ideally, the user doesn't care much about whether a parameter is vectorized.
They just do `P.aides_logement[zone]`, and they are happy that is works.
If it doesn't, then only they are sent on the doc to see why what they tried to do is not allowed.

openfisca_core/parameters.py
```diff
@@ -501,7 +503,7 @@ def add_child(self, name, child):
         :param child: The new child, an instance of :any:`Scale` or :any:`Parameter` or :any:`ParameterNode`.
         """
         if name in self.children:
-            raise ValueError("{} has already a child named {}".format(self.name, name)).encode('utf-8')
+            raise ValueError("{} has already a child named {}".format(self.name, name).encode('utf-8'))
```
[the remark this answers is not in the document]
fpagnoux: Because `ValueError(string).encode('utf-8')` doesn't make sense (you can encode a string, not an exception).


`ValueError(string.encode('utf-8'))` makes sense

openfisca_core/parameters.py
```diff
@@ -559,17 +561,178 @@ def __getattr__(self, key):
...
         return iter(self._children)
 
 
+class VectorialParameterNodeAtInstant(object):
+    """
+        Parameter node of the legislation at a given instant which has been vectorized: it may differ for each entity.
+    """
+
+    @classmethod
+    def build_from_node(cls, node):
```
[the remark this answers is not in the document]
fpagnoux: `cls` represents the class in a python [class method](https://julien.danjou.info/blog/2013/guide-python-static-class-abstract-methods). But looking at that doc, it seems I should have used a `staticmethod` instead 👍 .

openfisca_core/parameters.py
fpagnoux: In this case, why not.
Added a comment too.

openfisca_core/parameters.py
```diff
@@ -559,17 +561,178 @@ def __getattr__(self, key):
...
+        cls.check_node_vectorisable(node)
+        subnodes_name = node._children.keys()
+        vectorial_subnodes = tuple([
```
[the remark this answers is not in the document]
fpagnoux: Ok, added some comments

openfisca_core/parameters.py
```diff
@@ -476,6 +477,7 @@ def __init__(self, name = "", directory_path = None, data = None, file_path = No
             data.pop('reference', None)
             data.pop('description', None)
             for child_name, child in data.items():
+                child_name = str(child_name)
```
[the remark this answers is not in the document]
fpagnoux: Where would you like this doc to be put ?

openfisca_core/formula_helpers.py
```diff
@@ -45,3 +45,12 @@ def switch(conditions, value_by_condition):
...
         ]
     return np.select(condlist, value_by_condition.values())
+
+
+def concat(this, that):
+    if isinstance(this, np.ndarray):
+        this = this.astype('str')
+    if isinstance(that, np.ndarray):
+        that = that.astype('str')
+
```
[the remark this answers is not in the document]
fpagnoux: `TypeError: cannot concatenate 'str' and 'dict' objects`

openfisca_core/parameters.py
```diff
@@ -559,17 +561,183 @@ def __getattr__(self, key):
...
+    def build_from_node(node):
+        VectorialParameterNodeAtInstant.check_node_vectorisable(node)
+        subnodes_name = node._children.keys()
+        # Recursively vectorize the children of the node
+        vectorial_subnodes = tuple([
+            VectorialParameterNodeAtInstant.build_from_node(node[subnode_name]).vector if isinstance(node[subnode_name], ParameterNodeAtInstant) else node[subnode_name]
+            for subnode_name in subnodes_name
+            ])
+        # A vectorial node is a wrapper around a numpy recarray
+        # We first build the recarray
```
Anna-Livia: 👍

openfisca_core/parameters.py
```diff
@@ -559,17 +561,183 @@ def __getattr__(self, key):
...
+                for node, name in zip(nodes, names)[1:]:
+                    if not isinstance(node, ParameterNodeAtInstant):
+                        raise_type_inhomogeneity_error(first_name, name)
+                    first_node_keys = first_node._children.keys()
+                    node_keys = node._children.keys()
+                    if not first_node_keys == node_keys:
+                        missing_keys = set(first_node_keys).difference(node_keys)
+                        if missing_keys:  # If the firs_node has a key that node hasn't
+                            raise_key_inhomogeneity_error(first_name, name, missing_keys.pop())
+                        else:  # If the node has a key that first_node hasn't
```
Anna-Livia: >  If the node has a key that first_node doesn't have

openfisca_core/parameters.py
```diff
@@ -559,17 +561,183 @@ def __getattr__(self, key):
...
+            if isinstance(first_node, ParameterNodeAtInstant):
+                children = extract_named_children(first_node)
+                for node, name in zip(nodes, names)[1:]:
+                    if not isinstance(node, ParameterNodeAtInstant):
+                        raise_type_inhomogeneity_error(first_name, name)
+                    first_node_keys = first_node._children.keys()
+                    node_keys = node._children.keys()
+                    if not first_node_keys == node_keys:
+                        missing_keys = set(first_node_keys).difference(node_keys)
+                        if missing_keys:  # If the firs_node has a key that node hasn't
```
Anna-Livia: > If the firs_node has a key that node doesn't have

openfisca_core/parameters.py
```diff
@@ -559,17 +561,183 @@ def __getattr__(self, key):
...
         return iter(self._children)
 
 
+class VectorialParameterNodeAtInstant(object):
+    """
+        Parameter node of the legislation at a given instant which has been vectorized.
+        Vectorized parameters allow requests such as parameters.housing_benefit[zipcode], where zipcode is a vector
+    """
+
+    @staticmethod
```
sandcha: Why is it a `@staticmethod` and not `@classmethod`. Shouldn't it be applied to the new declared class? Or, if it's mainly a way to unite functions, why is `contains_nan` outside of this class? 🤔

openfisca_core/scripts/test_perf_fancy_indexing.py
```diff
@@ -0,0 +1,75 @@
+import numpy as np
```
sandcha: Rename file to `measure_performances_fancy_indexing` to unify with other filenames in this directory?

tests/core/parameters_fancy_indexing/test_fancy_indexing.py
```diff
@@ -0,0 +1,146 @@
+# -*- coding: utf-8 -*-
+
```
sandcha: ❤️

openfisca_core/parameters.py
```diff
@@ -559,17 +561,183 @@ def __getattr__(self, key):
         param_name = _compose_name(self._name, key)
         raise ParameterNotFound(param_name, self._instant_str)
 
-    def __getitem__(self, key):  # deprecated
-        return getattr(self, key)
+    def __getitem__(self, key):
```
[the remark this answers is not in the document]
fpagnoux: Because it's not deprecated anymore. What kind of explanation would you like ?

openfisca_core/parameters.py
fpagnoux: > Why is it a @staticmethod and not @classmethod

Because it doesn't rely on the class. But it's clearly related to `VectorialParameterNodeAtInstant`, as this method is a builder for this class. However, `contains_nan` is not directly related to `VectorialParameterNodeAtInstant`.

openfisca_core/parameters.py
```diff
@@ -559,17 +561,183 @@ def __getattr__(self, key):
...
+        # If fancy indexing is used, cast to a vectorial node
+        if isinstance(key, np.ndarray):
+            return VectorialParameterNodeAtInstant.build_from_node(self)[key]
+        return self._children[key]
 
     def __iter__(self):
         return iter(self._children)
 
 
+class VectorialParameterNodeAtInstant(object):
```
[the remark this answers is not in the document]
fpagnoux: Agreed, but I can't merge `VectorialParameterNodeAtInstant` and `ParameterNodeAtInstant` as the latter can contain `Baremes`, which are not easily vectorisable.

Maybe we can merge this latter if we simplify baremes.

CHANGELOG.md
```diff
@@ -1,5 +1,31 @@
...
+
+- Enable to calculate parameters according to a variable
```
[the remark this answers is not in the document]
fpagnoux: Done

PR #581, opened 2017-10-04
Track IPs from API visitors

Anna-Livia:
Connected to openfisca/openfisca-france#821

#### Technical changes

- By adding your Piwik AUTH token to the gunicorn environment variable when serving the API, you can track your API visitor's by IP on Piwik

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ The tracker is activated when these two environment variables are set:
 
 * `TRACKER_URL`: An URL ending with `piwik.php`. It defines the Piwik instance that will receive the tracking information. To use the main OpenFisca Piwik instance, use `https://stats.data.gouv.fr/piwik.php`.
 * `TRACKER_IDSITE`: An integer. It defines the identifier of the tracked site on your Piwik instance. To use the main OpenFisca piwik instance, use `4`.
+* `TRACKER_TOKEN`: A string. It defines the Piwik API Authentification token to differentiate API calls based on the user IP. Otherwise, all API calls will seem to come from your server. The Piwik API Authentification token can be found in your Piwik interface, when you are logged.
 
 For instance, to run the Web API with the mock country package `openfisca_country_template` and the tracker activated, run:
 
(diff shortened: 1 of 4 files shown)


openfisca_web_api_preview/app.py
```diff
@@ -128,7 +131,7 @@ def apply_headers(response):
     @app.after_request
     def track_requests(response):
         if tracker:
-            tracker.track(request.url)
+            tracker.track(request.url, request.remote_addr)
```
fpagnoux: The IP address should not be sent if there is a ["Do not track" header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/DNT)

openfisca_web_api_preview/app.py
```diff
@@ -128,7 +129,10 @@ def apply_headers(response):
     @app.after_request
     def track_requests(response):
         if tracker:
-            tracker.track(request.url)
+            if 'dnt' in request.headers and request.headers['dnt'] == '1':
```
fpagnoux: `request.headers` seems to have the same API than a python `dict`.
You can thus just use `if request.headers.get('dnt')`.

If the 'DNT' header is not set, the value will be None, evaluated to false.
If the 'DNT' header is set to 0, the value will be 0, evaluated to false.
if the 'DNT' header is set to 1, true, TRUE, the value will be evaluated to true.

README.md
```diff
@@ -55,6 +55,7 @@ The tracker is activated when these two environment variables are set:
 
 * `TRACKER_URL`: An URL ending with `piwik.php`. It defines the Piwik instance that will receive the tracking information. To use the main OpenFisca Piwik instance, use `https://stats.data.gouv.fr/piwik.php`.
 * `TRACKER_IDSITE`: An integer. It defines the identifier of the tracked site on your Piwik instance. To use the main OpenFisca piwik instance, use `4`.
+* `TRACKER_TOKEN`: A string. It defines the Piwik API Authentification token to differentiate API calls based on the user IP. Otherwise, all API calls will seem to come from your server. The Piwik API Authentification token can be found in your Piwik interface, when you are logged.
```
fpagnoux: > 👍

Issue #515, opened 2017-05-18
Debug calculations


fpagnoux:
As a business expert,
I can get get all the steps of a given calculation, to be able to understand it and debug it.

Details to be determined, but this debugging tool must be user-friendly and functional.

fpagnoux:
Basic solution : 

- Be isofunctional to the legacy API

fpagnoux:
Process:

- Show examples to web api users
- Organise a workshop
-  Define
     - A response format
     - A requesting format
     - A url
- Deliver :rocket:

PR #582, opened 2017-10-09
Add /trace route

fpagnoux:
Connected to #515 
Depends on https://github.com/openfisca/country-template/pull/22

#### New features

- Introduce new Web API route `/trace` to get a simulation's detailed calculation steps
  - See the [swagger documentation for OpenFisca-France](https://legislation.openfisca.fr/swagger)

- Introduce `simulation.tracer.print_computation_log` to print a simulation's detailed calculation steps
  - This is available if the `trace` argument has been set to `True` in the `Simulation` constructor.
  - See the [reference documentation](https://openfisca.readthedocs.io/en/latest/tracer.html) about the tracer.

_Warning: Both of these features are still currently under experimentation. You are very welcome to use them and send us precious feedback, but keep in mind that the way they are used and the results they give might change without any major version bump._

#### Breaking changes

- Deprecate and remove:
  - `simulation.traceback`
  - `simulation.stack_trace`
  - `simulation.find_traceback_step`
  - `simulation.stringify_input_variables_infos`
  - `simulation.stringify_variable_for_period_with_array`

- Remove argument `debug_all` from:
  - `Scenario.new_simulation` method
  - `Simulation` constructor


openfisca_core/tools/test_runner.py
```diff
@@ -110,12 +110,13 @@ def run_tests(tax_benefit_system, paths, options = {}):
     +-------------------------------+-----------+-------------------------------------------+
 
     """
+    argv = sys.argv[:1]  # Nose crashes if it gets any unexpected argument.
+    if options.get('verbose'):
```
Anna-Livia: Is this documented somewhere ?

openfisca_core/tracers.py
```diff
@@ -0,0 +1,135 @@
...
+            self.trace[parent]['dependencies'].append(key)
+        else:  # The variable has been requested by the client
+            self.requested_calculations.add(key)
+
+        if not self.trace.get(key):
+            self.trace[key] = {'dependencies': []}
+        self.stack.append(key)
+        self._computation_log.append((key, len(self.stack)))
+
+    def record_calculation_end(self, variable_name, period, result, **parameters):
```
Anna-Livia: result doesn't seem to be documented

openfisca_core/tracers.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        key = self._get_key(variable_name, period, **parameters)
+        expected_key = self.stack.pop()
+
+        if not key == expected_key:
+            raise ValueError(
+                u"Something went wrong with the simulation tracer: result of '{0}' was expected, got results for '{1}' instead. This does not make sense as the last variable we started computing was '{0}'."
+                .format(expected_key, key).encode('utf-8')
+                )
+        self.trace[key]['value'] = result.tolist()
```
Anna-Livia: Would it be hepful to indicate that `.tolist()` transforms nparray into python lists ?

fpagnoux: Ok

openfisca_core/tools/test_runner.py
fpagnoux: Yep, but it was broken. Will be repaired by this PR.

http://openfisca.readthedocs.io/en/trace/test_runner.html
http://openfisca.readthedocs.io/en/trace/openfisca-run-test.html

Issue #576, opened 2017-09-26
Serve extensions in the API


fpagnoux:
The legacy API was able to serve a tax and benefit system with extensions, with:

```sh
openfisca-serve --country_package openfisca_france --extensions openfisca_paris
```

The new api doesn't do it, which prevents user such as mes-aides from using it.

fpagnoux:
It could also be useful to expose in the API whether a variable has been defined by an extension.

fpagnoux:
Dont forget to write some doc about it

fpagnoux:
Some info that you may (or not) need to treat this issue:

# CLI in python

- CLI commands associated to a Python package are defined [in the setup.py](https://github.com/openfisca/openfisca-web-api/blob/master/setup.py#L41)
- To handle CLI arguments, use [argparse](https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/scripts/run_test.py#L13). It builds the `--help` for you, and offers many features.
- There is [a helper](https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/scripts/__init__.py#L19) to add to your argument parser the usual options to build your tax and benefit system (which country, which extensions, which reforms)

fpagnoux:
### Extensions

- There is already [a helper](https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/scripts/__init__.py#L27) to build a tax and benefit system from a country package name, a list of reform, and a list of extensions.

fpagnoux:
If we don't want to invest in a CLI tool, other options that can be considered:

- Keep using environment variables. This implies encode a list of strings in a variable, which is not usual.
- Use a configuration file. There is probably a standard way of doing it with flask.

PR #583, opened 2017-10-09
Serve extensions in the API

sandcha:
Connected to #576 

#### New features

- Introduce `openfisca-serve` for new API
  - Allows for serving OpenFisca extensions with `-e` option. Example: openfisca-serve -c openfisca_country_template -e openfisca_extension_template

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,4 +7,4 @@ test:
 	nosetests
 
 api:
-	COUNTRY_PACKAGE=openfisca_country_template gunicorn "openfisca_web_api_preview.app:create_app()" --bind localhost:5000 --workers 3
+	openfisca-serve -c openfisca_country_template -e openfisca_extension_template
--- a/Makefile
+++ b/Makefile
@@ -7,4 +7,4 @@ test:
 	nosetests
 
 api:
-	openfisca-serve -c openfisca_country_template -e openfisca_extension_template
+	openfisca-serve --country_package openfisca_country_template --extensions openfisca_extension_template
--- a/Makefile
+++ b/Makefile
@@ -7,4 +7,4 @@ test:
 	nosetests
 
 api:
-	openfisca-serve --country_package openfisca_country_template --extensions openfisca_extension_template
+	openfisca serve --country-package openfisca_country_template --extensions openfisca_extension_template
--- a/openfisca_core/parameters.py
+++ b/openfisca_core/parameters.py
@@ -22,7 +22,7 @@ try:
     from yaml import CLoader as Loader
 except ImportError:
     log.warning(
-        "libyaml is not installed in your environement. This can make OpenFisca slower to start. Once you have installed libyaml, run 'pip uninstall pyyaml && pip install pyyaml' so that it is used in your Python environement.")
+        "libyaml is not installed in your environement. This can make OpenFisca slower to start. Once you have installed libyaml, run 'pip uninstall pyyaml && pip install pyyaml' so that it is used in your Python environement." + os.linesep)
     from yaml import Loader
 
 
--- a/openfisca_web_api_preview/app.py
+++ b/openfisca_web_api_preview/app.py
@@ -36,9 +36,11 @@ def init_tracker(url, idsite, tracker_token):
 
 
 def create_app(country_package = os.environ.get('COUNTRY_PACKAGE'),
+               extensions = os.environ.get('EXTENSIONS'),
                tracker_url = os.environ.get('TRACKER_URL'),
                tracker_token = os.environ.get('TRACKER_TOKEN'),
                tracker_idsite = os.environ.get('TRACKER_IDSITE')):
+
     if country_package is None:
         raise ValueError(
             u"You must specify a country package to start the API. "
@@ -57,7 +59,7 @@ def create_app(country_package = os.environ.get('COUNTRY_PACKAGE'),
 
     app.url_map.strict_slashes = False  # Accept url like /parameters/
 
-    data = build_data(country_package)
+    data = build_data(country_package, extensions)
 
     @app.route('/parameters')
     def get_parameters():
--- a/openfisca_web_api_preview/app.py
+++ b/openfisca_web_api_preview/app.py
@@ -36,10 +36,11 @@ def init_tracker(url, idsite, tracker_token):
 
 
 def create_app(country_package,
-               extensions,
-               tracker_url,
-               tracker_idsite,
-               tracker_token):
+               extensions = None,
+               tracker_url = None,
+               tracker_idsite = None,
+               tracker_token = None
+               ):
 
     if country_package is None:
         raise ValueError(
--- a/openfisca_web_api_preview/app.py
+++ b/openfisca_web_api_preview/app.py
@@ -36,14 +36,11 @@ def init_tracker(url, idsite, tracker_token):
         log.warn(message)
 
 
-def create_app(country_package,
-               extensions = None,
+def create_app(tax_benefit_system,
                tracker_url = None,
                tracker_idsite = None,
                tracker_token = None
                ):
-
-    # country_package defined here or detected in openfisca_core.scripts.build_tax_benefit_system.
 
     if not tracker_url or not tracker_idsite:
         tracker = None
@@ -56,7 +53,7 @@ def create_app(country_package,
 
     app.url_map.strict_slashes = False  # Accept url like /parameters/
 
-    data = build_data(country_package, extensions)
+    data = build_data(tax_benefit_system)
 
     @app.route('/parameters')
     def get_parameters():
--- a/openfisca_web_api_preview/app.py
+++ b/openfisca_web_api_preview/app.py
@@ -37,9 +37,9 @@ def init_tracker(url, idsite, tracker_token):
 
 def create_app(country_package,
                extensions,
-               tracker_url = os.environ.get('TRACKER_URL'),
-               tracker_token = os.environ.get('TRACKER_TOKEN'),
-               tracker_idsite = os.environ.get('TRACKER_IDSITE')):
+               tracker_url,
+               tracker_idsite,
+               tracker_token):
 
     if country_package is None:
         raise ValueError(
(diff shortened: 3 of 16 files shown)


Makefile
```diff
@@ -7,4 +7,4 @@ test:
 	nosetests
 
 api:
-	COUNTRY_PACKAGE=openfisca_country_template gunicorn "openfisca_web_api_preview.app:create_app()" --bind localhost:5000 --workers 3
+	openfisca-serve -c openfisca_country_template -e openfisca_extension_template
```
fpagnoux: Always use long options in a script, for readability

openfisca_web_api_preview/app.py
```diff
@@ -34,13 +34,15 @@ def init_tracker(url, idsite):
...
+def create_app(country_package,
+               extensions = None,
+               tracker_url = None,
+               tracker_idsite = None):
+
     if country_package is None:
         raise ValueError(
-            u"You must specify a country package to start the API. "
-            u"For instance, `COUNTRY_PACKAGE=openfisca_france flask run`"
+            u"You must specify a country package with `-c` option to start the API. "
```
fpagnoux: Prefer long options (`--contry_template`) for readability

openfisca_web_api_preview/loader/__init__.py
```diff
@@ -13,11 +13,17 @@ def extract_description(items):
...
     tax_benefit_system = build_tax_benefit_system(country_package_name)
     country_package_metadata = tax_benefit_system.get_package_metadata()
+
+    if extensions is not None:
+        for extension in extensions:
+            tax_benefit_system.load_extension(extension)
+
     parameters = build_parameters(tax_benefit_system)
     variables = build_variables(tax_benefit_system, country_package_metadata)
+
```
fpagnoux: Why adding a line break here ?

openfisca_web_api_preview/loader/__init__.py
```diff
@@ -13,11 +13,17 @@ def extract_description(items):
         }
 
 
-def build_data(country_package_name):
+def build_data(country_package_name, extensions):
     tax_benefit_system = build_tax_benefit_system(country_package_name)
     country_package_metadata = tax_benefit_system.get_package_metadata()
+
+    if extensions is not None:
```
fpagnoux: Any reason not to use [this common function](https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/scripts/__init__.py#L27) ?

openfisca_web_api_preview/app.py
fpagnoux: But a regular user has only one country package installed on their environment. Why not attempt a [country package detection](https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/scripts/__init__.py#L29) ?

tests/web_api/test_extensions.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
```
fpagnoux: > 👍

openfisca_web_api_preview/scripts/serve.py
```diff
@@ -0,0 +1,54 @@
...
+import argparse
+from gunicorn.app.base import BaseApplication
+from gunicorn.six import iteritems
+
+from openfisca_core.scripts import add_tax_benefit_system_arguments, detect_country_package
+from ..app import create_app
+
+
+DEFAULT_PORT = '5000'
+BASE_CONF = {
```
fpagnoux: `DEFAULT_CONF` ?

openfisca_web_api_preview/scripts/serve.py
```diff
@@ -0,0 +1,54 @@
...
+    'bind': '%s:%s' % ('127.0.0.1', DEFAULT_PORT),
```
fpagnoux: Any reason not to use the more usual `'127.0.0.1:{}'.format(DEFAULT_PORT)` ?

openfisca_web_api_preview/scripts/serve.py
```diff
@@ -0,0 +1,54 @@
...
+    'workers': '3',
+}
+
+
+def new_configured_app():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-p', '--port', action = 'store', default = DEFAULT_PORT, help = "port to serve on", type = int)
+    parser.add_argument('-t', '--tracker_url', action = 'store', help = "tracking service url", type = str)
+    parser.add_argument('-s', '--tracker_idsite', action = 'store', help = "tracking service id site", type = int)
```
fpagnoux: > `-s` for `--tracker_idsite` is not obvious to me.
>
> It would be fine IMO if the two trackers options had no short version (only `--tracker_url` and `--tracker_idsite`)

openfisca_web_api_preview/scripts/serve.py
```diff
@@ -0,0 +1,54 @@
...
+    parser = add_tax_benefit_system_arguments(parser)
+    args = parser.parse_args()
+
+    app = create_app(args.country_package, args.extensions, args.tracker_url, args.tracker_idsite)
+    return app
+
+
+class StandaloneApplication(BaseApplication):
+
+    def __init__(self, app, options=None):
```
fpagnoux: In OpenFisca, we add spaces around `=`

openfisca_web_api_preview/scripts/serve.py
```diff
@@ -0,0 +1,54 @@
...
+        self.options = options or {}
+        self.application = app
+        super(StandaloneApplication, self).__init__()
+
+    def load_config(self):
+        config = dict([(key, value) for key, value in iteritems(self.options)
```
fpagnoux:
```py
{key: value for key, value in iteritems(self.options) if key in self.cfg.settings and value is not None}
```
should do the same

fpagnoux: If `self.options` is just a `dict`, use `self.options.iteritems()`

openfisca_web_api_preview/scripts/serve.py
```diff
@@ -0,0 +1,54 @@
...
+                       if key in self.cfg.settings and value is not None])
+        for key, value in iteritems(config):
```
fpagnoux: `config.iteritems()`

openfisca_web_api_preview/scripts/serve.py
```diff
@@ -0,0 +1,54 @@
...
+            self.cfg.set(key.lower(), value)
```
fpagnoux: This function seems very complicated for what it does. What about:

```
    def load_config(self):
        for key, value in self.options.iteritems():
            if key in self.cfg.settings and value is not None:
                self.cfg.set(key.lower(), value)
```

But why checking `if key in self.cfg.settings and value is not None` ?
- if the user somehow pass a wrong config item, is it really a good idea to filter out the wrong config item, and not throw any error ?
- In which case `value` can be `None`, in the current `implem` ?

openfisca_web_api_preview/loader/__init__.py
sandcha: To separate the work on the package carcass from the inner building. 😅 
Removed.

openfisca_web_api_preview/scripts/serve.py
sandcha: A bad taste for novelty? 😅 
Replaced.

openfisca_web_api_preview/scripts/serve.py
sandcha: Short version removed.

openfisca_web_api_preview/scripts/serve.py
sandcha: Function updated.
No information is logged if `key not in self.cfg.settings` as it doesn't accept `port` or any other python variable that might be useful for our users.

But an information is now logged for `value is None` and the execution isn't interrupted. If no value is given in the configuration file, an error is thrown as the python configuration file won't be valid.

openfisca_web_api_preview/scripts/serve.py
sandcha: gunicorn.six.iteritems seems more secure but no good documentation was found. 😕

openfisca_web_api_preview/scripts/serve.py
```diff
@@ -0,0 +1,113 @@
...
+                log.debug('Undefined value for key `{}`.'.format(key))
+
+            if key in self.cfg.settings and value is not None:
+                self.cfg.set(key.lower(), value)
+
+    def load(self):
+        return self.application
+
+
+def main(parser):
```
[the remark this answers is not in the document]
fpagnoux: Got it, you're calling it from `openfisca_serve.py`

openfisca_web_api_preview/scripts/serve.py
```diff
@@ -0,0 +1,113 @@
...
+        'workers': GUNICORN_WORKERS_NUMBER,
+        }
+    configuration = read_user_configuration(configuration, command_line_parser)
+
+    app = create_app(configuration['country_package'], configuration['extensions'], configuration['tracker_url'], configuration['tracker_idsite'])
+    StandaloneApplication(app, configuration).run()
+
+
+if __name__ == '__main__':
+    sys.exit(main())
```
fpagnoux: This probably doesn't work, as main excepts an argument

openfisca_core/parameters.py
```diff
@@ -20,7 +20,7 @@
     from yaml import CLoader as Loader
 except ImportError:
     log.warning(
-        "libyaml is not installed in your environement. This can make OpenFisca slower to start. Once you have installed libyaml, run 'pip uninstall pyyaml && pip install pyyaml' so that it is used in your Python environement.")
+        "libyaml is not installed in your environement. This can make OpenFisca slower to start. Once you have installed libyaml, run 'pip uninstall pyyaml && pip install pyyaml' so that it is used in your Python environement." + os.linesep)
```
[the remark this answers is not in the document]
sandcha: To improve the reading experience of outputs. `libyaml` is needed for France but doesn't seem necessary for other environments.

sandcha:
@fpagnoux Ready to merge if you agree.

PR #585, opened 2017-10-10
Track the real visitor IP

fpagnoux:
Connected to openfisca/openfisca-france#821

🎉 

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+### 18.0.2 [#585](https://github.com/openfisca/openfisca-core/pull/585)
+
+- Track the real visitor IP in the web API
+  - Handle the nginx proxy
+
 ### 18.0.1 [#584](https://github.com/openfisca/openfisca-core/pull/584)
 
 - Add a link in `openAPI.yml` to the OpenFisca documentation page explaining the web API inputs and outputs for the /calculate route.
--- a/openfisca_web_api_preview/app.py
+++ b/openfisca_web_api_preview/app.py
@@ -3,6 +3,7 @@
 import os
 from os import linesep
 from flask import Flask, jsonify, abort, request, make_response
+from werkzeug.contrib.fixers import ProxyFix
 from flask_cors import CORS
 import dpath
 
@@ -51,6 +52,7 @@ def create_app(country_package = os.environ.get('COUNTRY_PACKAGE'),
         tracker = init_tracker(tracker_url, tracker_idsite, tracker_token)
 
     app = Flask(__name__)
+    app.wsgi_app = ProxyFix(app.wsgi_app, num_proxies = 1)  # Fix request.remote_addr to get the real client IP address
     CORS(app, origins = '*')
 
     app.url_map.strict_slashes = False  # Accept url like /parameters/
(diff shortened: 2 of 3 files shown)


fpagnoux:
Ignoring the integration tests from the legislation explorer will be much harder : seems like they can use any AWS IP. That's a lot to exclude...

fpagnoux:
Source: http://esd.io/blog/flask-apps-heroku-real-ip-spoofing.html

Anna-Livia:
If we only use the ProxyFix function of the library, would it make sense to have the function in our code ?
```python
class ProxyFix(object):

    """This middleware can be applied to add HTTP proxy support to an
    application that was not designed with HTTP proxies in mind.  It
    sets `REMOTE_ADDR`, `HTTP_HOST` from `X-Forwarded` headers.  While
    Werkzeug-based applications already can use
    :py:func:`werkzeug.wsgi.get_host` to retrieve the current host even if
    behind proxy setups, this middleware can be used for applications which
    access the WSGI environment directly.
    If you have more than one proxy server in front of your app, set
    `num_proxies` accordingly.
    Do not use this middleware in non-proxy setups for security reasons.
    The original values of `REMOTE_ADDR` and `HTTP_HOST` are stored in
    the WSGI environment as `werkzeug.proxy_fix.orig_remote_addr` and
    `werkzeug.proxy_fix.orig_http_host`.
    :param app: the WSGI application
    :param num_proxies: the number of proxy servers in front of the app.
    """

    def __init__(self, app, num_proxies=1):
        self.app = app
        self.num_proxies = num_proxies

    def get_remote_addr(self, forwarded_for):
        """Selects the new remote addr from the given list of ips in
        X-Forwarded-For.  By default it picks the one that the `num_proxies`
        proxy server provides.  Before 0.9 it would always pick the first.
        .. versionadded:: 0.8
        """
        if len(forwarded_for) >= self.num_proxies:
            return forwarded_for[-self.num_proxies]

    def __call__(self, environ, start_response):
        getter = environ.get
        forwarded_proto = getter('HTTP_X_FORWARDED_PROTO', '')
        forwarded_for = getter('HTTP_X_FORWARDED_FOR', '').split(',')
        forwarded_host = getter('HTTP_X_FORWARDED_HOST', '')
        environ.update({
            'werkzeug.proxy_fix.orig_wsgi_url_scheme':  getter('wsgi.url_scheme'),
            'werkzeug.proxy_fix.orig_remote_addr':      getter('REMOTE_ADDR'),
            'werkzeug.proxy_fix.orig_http_host':        getter('HTTP_HOST')
        })
        forwarded_for = [x for x in [x.strip() for x in forwarded_for] if x]
        remote_addr = self.get_remote_addr(forwarded_for)
        if remote_addr is not None:
            environ['REMOTE_ADDR'] = remote_addr
        if forwarded_host:
            environ['HTTP_HOST'] = forwarded_host
        if forwarded_proto:
            environ['wsgi.url_scheme'] = forwarded_proto
        return self.app(environ, start_response)
``` 
from `https://github.com/pallets/werkzeug/blob/master/werkzeug/contrib/fixers.py`

fpagnoux:
`werkzeug` is packaged with `flask`, importing it doesn't cost anyting :)

PR #598, opened 2017-12-04
Introduce new features to allow merge with taxipp

fpagnoux:
- Improve `Tracer`:

  - Introduce an `aggregate` option in [`tracer.print_computation_log`](http://openfisca.readthedocs.io/en/latest/tracer.html#openfisca_core.tracers.Tracer.print_computation_log) to handle large population simulations.
  - Introduce [`tracer.usage_stats`](http://openfisca.readthedocs.io/en/latest/tracer.html#openfisca_core.tracers.Tracer.usage_stats) to keep track of the number of times a variable is computed.

- Introduce methods to keep track of memory usage:

  - Introduce [`holder.get_memory_usage`](http://localhost:8000/holder.html#openfisca_core.holders.Holder.get_memory_usage)
  - Introduce `entity.get_memory_usage`
  - Introduce `simulation.get_memory_usage`

- Improve `Holder` public interface:

  - Enhance [`holder.delete_arrays`](http://openfisca.readthedocs.io/en/latest/holder.html#openfisca_core.holders.Holder.get_memory_usage) to be able to remove known values only for a specific period
  - Introduce [`holder.get_known_periods`](http://openfisca.readthedocs.io/en/latest/holder.html#openfisca_core.holders.Holder.get_known_periods)

- Re-introduce `taxscales.combine_tax_scales` to combine several tax scales.


openfisca_core/holders.py
```diff
@@ -251,6 +270,71 @@ def graph(self, edges, get_input_variables_and_parameters, nodes, visited):
...
+            >>>    }
+        """
+
+        usage = dict(
+            nb_cells_by_array = self.entity.count,
+            dtype = self.variable.dtype,
+            )
+
+        if self._array is not None:
+            # Only used when definition period is ETERNITY
```
Anna-Livia: Should ` self._array` be called `all_period_array`, or `eternity_array` ?
Otherwise, should we amend the comment to :
"# self._array is only used when definition period is ETERNITY"

openfisca_core/holders.py
```diff
@@ -251,6 +270,71 @@ def graph(self, edges, get_input_variables_and_parameters, nodes, visited):
...
+            usage.update(dict(
+                nb_arrays = 1,
+                total_nb_bytes = self._array.nbytes,
+                cell_size = self._array.itemsize,
+                ))
+            return usage
+        elif self._array_by_period is not None:
+            nb_arrays = sum([
```
Anna-Livia: > In what case are there dictionnaries in `self._array_by_period` ?

openfisca_core/holders.py
```diff
@@ -251,6 +270,71 @@ def graph(self, edges, get_input_variables_and_parameters, nodes, visited):
...
+                len(array_or_dict) if isinstance(array_or_dict, dict) else 1
+                for array_or_dict in self._array_by_period.itervalues()
+                ])
+            array = self._array_by_period.itervalues().next()
```
Anna-Livia: Why not just take the first item ?

openfisca_core/entities.py
```diff
@@ -364,6 +381,16 @@ def members_position(self):
...
         self._members_role = members_role
 
+    @property
+    def roles_count(self):
+        if self._roles_count is None:
+            self._roles_count = self.members_legacy_role.max() + 1
+        return self._roles_count
+
+    @roles_count.setter
+    def roles_count(self, roles_count):
```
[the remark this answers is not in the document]
fpagnoux: This is Python [property](https://www.programiz.com/python-programming/property).

The goal of this change is that if you call `entity.roles_count`, and that this attribute has not been set yet, then it will be deducted from `entity. members_legacy_role`

FI the attribute `roles_count ` is only needed for the "old" syntax.

openfisca_core/holders.py
```diff
@@ -118,10 +122,12 @@ def clone(self, entity):
         return new
 
     def compute(self, period, **parameters):
-        """Compute array if needed and return a dated holder containing it.
+        """
+            Compute the variable for ``period`` and return a dated holder containing the value.
 
-        The returned dated holder is always of the requested period and this method never returns None.
+            The returned :any:`DatedHolder` is always of the requested period and this method never returns ``None``.
```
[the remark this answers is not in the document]
fpagnoux: A while ago, when you asked to compute a variable for a given period, your response would be the "output period" and a result. The output period didn't necessarily match the requested period.

This has changed (for good!), and this comment highlight this change. However, a doc string is not a changelog and should explain how a function works, not how it has evolved 👍

openfisca_core/holders.py
fpagnoux: When the variable has "extra params". 

Extra params were introduced to be able to calculate the PPA, and the RSA from 2017. The way these benefits are calculated don't really match the OpenFisca model, as there are two periods involved: the "demand month" and the "running month". Hence the introduction of extra params. 

They for sure bring a lot of complexity for a single use case. @michelbl had ideas to remove them, but we never prioritized it.

openfisca_core/holders.py
fpagnoux: There is no order in a dictionary, so no first element either.
The alternative would be 

```py
first_key = self._array_by_period.keys()[0]
first_value = self._array_by_period[first_key]
```

which is more heavy

openfisca_core/holders.py
fpagnoux: The `_array` vs `_array_by_period` is ugly and annoying, and probably requires more than a renaming. I'll just improve the comment for this PR 🙂

openfisca_core/entities.py
```diff
@@ -229,6 +229,22 @@ def get_holder(self, variable_name):
             holder.formula = variable.formula(holder = holder)  # Instanciates a Formula
         return holder
 
+    def get_memory_usage(self, variables = None):
+        holders_memory_usage = {
+            variable_name: holder.get_memory_usage()
+            for variable_name, holder in self._holders.iteritems()
```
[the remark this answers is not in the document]
fpagnoux: I'm not sure I understand what you suggest.

To which class would be attached such a function ?

tests/core/test_reforms.py
```diff
@@ -43,6 +52,22 @@ def test_formula_neutralization():
     assert_near(disposable_income_reform, 0)
 
 
+def test_default_variable_neutralization():
```
Anna-Livia: I don't understand what is being tested here 😕 
--> I feel `test_neutralized_variable_returns_default_value_when_computed` would be more explicit :)

openfisca_core/holders.py
fpagnoux: You were right, I forgot about `self._array_by_period.values()[0]`, which may be simpler.

openfisca_core/entities.py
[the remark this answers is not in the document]
fpagnoux: Hmm, if I understand well you are suggesting to make a shortcut `self.holders_iterator` for `self._holders.iteritems()`. As `self._holders` changes during the simulation, `self.holders_iterator` would have to be a function (or a property), which excludes any gain in term of performance.

From the complexity point of view, I'm not sure introducing this helper function brings much simplicity, as `.iteritems()` is a common python method.

Readers of the class would see `holders_iterator`, and would have to do an extra step: checking what this function does. For a one line basic operation, it doesn't seem to be a good deal.

But I'm maybe not understanding well what you suggest. If so, don't hesitate to open a PR on this branch, I'll definitely have a look 🙂

openfisca_core/entities.py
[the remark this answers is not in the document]
fpagnoux: Oooh got it !

openfisca_core/holders.py
```diff
@@ -279,6 +362,9 @@ def set_input(self, period, array):
     def put_in_cache(self, value, period, extra_params = None):
         simulation = self.simulation
 
+        if value.dtype != self.variable.dtype:
+            value = value.astype(self.variable.dtype)
```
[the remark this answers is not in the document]
fpagnoux: It happens if the user tries to set an input that doesn't have the same `dtype` than the variable declared `dtype`.

The typically "innocent" case this code bit handle is:

 ```py
holder.set_input(period, np.asarray([200, 400]))
```
By default, numpy will create a 64-bits int arrays. However, in OpenFisca, we use 32-bits numbers, as it saves a lot of memory and money amounts don't need a great precision (as they are anyway rounded to cents).

But yes, you're right, we should also handle the case where the user tries to set data that doesn't match at all the expected type. In which case, we should have a clear and explicit error message 👍 .

File: openfisca_core/simulations.py [hunk 9]
```diff
@@ -244,6 +244,17 @@ def get_entity(self, entity_type = None, plural = None):
         if plural:
             return [entity for entity in self.entities.values() if entity.plural == plural][0]
 
+    def get_memory_usage(self, variables = None):
+        result = dict(
+            total_nb_bytes = 0,
+            by_variable = {}
+            )
+        for entity in self.entities.itervalues():
+            entity_memory_usage = entity.get_memory_usage(variables = variables)
```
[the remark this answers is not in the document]
fpagnoux: The algorithm could be improved performance-wise, I agree. If this function was in the "reactor core" and called each time a variable was computed, it would definitely make sense to optimize it.

As of today, this function is though used in interactive mode for investigation, or in tests scripts. The tiny performance difference would probably not bring much, so I'm not sure whether it's worth the complexity of optimization. What do you think ?

fpagnoux:
> Still need a bit of documentation on openfisca_core/periods.py - def contains :)

Github didn't mark the comment as outdated, but I promise it's there :octocat:

PR #604, opened 2017-12-12
Introduce simulation generators

fpagnoux:
Introduce a helper function to generate arbitrarily big populations. 

I used it a lot to investigate performances, and it could be reused. I put in in tests, but there may be a better spot.

This is clearly low priority

### DIFF
--- a/tests/core/simulation_generator.py
+++ b/tests/core/simulation_generator.py
@@ -1,0 +1,49 @@
+import numpy as np
+
+import random
+from openfisca_core.simulations import Simulation
+from openfisca_core.periods import period as make_period
+
+
+def make_simulation(tax_benefit_system, nb_persons, nb_groups, **kwargs):
+    """
+        Generate a simulation containing nb_persons persons spread in nb_groups groups
+    """
+    simulation = Simulation(tax_benefit_system = tax_benefit_system, **kwargs)
+    simulation.persons.ids = np.arange(nb_persons)
+    simulation.persons.count = nb_persons
+    adults = [0] + sorted(random.sample(xrange(1, nb_persons), nb_groups - 1))
+
+    members_entity_id = np.empty(nb_persons, dtype = int)
+    members_legacy_role = np.empty(nb_persons, dtype = int)
+
+    id_group = -1
+    for id_person in range(nb_persons):
+        if id_person in adults:
+            id_group += 1
+            legacy_role = 0
+        else:
+            legacy_role = 2 if legacy_role == 0 else legacy_role + 1
+        members_legacy_role[id_person] = legacy_role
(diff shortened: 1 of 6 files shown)


tests/core/simulation_generator.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+        Generate a simulation containing nb_persons persons spread in nb_groups groups
+    """
+    simulation = Simulation(tax_benefit_system = tax_benefit_system, **kwargs)
+    simulation.persons.ids = np.arange(nb_persons)
+    simulation.persons.count = nb_persons
+    adults = [0] + sorted(random.sample(xrange(1, nb_persons), nb_groups - 1))
+
+    members_entity_id = np.empty(nb_persons, dtype = int)
+    members_legacy_role = np.empty(nb_persons, dtype = int)
```
Anna-Livia: Would it be helpful to write a reminder of the `legacy_role` indexes ?

tests/core/simulation_generator.py
```diff
@@ -0,0 +1,49 @@
+import numpy as np
+
+import random
+from openfisca_core.simulations import Simulation
+from openfisca_core.periods import period as make_period
+
+
+def make_simulation(tax_benefit_system, nb_persons, nb_groups, **kwargs):
+    """
+        Generate a simulation containing nb_persons persons spread in nb_groups groups
```
Anna-Livia: Would an example (or a link to one) would be useful for fist time users ?

tests/core/simulation_generator.py
```diff
@@ -0,0 +1,49 @@
...
+            entity.members_entity_id = members_entity_id
+            entity.members_legacy_role = members_legacy_role
+            entity.count = nb_groups
+            entity.members_role = np.where(members_legacy_role == 0, entity.flattened_roles[0], entity.flattened_roles[-1])
+    return simulation
+
+
+def randomly_init_variable(simulation, variable_name, period, max_value, condition = None):
+    """
+        Initialise a variable with random values (from 0 to max_value) for the given period.
```
Anna-Livia: Would an example (or a link to one) would be useful for fist time users ?

Anna-Livia:
As for the directory, would openfisca_core/tools/script be more appropriate ?

tests/core/simulation_generator.py
fpagnoux: Added a comment to explain what legacy roles are

tests/core/simulation_generator.py
fpagnoux: Added an example

tests/core/simulation_generator.py
fpagnoux: Added an example

PR #608, opened 2018-01-05
Improve API response time

guillett:
Setting values in a complex object seems very slow using a proxy object improves performance

#### Technical changes
- Make API faster

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 21.3.3 [#608](https://github.com/openfisca/openfisca-core/pull/608)
+
+- Improve API response time
+
 ## 21.3.2 [#617](https://github.com/openfisca/openfisca-core/pull/611)
 
 - Make decompositions more robust
@@ -14,7 +18,7 @@
 
 Add `--only-variables` and `--ignore-variables` options to `openfisca-run-test` to filter out tested output variables if needed.
 
-## 21.2.2 [#612](https://github.com/openfisca/openfisca-core/pull/612)
+### 21.2.2 [#612](https://github.com/openfisca/openfisca-core/pull/612)
 
 - When a variable file is loaded twice in the same python interpreter, make sure the second loading doesn't corrupt the first one.
     - This fixes a bug introduced in 21.0.2, which could lead to a corruption of the tax and benefit in rare edge cases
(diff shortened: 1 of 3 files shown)


guillett:
Running simulation with multiple families in mes-aides.gouv.fr was very slow. I run a profiler and here are some results for 5 families.

### Before the change: 30 seconds
capture d ecran 2018-01-05 a 17 29 00

### After the change: 2 seconds
capture d ecran 2018-01-05 a 17 26 34

## Context

`dpath.util.set(input_data, path, entity_result)` was very slow on `input_data` which is a relatively big object.

Using a proxy object `results` in my commit, adding results with `dpath.util.new` and then _filling the holes_ in one call _dpath.util.merge(input_data, results)_ is much faster.


### For reference
Here is the profile for 500 families
capture d ecran 2018-01-05 a 17 33 36

guillett:
For reference again, a great non-invasive way to profile the web API.

Gunicorn configution file updated from [mes-aides-ui/openfisca/api_config.py](https://github.com/sgmap/mes-aides-ui/blob/master/openfisca/api_config.py) (then use _snakeviz_ on generated profiling files, thanks @fpagnoux).

```python
port = 2000
workers = 1
timeout = 600

country_package = 'openfisca_france'
extensions = ['openfisca_paris', 'openfisca_brestmetropole', 'openfisca_rennesmetropole']

import cProfile
import pstats
import StringIO
import logging
import os
import time

def pre_request(worker, req):
    import cProfile
    import pstats
    import StringIO
    import logging
    import os
    import time
    worker.start_time = time.time()
    worker.profile = cProfile.Profile()
    worker.profile.enable()
    worker.log.info("PROFILING %d: %s" % (worker.pid, req.uri))

def post_request(worker, req, *args):
    import cProfile
    import tempfile
    tf = tempfile.NamedTemporaryFile(delete = False)
    print("http://127.0.0.1:8080/snakeviz/" + tf.name)
    worker.profile.dump_stats(tf.name)
```

fpagnoux:
@guillett I feel like some of your changes have been eaten by the rebase 😬 .
The current diff doest not include any code removal. Maybe you want to remove [this line](https://github.com/openfisca/openfisca-core/pull/608/files#diff-42317350db214ac58ef6f760ac548574R120) ?

guillett:
Thanks @fpagnoux I tried to rebase and fix the highlighted issue but tests are now failing and I don't have time to focus on that now.

PR #610, opened 2018-01-11
Add run test options

benjello:

#### New features

- Introduce new options `--only` and '--ignore` to openfisca-run-test
  - Allows for filter tthe variables to test in YAML tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 21.3.0
+
+Add `--only-variables` and `--ignore-variables` options to `openfisca-run-test` to filter out tested output variables if needed.
+
 ### 21.2.2 [#612](https://github.com/openfisca/openfisca-core/pull/612)
 
 - When a variable file is loaded twice in the same python interpreter, make sure the second loading doesn't corrupt the first one.
--- a/openfisca_core/tools/test_runner.py
+++ b/openfisca_core/tools/test_runner.py
@@ -128,6 +128,8 @@ def _generate_tests_from_file(tax_benefit_system, path_to_file, options):
     if isinstance(name_filter, str):
         name_filter = name_filter.decode('utf-8')
     verbose = options.get('verbose')
+    only_variables = options.get('only_variables')
+    ignore_variables = options.get('ignore_variables')
 
     tests = _parse_test_file(tax_benefit_system, path_to_file)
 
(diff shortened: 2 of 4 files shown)


benjello:
Submitting this PR on the behalf of @bfabre01 which didn't have the right to push brances here and got confused on the different core branches. These are convenient options when you test for particular variables on a more complete test like the one produced by the openfisca/DGFiP comparator used for examples in this [branch](https://github.com/openfisca/openfisca-france/tree/update-and-test-irpp).

benjello:
@Anna-Livia @sandcha @fpagnoux : any chance you can have a look at this soon ?
It is simple and easy and very helpful for users like @bfabre01 and @ClaireLeroyIPP ;-)
Thanks !

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 21.2.2
+
+Add `--only` and `--ignore` options to `openfisca-run-test` to filter out tested output variables if needed.
```
fpagnoux: `--only` seems a little too general. `-n` also _only tests with the given name_filter in their name_. What about `--only-variables` and `--ignore-variables` ?

openfisca_core/tools/test_runner.py
```diff
@@ -230,26 +231,35 @@ def _run_test(period_str, test, verbose = False, options = {}):
...
     simulation = scenario.new_simulation(trace = verbose)
     output_variables = test.get(u'output_variables')
+    tested_variables = output_variables.keys()
+    if test_only is not None:
+        tested_variables = test_only
     if output_variables is not None:
         try:
+            if test_only is not None:
+                for variable in test_only:
+                    assert variable in output_variables.keys(), "La variable {} n'est pas présente dans les outputs variables du test".format(variable)
```
fpagnoux: Any reason to throw an error in that case ? It seems limitating : I cannot run all the `rsa` tests in mes-aides  for instance, as an error would be raised for files that don't test `rsa`.

benjello: @bfabre01 @ClaireLeroyIPP :  
 - should we raise an error to force the user to be sure to list the right tests to test his specific variable ?
 - should we ignore tests when the variable is not present (my favorite)  ?

benjello:
@fpagnoux : c'est bon !
@sandcha @Anna-Livia si c'est bon pour vous let's merge !

PR #617, opened 2018-02-02
Decode API /parameter and /variable

sandcha:
Fixes openfisca/openfisca-france#891

#### Bug Fix
- Fixes encoding for API `/parameter` and `/variable` endpoints

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ from setuptools import setup, find_packages
 
 setup(
     name = 'OpenFisca-Core',
-    version = '21.3.0',
+    version = '21.3.1',
     author = 'OpenFisca Team',
     author_email = 'contact@openfisca.fr',
     classifiers = [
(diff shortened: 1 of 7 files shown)


sandcha:
@Anna-Livia @fpagnoux Savez-vous pourquoi nous avons deux lignes pour `country-template` dans `setup.py` ? i.e, avant modification : 
```
'openfisca-country-template == 2.0.1',
'openfisca-country-template >= 2.0.1rc0, <= 2.0.1'
```

setup.py
```diff
@@ -32,8 +32,8 @@
         'test': [
             'nose',
             'flake8 >= 3.4.0, < 3.5.0',
-            'openfisca-country-template == 2.0.1',
-            'openfisca-country-template >= 2.0.1rc0, <= 2.0.1',
+            'openfisca-country-template == 2.1.0',
+            'openfisca-country-template >= 2.0.1rc0, <= 2.1.0',
```
Anna-Livia: Is this line still useful ?

Morendil:
@sandcha Sais-tu quelle modification (quel commit ou quelle PR) a introduit le bug ?

setup.py
sandcha: I took me some time to understand. 🙃 
Now that it's clearer: yes the line is still useful as Core depends on Country Template for its tests and Country Template gets the following error on `make test` with Core latest version `OpenFisca-Core (21.3.0)`: 
```
ERROR:openfisca_core.taxbenefitsystems:Unable to load OpenFisca variables from file "/Users/sch/Dev/github/country-template/openfisca_country_template/variables/benefits.py"
(...)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 40: ordinal not in range(128)
```
So, I'm removing the line before.

Issue #630, opened 2018-03-23
Make openfisca-core compatible with Python 3


maukoquiroga:
From 1/1/2109 on NumPy won't be supporting Python 2.x anymore, so we'll probably have to move from 2.x to 3.x

**Definition of done :**

Assessment of general complexity and efforts to do this, and best-effort enumeration of things to do.

- [x] Foray 1 #631 
- [x] Foray 2 #632 
- [x] Make package compatible with Python 3.6 #660
- [x] Make package dependencies compatible with Python 3 openfisca/country-template#41
- [x] Automatic CI/CD #662 -> #666
- [ ] Performance assessment

Thanks for your information if you already have cc @fpagnoux and @benjello 

Should be compatible with 2.7 **and** 3.6

fpagnoux:
What is the difference between Foray 1 and Foray 2 ?

I don't think moving to Python 3 will be a hard thing to do, but keeping compatibility with python 2 may be if we use python 3 specific features. 

What is the strategy about retro-compatibility ? Do we intend to keep OpenFisca compatible with Python 2 ? For how long ?

benjello:
For the IPP, I think we can prepare to do a full switch to python 3 around next january (2019).
We can test it a bit locally during the summer but switching to python 3 on the CASD is risky and costly and we won't be able to do it in the summer or the autumn.

fpagnoux:
> switching to python 3 on the CASD is risky and costly and we won't be able to do it in the summer or the autumn.

I'm suspecting that several of OpenFisca reusers will have the same issue. So I think OpenFisca will have to be compatible with Python 2 ou 3 for at least a couple of months.

To this aim, I'd suggest the following methodology:
- Try to run OpenFisca with python3
- Run into an issue
- Fix it in a way that works **both** for Python 2 and 3.
    - If there is a way of doing things that works for both version, I'd suggest using it instead of a more fancy python-3 way.
    - If we end up using new features brought by python3, polyfills (such as enum34) can be used
    - In the other case, and `if`/`else` block is always possible. 
    - In all case, the interface of OpenFisca should be the same in python 2 and 3.
- Move to the next issue 🚅 

I'd suggest _not_ to wait to have made the whole code python-3 compatible to start worrying about retro-compatibility. This strategy could lead to higher costs, as part of the work may need to be undone.

maukoquiroga:
> What is the difference between Foray 1 and Foray 2 ?

[...]

PR #639, opened 2018-04-11
Update CircleCI configuration to its v2

maukoquiroga:
#### Technical changes

- Update CircleCI configuration to its v2
- Makes it easier to continuously test #630 

---

CircleCI 1.0 will be sunsetted [August, 31 2018](https://circleci.com/blog/sunsetting-1-0/). We have to migrate all of our repos currently using the V1, and this is a good way to start.

Additionally, thanks to workflows, we'll be able to migrate to Python 3 softly.


benjello:
Suis pas compétent :-(

.circleci/config.yml
```diff
@@ -0,0 +1,121 @@
...
+    working_directory: ~/openfisca-core
+    docker:
+      - image: python:2.7.14
+
+    steps:
+      - restore_cache:
+          keys:
+            - v1-checkout-{{ .Environment.CIRCLE_SHA1 }}
+
+      - restore_cache:
```
sandcha: Isn't `v1-dependencies` set in this `dependencies` job?

.circleci/config.yml
```diff
@@ -0,0 +1,121 @@
...
+          keys:
+            - v1-dependencies-{{ .Environment.CIRCLE_SHA1 }}
+
+      - run:
+          name: Create virtualenv
```
sandcha: Why do we need to create a virtualenv to test one Core revision?

.circleci/config.yml
```diff
@@ -0,0 +1,121 @@
...
+
+      # pip >= 8.0 needed to be compatible with "manylinux" wheels, used by numpy >= 1.11
+      - run:
+          name: Install dependencies
+          command: |
+            . /tmp/venv/openfisca-core/bin/activate
+            pip install --upgrade pip twine wheel
+            pip install .[test] --upgrade
+
+      # Uncomment and adapt the next line to use a particular feature branch of OpenFisca-Country-Template to run Circle CI tests
```
sandcha: 👍

.circleci/config.yml
```diff
@@ -0,0 +1,121 @@
...
+      # - run:
+      #     name: Install OpenFisca-Country-Template
```
sandcha: > Isn't it an *update* of a `test` dependency?

.circleci/config.yml
```diff
@@ -0,0 +1,121 @@
...
+            - v1-checkout-{{ .Environment.CIRCLE_SHA1 }}
+
+      - restore_cache:
+          keys:
+            - v1-dependencies-{{ .Environment.CIRCLE_SHA1 }}
+
+      - run:
+          name: Deploy (if version bump)
+          command: |
+            git fetch
```
sandcha: Isn't it a bit late to `git fetch`? Could this be done on `checkout` for example?

.circleci/config.yml
sandcha: Circle warning: `No cache is found for key xxxxxx`

.circleci/config.yml
maukoquiroga: Because we're passing dependencies from job to job: each job corresponds to an empty Docker image.

.circleci/config.yml
maukoquiroga: > Isn't v1-dependencies set in this dependencies job?

Yes, the idea is that, it is cached the first time (no cache found), and it is restored subsequently.

Now, given cache is saved by commit, it won't ever be restored here, so I'll delete the step.

.circleci/config.yml
maukoquiroga: I'll add:

> Install a particular feature branch of OpenFisca-Country-Template

.circleci/config.yml
```diff
@@ -0,0 +1,120 @@
+version: 2
+
+jobs:
+  checkout:
+    working_directory: ~/openfisca-core
+    docker:
+      - image: python:2.7.14
+
+    steps:
+      - checkout
```
sandcha: > Could be `fetch` then.

fpagnoux:
@sandcha @benjello  For information, a version bump and a changelog edition were not necessary in this case, as you were not editing the code of the package.

The version update enforcing is not that strict ;)

maukoquiroga:
Hi @fpagnoux, thanks for the tip, and I've just checked the [script](https://github.com/openfisca/openfisca-core/blob/master/check-version-bump.sh#L5).

PR #644, opened 2018-04-13
Improve parameters performances

fpagnoux:
On `france`:
-  `nosetests tests/test_basics.py` drops from `35s` to `25s`
-  [This file](https://github.com/openfisca/tutorial/blob/master/python/scripts/generate_situation_examples/try_situation_examples.py) drops from `5.26s` to `3.55s`

Interesting lessons:
- Performances waste does not necessarily happen where you think it does
- Flexibles user-friendly interfaces can ruin performances if called under the hood

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 22.0.6 [#642](https://github.com/openfisca/openfisca-core/pull/642)
+
+* Improve parameters performances
+
 ### 22.0.5 [#639](https://github.com/openfisca/openfisca-core/pull/639)
 
 * Update country-template dependency to >= 3.0.0, < 4.0.0
--- a/openfisca_core/parameters.py
+++ b/openfisca_core/parameters.py
@@ -159,6 +159,9 @@ class Parameter(object):
 
     def get_at_instant(self, instant):
         instant = str(periods.instant(instant))
+        return self._get_at_instant(instant)
+
+    def _get_at_instant(self, instant):
         for value_at_instant in self.values_list:
             if value_at_instant.instant_str <= instant:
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+### 22.0.6 [#642](https://github.com/openfisca/openfisca-core/pull/642)
+
+* Improve parameters performances
```
maukoquiroga: Parameters' performance?

fpagnoux:
> How did you discover that `'amount'` in bracket is faster then `hasattr(bracket, 'amount')`

By profiling and discovering that weirdly a lot of time was spent creating `ParameterNotFound` errors, which didn't think to make sense. 

I turned out that each time, `hasattr` calls `__getattr__`, which raise a `ParameterNotFound` if the attributes doesn't exist, and catches it under the hood (as it's a subclass of AttributeError). That's more work than just checking if the item is in a list. 

I also found [resources](https://hynek.me/articles/hasattr/) indicating that `hasattr` is generally not a good option.

fpagnoux:
For the record:
- I tried to optimize [this function](https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/parameters.py#L165-L168) using [binary search](https://en.wikipedia.org/wiki/Binary_search_algorithm), but it was counter productive. Probably because parameters usually don't have a lot of historical values, or because we tend to look for the most recent ones.

- I think there is a time to be saved by lazy evaluating parameters: each time someone needs a parameter for a specific date, the whole tree is evaluated for this time, which seems overkill.

guillett:
Many, many thanks @fpagnoux !

PR #645, opened 2018-04-13
Make OpenFisca more compatible with Python3

### LINKED ISSUE #631: Python 3 Foray 1
Related to #630

**Definition of done**

- [x] Make 1 test pass in Python 2.7 and 3.x
- [x] Document learning
- [ ] Merge changes upstream (1 test passing)

### LINKED ISSUE #632: Python 3 Foray 2
Related to #630

Anna-Livia:
Fixes #631
Fixes #632 

#### Technical changes
Python3 has deprecated a lot of elements present in python2 and in OpenFisca : 

- Regex ur"string" --> Regex r"string"
- No more Implicit imports --> explicit imports
- No more `basestring` or `unicode` types (all strings (`str`) are unicode) --> `text_unicode = u"".__class__` and `text_basestring = (str, text_unicode)` in type checking.
- `cmp` is an attribute of `sort()` --> Change `cmp` to `key` 
- `dict.iteritems()` --> `dict.items()`
- `asser_items_equal' --> equivalent assert equal expressions

To work with Python 3, it needs version 0.10.6dev of `biryani`
nosetests core/test_base_functions.py passes :tada:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+### 23.0.2 [#645](https://github.com/openfisca/openfisca-core/pull/645)
+
+* Technical improvement
+* Details:
+
+Start adapting OpenFisca to Python 3
+- Imports are now all absolute imports
+- `unicode_type` and `basestring_type` are now used for compatible type checking.
+- All calls for sorted() use the key parameter
+- Iteration on dict now uses dict.items()
+
 ### 23.0.1 [656](https://github.com/openfisca/openfisca-core/pull/656)
 
 * Re-accept `int` values for `instant` in `tax_benefit_system.get_parameter_at_instant(instant)`
--- a/openfisca_core/base_functions.py
+++ b/openfisca_core/base_functions.py
@@ -27,7 +27,7 @@ def requested_period_last_value(holder, period, *extra_params, **kwargs):
     known_periods = holder.get_known_periods()
     if not known_periods:
         return holder.default_array()
-    known_periods = sorted(known_periods, cmp = periods.compare_period_start, reverse = True)
+    known_periods = sorted(known_periods, key=lambda period: period.start, reverse = True)
     for last_period in known_periods:
         if last_period.start <= period.start:
             return holder.get_array(last_period, extra_params)
--- a/openfisca_core/base_functions.py
+++ b/openfisca_core/base_functions.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 
-from . import periods
 
 """
     base_function is an optional variable attribute that can optionally be set to one of the functions defined in this module.
--- a/openfisca_core/columns.py
+++ b/openfisca_core/columns.py
@@ -8,7 +8,7 @@ import numpy as np
 
 from openfisca_core import conv, periods
 from openfisca_core.indexed_enums import Enum
-from openfisca_core.commons import unicode_type, basestring_type, unicode_this
+from openfisca_core.commons import unicode_type, basestring_type, to_unicode
 
 """
 Columns are the ancestors of Variables, and are now considered deprecated. Preferably use `Variable` instead.
@@ -180,7 +180,7 @@ class DateCol(Column):
 
     def json_default(self):
         default = np.array(self.default_value, self.dtype)
-        return unicode_this(default)
+        return to_unicode(default)
 
     @property
     def json_to_dated_python(self):
@@ -240,7 +240,7 @@ class FloatCol(Column):
     @property
     def json_to_dated_python(self):
         return conv.pipe(
-            conv.test_isinstance((float, int, str, basestring_type)),
+            conv.test_isinstance((float, int, basestring_type)),
             conv.make_anything_to_float(accept_expression = True),
             )
 
@@ -256,7 +256,7 @@ class IntCol(Column):
     @property
     def json_to_dated_python(self):
         return conv.pipe(
-            conv.test_isinstance((int, str, basestring_type)),
+            conv.test_isinstance((int, basestring_type)),
             conv.make_anything_to_int(accept_expression = True),
             )
 
@@ -329,8 +329,8 @@ class EnumCol(Column):
 
     def json_default(self):
         default = self.default_value
-        if default is not None and not isinstance(default, unicode_type):
-            return unicode(default)
+        if default is not None:
+            to_unicode(default)
         return default
 
     @property
--- a/openfisca_core/columns.py
+++ b/openfisca_core/columns.py
@@ -6,8 +6,9 @@ import re
 
 import numpy as np
 
-from indexed_enums import Enum
-from . import conv, periods
+from openfisca_core import conv, periods
+from openfisca_core.indexed_enums import Enum
+from openfisca_core.commons import unicode_type, basestring_type, unicode_this
 
 """
 Columns are the ancestors of Variables, and are now considered deprecated. Preferably use `Variable` instead.
--- a/openfisca_core/columns.py
+++ b/openfisca_core/columns.py
@@ -180,9 +180,7 @@ class DateCol(Column):
 
     def json_default(self):
         default = np.array(self.default_value, self.dtype)
-        if isinstance(default, unicode_type):
-            return default
-        return unicode(default)
+        return unicode_this(default)
 
     @property
     def json_to_dated_python(self):
--- a/openfisca_core/columns.py
+++ b/openfisca_core/columns.py
@@ -159,7 +159,7 @@ class BoolCol(Column):
     @property
     def json_to_dated_python(self):
         return conv.pipe(
-            conv.test_isinstance((basestring, bool, int)),
+            conv.test_isinstance((basestring_type, bool, int)),
             conv.guess_bool,
             )
 
@@ -179,7 +179,10 @@ class DateCol(Column):
             )
 
     def json_default(self):
-        return unicode(np.array(self.default_value, self.dtype))
+        default = np.array(self.default_value, self.dtype)
+        if isinstance(default, unicode_type):
+            return default
+        return unicode(default)
 
     @property
     def json_to_dated_python(self):
@@ -194,7 +197,7 @@ class DateCol(Column):
                         conv.function(lambda year: datetime.date(year, 1, 1)),
                         ),
                     conv.pipe(
-                        conv.test_isinstance(basestring),
+                        conv.test_isinstance(basestring_type),
                         conv.test(year_or_month_or_day_re.match, error = N_(u'Invalid date')),
                         conv.function(lambda birth: u'-'.join((birth.split(u'-') + [u'01', u'01'])[:3])),
                         conv.iso8601_input_to_date,
@@ -221,9 +224,9 @@ class FixedStrCol(Column):
             conv.condition(
                 conv.test_isinstance((float, int)),
                 # YAML stores strings containing only digits as numbers.
-                conv.function(unicode),
+                conv.function(unicode_type),
                 ),
-            conv.test_isinstance(basestring),
+            conv.test_isinstance(basestring_type),
             conv.test(lambda value: len(value) <= self.variable.max_length),
             )
 
@@ -239,7 +242,7 @@ class FloatCol(Column):
     @property
     def json_to_dated_python(self):
         return conv.pipe(
-            conv.test_isinstance((float, int, basestring)),
+            conv.test_isinstance((float, int, str, basestring_type)),
             conv.make_anything_to_float(accept_expression = True),
             )
 
@@ -255,7 +258,7 @@ class IntCol(Column):
     @property
     def json_to_dated_python(self):
         return conv.pipe(
-            conv.test_isinstance((int, basestring)),
+            conv.test_isinstance((int, str, basestring_type)),
             conv.make_anything_to_int(accept_expression = True),
             )
 
@@ -272,9 +275,9 @@ class StrCol(Column):
             conv.condition(
                 conv.test_isinstance((float, int)),
                 # YAML stores strings containing only digits as numbers.
-                conv.function(unicode),
+                conv.function(unicode_type),
                 ),
-            conv.test_isinstance(basestring),
+            conv.test_isinstance(basestring_type),
             )
 
 
@@ -320,14 +323,17 @@ class EnumCol(Column):
     def input_to_dated_python(self):
         enum = self.variable.possible_values
         if enum is None:
-            return conv.test_isinstance((basestring, basestring))
+            return conv.test_isinstance(basestring_type)
         return conv.pipe(
             # Verify that item index belongs to enumeration.
             conv.test_in([item.name for item in list(enum)])
             )
 
     def json_default(self):
-        return unicode(self.default_value) if self.default_value is not None else None
+        default = self.default_value
+        if default is not None and not isinstance(default, unicode_type):
+            return unicode(default)
+        return default
 
     @property
     def json_to_dated_python(self):
@@ -336,10 +342,10 @@ class EnumCol(Column):
 
         if enum is None:
             return conv.pipe(
-                conv.test_isinstance((basestring, basestring))
+                conv.test_isinstance(basestring_type)
                 )
         return conv.pipe(
-            conv.test_isinstance((basestring, basestring)),
+            conv.test_isinstance(basestring_type),
             conv.pipe(
                 # Verify that item belongs to enumeration.
                 conv.test_in(possible_names),
--- a/openfisca_core/columns.py
+++ b/openfisca_core/columns.py
@@ -20,7 +20,7 @@ def N_(message):
     return message
 
 
-year_or_month_or_day_re = re.compile(ur'(18|19|20)\d{2}(-(0?[1-9]|1[0-2])(-([0-2]?\d|3[0-1]))?)?$')
+year_or_month_or_day_re = re.compile(r'(18|19|20)\d{2}(-(0?[1-9]|1[0-2])(-([0-2]?\d|3[0-1]))?)?$')
 
 
 # Base Column
--- a/openfisca_core/commons.py
+++ b/openfisca_core/commons.py
@@ -5,23 +5,20 @@ unicode_type = u"".__class__
 basestring_type = (str, unicode_type)
 
 
-def unicode_this(string, encoding=None):
+def to_unicode(string):
     """
     :param string: a string that needs to be unicoded
     :param encoding: a string that represent the encoding type
     :return: a unicode string
     if the string is a python 2 str type, returns a unicode version of the string.
     """
+    if not isinstance(string, basestring_type):
+        string = str(string)
     if isinstance(string, unicode_type):
         return string
-    # These lines only gets triggered if the code is run in python 2
-    else:
-        if not encoding:
-            return unicode(string)
-        elif encoding == 'utf-8':
-            return unicode(string, 'utf-8')
-        else:
-            raise NameError('the encoding type {} is not handled'.format(encoding))
+
+    # Next line only gets triggered if the code is run in python 2
+    return unicode(string, 'utf-8')
 
 
 class Dummy(object):
--- a/openfisca_core/commons.py
+++ b/openfisca_core/commons.py
@@ -1,7 +1,27 @@
 # -*- coding: utf-8 -*-
 
+# The following two variables and the is_unicode function are there to bridge string types across Python 2 & 3
 unicode_type = u"".__class__
 basestring_type = (str, unicode_type)
+
+
+def unicode_this(string, encoding=None):
+    """
+    :param string: a string that needs to be unicoded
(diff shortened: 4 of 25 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 # Changelog
 
+### 22.0.5 [#645](https://github.com/openfisca/openfisca-core/pull/645)
+
+* Technical improvement:
+* Details:
+
+    Python3 has deprecated a lot of elements present in python2 and in OpenFisca :
```
maukoquiroga: > OpenFisca :

OpenFisca:

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 # Changelog
 
+### 22.0.5 [#645](https://github.com/openfisca/openfisca-core/pull/645)
+
+* Technical improvement:
```
maukoquiroga: > Technical improvement:

Technical improvement

fpagnoux:
> it needs version 0.10.6dev of biryani

I didn't find this version on [pypi](https://pypi.org/project/Biryani/0.10.5.dev0/). Could we publish it, preferably under a [regular version number](https://github.com/etalab/biryani/pull/3), like `0.10.7` ?

openfisca_core/conv.py
```diff
@@ -9,6 +9,8 @@
 from biryani.jsonconv import *  # noqa
 from biryani.states import State
 
+text_unicode = u"".__class__
+text_basestring = (str, text_unicode)
```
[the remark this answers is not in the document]
fpagnoux: I'm not sure this biryani-related module is the best place to define this types that are going to be imported from everywhere. What about `commons.py` ?

openfisca_core/periods.py
```diff
@@ -711,7 +717,7 @@ def instant(instant):
         return None
     if isinstance(instant, Instant):
         return instant
-    if isinstance(instant, basestring):
+    if isinstance(instant, (str, text_unicode)):
```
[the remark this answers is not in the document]
Anna-Livia: Excellent question :) 
I think it is because I did a few `(str, text_unicode)` before creating `conv.text_basestring` and did not replace them all :sweat_smile:

openfisca_core/simulations.py
```diff
@@ -277,16 +277,15 @@ def get_memory_usage(self, variables = None):
...
 
-def check_type(input, type, path = []):
+def check_type(input, input_type, path = []):
     json_type_map = {
-        dict: "Object",
-        list: "Array",
-        basestring: "String"
+        "<type 'dict'>": "Object",
+        "<type 'list'>": "Array",
+        "(<type 'str'>, <type 'unicode'>)": "string"
```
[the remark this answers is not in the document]
Anna-Livia: this :

Anna-Livia:
I still need to publish byriani and have python3 in the tests

openfisca_core/simulations.py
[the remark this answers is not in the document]
fpagnoux: I meant replacing `basestring` by `text_basestring` from the initial version of the code:

```
def check_type(input, input_type, path = []):
    json_type_map = {
        dict : "Object",
        list: "Array",
        text_basestring: "string",
        }
    if not isinstance(input, input_type):
        raise SituationParsingError(path,
            u"Invalid type: must be of type '{}'.".format(json_type_map[input_type]))
```

openfisca_core/periods.py
```diff
@@ -327,7 +328,10 @@ def __str__(self):
         if (unit == MONTH and size == 12 or unit == YEAR and size == 1):
             if month == 1:
                 # civil year starting from january
-                return unicode(year)
+                if isinstance(year, unicode_type):
```
sandcha: > This pattern (if unicode, return else, unicode()) is used multiple times. Define a `to_unicode(...)` in `common`?

openfisca_core/periods.py
```diff
@@ -841,27 +847,16 @@ def raise_error(value):
     return Period((unit, base_period.start, size))
 
 
-def compare_period_size(a, b):
-    unit_a, start_a, size_a = a
-    unit_b, start_b, size_b = b
+def key_period_size(period):
```
sandcha: Add a comment to explain the role of this function and its return format? In this context, the reader might not understand what `key` refers to.

sandcha:
@Anna-Livia No new blocking point (mainly comments to add) but I agree with @fpagnoux on the risk of python 2 syntax reintroduction and relative imports hence comments rather than a changes request/approve.

openfisca_core/columns.py
```diff
@@ -6,8 +6,9 @@
 
 import numpy as np
 
-from indexed_enums import Enum
-from . import conv, periods
+from openfisca_core.indexed_enums import Enum
+from openfisca_core import conv, periods
+from .commons import unicode_type, basestring_type
```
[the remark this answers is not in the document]
fpagnoux: `from commons` would be an implicit relative import, which doesn't work in Python 3.

openfisca_core/columns.py
```diff
@@ -167,7 +168,7 @@ def input_to_dated_python(self):
     @property
     def json_to_dated_python(self):
         return conv.pipe(
-            conv.test_isinstance((basestring, bool, int)),
+            conv.test_isinstance((str, unicode_type, bool, int)),
```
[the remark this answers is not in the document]
fpagnoux: `isinstance('x', ((str, unicode), int, bool))` works, so I'd assume `conv.test_isinstance((basestring_type, bool, int))` would work too.

openfisca_core/columns.py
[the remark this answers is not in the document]
fpagnoux: Though if we want to only use absolute imports, we should be consistent :)

openfisca_core/base_functions.py
```diff
@@ -2,7 +2,7 @@
 
 import warnings
 
-from . import periods
```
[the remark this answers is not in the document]
fpagnoux: Ok to use only absolute import, as it seems to be considered the best practice in Python.

openfisca_core/data_storage.py
```diff
@@ -5,9 +5,9 @@
 
 import numpy as np
 
-import periods
-from periods import ETERNITY
-from indexed_enums import EnumArray
+import openfisca_core.periods as periods
```
fpagnoux: `from openfisca_core import periods`

openfisca_core/holders.py
```diff
@@ -9,12 +9,12 @@
...
 import psutil
 
-from commons import empty_clone
-import periods
-from periods import MONTH, YEAR, ETERNITY
-from columns import make_column_from_variable
-from indexed_enums import Enum, EnumArray
-from data_storage import InMemoryStorage, OnDiskStorage
+from openfisca_core.commons import empty_clone
+import openfisca_core.periods as periods
```
fpagnoux: `from openfisca_core import periods`

openfisca_core/simulations.py
```diff
@@ -8,9 +8,9 @@
 
 import dpath
 
-import periods
-from commons import empty_clone
-from tracers import Tracer
+import openfisca_core.periods as periods
```
fpagnoux: `from openfisca_core import periods`

openfisca_core/taxbenefitsystems.py
```diff
@@ -12,12 +12,13 @@
 import traceback
 from setuptools import find_packages
 
-import conv
-from parameters import ParameterNode
-from variables import Variable
-from scenarios import AbstractScenario
-from formulas import get_neutralized_variable
-from errors import VariableNotFound
+import openfisca_core.conv as conv
```
fpagnoux: `from openfisca_core import conv`

openfisca_core/columns.py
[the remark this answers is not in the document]
fpagnoux: Though for consistency, we may want to stick to absolute imports.

openfisca_core/base_functions.py
```diff
@@ -51,7 +49,7 @@ def requested_period_last_value(formula, simulation, period, *extra_params, **kw
     known_periods = holder.get_known_periods()
     if not known_periods:
         return holder.default_array()
-    known_periods = sorted(known_periods, cmp = periods.compare_period_start, reverse = True)
+    known_periods = sorted(known_periods, key=lambda period: period.start, reverse = True)
```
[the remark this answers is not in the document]
Anna-Livia: Calling the attribute directly seems to be the pythonic way to do it
https://wiki.python.org/moin/HowTo/Sorting#Sortingbykeys
```
>>> sorted(student_objects, key=lambda student: student.age)   # sort by age
[('dave', 'B', 10), ('jane', 'B', 12), ('john', 'A', 15)]
```
it there an added benefit to 
`periods = sorted(known_periods, key=lambda period: getattr(period, 'start'), reverse = True)`
that I'm not seeing ?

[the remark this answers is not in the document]
fpagnoux: I think @sandcha meant `known_periods = sorted(known_periods, key=operator.attrgetter('start'), reverse = True)`. See [doc](https://docs.python.org/3/library/operator.html#operator.attrgetter).

I find lambdas clearer, because more similar to other languages, but it seems to be [debated](https://utcc.utoronto.ca/~cks/space/blog/python/AttrgetterVsLamba)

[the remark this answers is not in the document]
Anna-Livia: I'm keeping the lambda for now, however, let's test the performance of `operator.attrgetter('start')`vs object.start to see if we could make Openfisca more efficient .

openfisca_core/periods.py
Anna-Livia: @fpagnoux In what case in size > 1? in set_inputs ?

fpagnoux: Yes, for instance you could set a `salary` for 3 years (2012 to 2014), then the unit would be YEAR and the size 3.  
This is indeed only possible if a `set_input` has been defined for the variable.

Anna-Livia:
@fpagnoux  are there still some blocking issues for you ?

openfisca_core/simulations.py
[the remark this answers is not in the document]
fpagnoux: @Anna-Livia this is still blocking for me

openfisca_core/columns.py
```diff
@@ -238,7 +240,7 @@ def input_to_dated_python(self):
     @property
     def json_to_dated_python(self):
         return conv.pipe(
-            conv.test_isinstance((float, int, basestring)),
+            conv.test_isinstance((float, int, str, basestring_type)),
```
fpagnoux: Isn't `str` included in `basestring_type ` ?

openfisca_core/columns.py
```diff
@@ -254,7 +256,7 @@ def input_to_dated_python(self):
     @property
     def json_to_dated_python(self):
         return conv.pipe(
-            conv.test_isinstance((int, basestring)),
+            conv.test_isinstance((int, str, basestring_type)),
```
fpagnoux: Isn't `str` included in `basestring_type ` ?

openfisca_core/columns.py
```diff
@@ -319,14 +321,17 @@ class EnumCol(Column):
...
+            return conv.test_isinstance(basestring_type)
         return conv.pipe(
             # Verify that item index belongs to enumeration.
             conv.test_in([item.name for item in list(enum)])
             )
 
     def json_default(self):
-        return unicode(self.default_value) if self.default_value is not None else None
+        default = self.default_value
+        if default is not None and not isinstance(default, unicode_type):
```
fpagnoux: > Any reason not to used `unicode_this` ?

openfisca_core/commons.py
```diff
@@ -1,5 +1,28 @@
 # -*- coding: utf-8 -*-
 
+# The following two variables and the is_unicode function are there to bridge string types across Python 2 & 3
+unicode_type = u"".__class__
+basestring_type = (str, unicode_type)
+
+
+def unicode_this(string, encoding=None):
```
fpagnoux: What about `to_unicode` ? The word `this` has a strong object-oriented connotation.

fpagnoux: Is this second parameter really necessary ? 
Why not encoding everything as `utf-8` ? I think it returns the same things for strings without special characters:

`unicode('2018')` and `unicode('2018', 'utf-8')` seem to be equal

openfisca_core/commons.py
```diff
@@ -1,5 +1,28 @@
...
+    if isinstance(string, unicode_type):
+        return string
+    # These lines only gets triggered if the code is run in python 2
+    else:
+        if not encoding:
+            return unicode(string)
+        elif encoding == 'utf-8':
+            return unicode(string, 'utf-8')
+        else:
+            raise NameError('the encoding type {} is not handled'.format(encoding))
```
fpagnoux: `NameError` is usually when you try to use a variable that has not been defined.
For an unexpected function parameter, `ValueError` is more appropriate.

openfisca_core/decompositionsxml.py
```diff
@@ -87,10 +88,10 @@ def validate_node_xml_json(node, state = None):
                                 ),
                             ),
                         conv.test(lambda colors: len(colors) == 3, error = N_(u'Wrong number of colors in triplet.')),
-                        conv.function(lambda colors: u','.join(unicode(color) for color in colors)),
+                        conv.function(lambda colors: u','.join(color if isinstance(color, unicode_type) else unicode(color) for color in colors)),
```
fpagnoux: Could we use the helper here ?

openfisca_core/json_to_test_case.py
```diff
@@ -1,27 +1,28 @@
...
+from openfisca_core.columns import make_column_from_variable
+from openfisca_core import conv
+from openfisca_core.commons import unicode_type
 
 
 def check_entity_fields(entity_json, entity_class, valid_roles, tax_benefit_system):
 
     def check_id(value):
-        if value is None or not isinstance(value, (basestring, int)):
+        if value is None or not isinstance(value, (str, unicode_type, int)):
```
fpagnoux: `basestring_type` ?

openfisca_core/json_to_test_case.py
```diff
@@ -1,27 +1,28 @@
...
             raise ValueError(u"Invalid id in entity {}".format(entity_json).encode('utf-8'))
 
     def check_role(value, key):
         role = valid_roles.get(key)
         if role.max == 1:
-            value, error = conv.test_isinstance((basestring, int))(value)
+            value, error = conv.test_isinstance((str, unicode_type, int))(value)
```
fpagnoux: `basestring_type` ?

openfisca_core/json_to_test_case.py
```diff
@@ -1,27 +1,28 @@
...
         else:
             value, error = conv.pipe(
                 conv.make_item_to_singleton(),
                 conv.test_isinstance(list),
                 conv.uniform_sequence(
-                    conv.test_isinstance((basestring, int)),
+                    conv.test_isinstance((str, unicode_type, int)),
```
fpagnoux: `basestring_type` ?

openfisca_core/simulations.py
```diff
@@ -456,16 +456,15 @@ def clone(self, debug = False, trace = False):
         return new
 
 
-def check_type(input, type, path = []):
+def check_type(input, input_type, path = []):
     json_type_map = {
-        dict: "Object",
-        list: "Array",
-        basestring: "String"
+        "<type 'dict'>": "Object",
```
fpagnoux: See [comment marked as outdated.](https://github.com/openfisca/openfisca-core/pull/645#discussion_r183847210)

openfisca_core/taxbenefitsystems.py
```diff
@@ -96,7 +98,10 @@ def prefill_cache(self):
         pass
 
     def load_variable(self, variable_class, update = False):
-        name = unicode(variable_class.__name__)
+        name = variable_class.__name__
+        # This block is only for Python 2
+        if not isinstance(name, unicode_type):
```
fpagnoux: Could we use the helper ?

openfisca_core/variables.py
```diff
@@ -143,9 +144,12 @@ class Variable(object):
     """
 
     def __init__(self, baseline_variable = None):
-        self.name = unicode(self.__class__.__name__)
+        self.name = self.__class__.__name__
+        # This block is only for Python 2
+        if not isinstance(self.name, unicode_type):
```
fpagnoux: Could we use the helper here ?

openfisca_core/variables.py
```diff
@@ -163,11 +167,11 @@ def __init__(self, baseline_variable = None):
...
         self.entity = self.set(attr, 'entity', required = True, setter = self.set_entity)
         self.definition_period = self.set(attr, 'definition_period', required = True, allowed_values = (MONTH, YEAR, ETERNITY))
-        self.label = self.set(attr, 'label', allowed_type = basestring, setter = self.set_label)
-        self.end = self.set(attr, 'end', allowed_type = basestring, setter = self.set_end)
+        self.label = self.set(attr, 'label', allowed_type = basestring_type, setter = self.set_label)
+        self.end = self.set(attr, 'end', allowed_type = basestring_type, setter = self.set_end)
         self.reference = self.set(attr, 'reference', setter = self.set_reference)
-        self.cerfa_field = self.set(attr, 'cerfa_field', allowed_type = (basestring, dict))
-        self.unit = self.set(attr, 'unit', allowed_type = basestring)
+        self.cerfa_field = self.set(attr, 'cerfa_field', allowed_type = (str, unicode_type, dict))
```
fpagnoux: `basestring_type `?

fpagnoux:
> @fpagnoux  are there still some blocking issues for you ?

If you were waiting for me to check the code, don't hesitate asking for a review, otherwise I don't know if you're still working on it.

openfisca_core/simulations.py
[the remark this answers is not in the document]
Anna-Livia: @fpagnoux fixed :)

maukoquiroga:
(BTW your last commit makes flake8 fail)

maukoquiroga:
Hi @fpagnoux, thanks for your feedback on this issue. Please find my comments below.

> Would that mean than a build would use 2 of our 4 containers ?

The way we define parallelism in v2 has changed. From https://circleci.com/gh/openfisca/openfisca-core/edit#parallel-builds _these settings are only for 1.0 builds. To define parallelism for 2.0 jobs [read our documentation](https://circleci.com/docs/2.0/configuration-reference/#jobs)._

Parallelism is defined now is at the *job* level. What I'm advocating for is for a [_Fan-Out/Fan-In_](https://circleci.com/docs/2.0/workflows/#fan-outfan-in-workflow-example) *workflow* strategy, not to parallelise jobs.

> Also, unless there is a significant issue with the current state of the code I didn't see, I'm really dubious about the relevance of requesting changes 5 days after a PR has been accepted, based on what seems to be a minor issue.

I did my best effort to give a review based on what I thought was the expected behaviour @Anna-Livia was looking for, and based on my current knowledge of the v2.

Is the very same reason why I also included an example https://github.com/openfisca/openfisca-core/compare/python3...python3-circleci?expand=1

Also, see that at the time of my review [tests were failing](https://github.com/openfisca/openfisca-core/pull/645#issuecomment-387738793).

Additionally, by the way cache is being used in the current implementation, they'll likely be errors as we're mixing up caches from different containers.

Finally, note that the current configuration hides a misconfiguration dependency: deployment should depend on both v2.7 *and* v3.6, and it is not, but this is obfuscated by the fact of the jobs running sequentially.

> Open to discuss running Py2 and Py3 tests in parallel, but this does not seem to be a good reason to not merge this PR that has been opened for a month now.

Pull Requests that rotten are a far more serious issue than this, so there's no problem.

For the next time, though, I'd suggest that :

- it be the OP that reacts to the review, dismissing it if found out of context or disproportionate
- if we're going to bypass a review, we verify all information to be accurate *before*

Bypassing a review is not light matter, and should not become a regular practice. It *must* be solid ground.

fpagnoux:
Digging more into Cirlcle V2, I understood what you meant, and yes, the Fan-Out/Fan-In seems the most appropriate for us.

Sorry if dismissing the review seemed a little violent. My ground for doing so was:

- I didn't get enough elements in your brief review indicating that there was a serious issue with the current approved implementation. Indeed, there was an issue with a test that appeared in a fixup from @Anna-Livia (if I recall correctly). But I was able to fix this test with a minor code edit, not justifying another review cycle, especially as you and Anna-Livia were off.
- I asked the requested change to be lifted, and waited two days. You were off (and that's fine), but this added a possibly long waiting time before merging this PR.  


>it be the OP that reacts to the review, dismissing it if found out of context or disproportionate

What is the OP? 

> Bypassing a review is not light matter, and should not become a regular practice. It *must* be solid ground.

Agreed. But I'd also add that requesting changes it not light matter, and need to be given the appropriate time and energy for it to be productive.

MattiSG:
> What is the OP?

Original Post (author).

PR #612, opened 2018-01-12
Insulate TBS

fpagnoux:
- When a variable file is loaded twice in the same python interpreter, make sure the second loading doesn't corrupt the first one.
   - This fixes a bug introduced in 21.0.2, which could lead to a corruption of the tax and benefit in rare edge cases

PR #648, opened 2018-04-16
Avoid collisions between module names

### LINKED ISSUE #642: Can't load system if two files have the same name
Hi there!

### Here is what I did:
Last week, I was in France countryside and I pair-programmed to code a local benefit.

I did not manage to show our product with all the extensions activated because I had an exception when loading the French tax benefit system.

To demo the issue, I created [guillett/test_openfisca_extensions](https://github.com/guillett/test_openfisca_extensions).

I managed to investigate the issue today and I think I found the culprit line.
#### [openfisca_core/taxbenefitsystems.py:152](https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/taxbenefitsystems.py#L152)
```python
    def add_variables_from_file(self, file_path):
        """
        Adds all OpenFisca variables contained in a given file to the tax and benefit system.
        """
        try:
            file_name = path.splitext(path.basename(file_path))[0]
            module_name = '{}_{}'.format(id(self), file_name)  # If two tax and benefit systems load the same module, the second one should not replace the first one. Hence this unique module name.
            module_directory = path.dirname(file_path)
            try:
                module = load_module(module_name, *find_module(file_name, [module_directory]))
            except NameError as e:
                logging.error(str(e) + ": if this code used to work, this error might be due to a major change in OpenFisca-Core. Checkout the changelog to learn more: <https://github.com/openfisca/openfisca-core/blob/master/CHANGELOG.md>")
                raise
            potential_variables = [getattr(module, item) for item in dir(module) if not item.startswith('__')]
            for pot_variable in potential_variables:
                # We only want to get the module classes defined in this module (not imported)
                if isclass(pot_variable) and \
                        issubclass(pot_variable, Variable) and \
                        pot_variable.__module__.endswith(module_name): #   <----------- THIS LINE HERE
                    self.add_variable(pot_variable)
        except:
            log.error(u'Unable to load OpenFisca variables from file "{}"'.format(file_path))
raise
```

## `pot_variable.__module__.endswith(module_name)`

If two modules have the same names (ie. two files) that will fail.

I managed to [replicate this issue](https://circleci.com/gh/openfisca/country-template/239) with less complexity (ie. only with the country-template package).

## Context

I identify more as a:

fpagnoux:
Fixes #642 

* Allow two variable file to have the same name in a tax and benefit system
  - A [collision](https://github.com/openfisca/openfisca-core/issues/642) between module names made it impossible so far.

I'm assuming #646 will be merged first, as more urgent, and I bumped the version number accordingly.

### DIFF
--- a/openfisca_core/taxbenefitsystems.py
+++ b/openfisca_core/taxbenefitsystems.py
@@ -137,7 +137,7 @@ class TaxBenefitSystem(object):
         """
         try:
             file_name = path.splitext(path.basename(file_path))[0]
-            module_name = '{}_{}_{}'.format(id(self), hash(file_path), file_name)  # If two tax and benefit systems load the same module, the second one should not replace the first one. Hence this unique module name.
+            module_name = '{}_{}_{}'.format(id(self), hash(path.abspath(file_path)), file_name)  # If two tax and benefit systems load the same module, the second one should not replace the first one. Hence this unique module name.
             module_directory = path.dirname(file_path)
             try:
                 module = load_module(module_name, *find_module(file_name, [module_directory]))
--- a/openfisca_core/taxbenefitsystems.py
+++ b/openfisca_core/taxbenefitsystems.py
@@ -137,7 +137,12 @@ class TaxBenefitSystem(object):
         """
         try:
             file_name = path.splitext(path.basename(file_path))[0]
-            module_name = '{}_{}_{}'.format(id(self), hash(path.abspath(file_path)), file_name)  # If two tax and benefit systems load the same module, the second one should not replace the first one. Hence this unique module name.
+
+            #  As Python remembers loaded modules by name, in order to prevent collisions, we need to make sure that:
+            #  - Files with the same name, but located in different directories, have a different module names. Hence the file path hash in the module name.
+            #  - The same file, loaded by different tax and benefit systems, has distinct module names. Hence the `id(self)` in the module name.
+            module_name = '{}_{}_{}'.format(id(self), hash(path.abspath(file_path)), file_name)
+
             module_directory = path.dirname(file_path)
             try:
                 module = load_module(module_name, *find_module(file_name, [module_directory]))
--- a/openfisca_core/taxbenefitsystems.py
+++ b/openfisca_core/taxbenefitsystems.py
@@ -137,7 +137,7 @@ class TaxBenefitSystem(object):
         """
         try:
             file_name = path.splitext(path.basename(file_path))[0]
-            module_name = '{}_{}'.format(id(self), file_name)  # If two tax and benefit systems load the same module, the second one should not replace the first one. Hence this unique module name.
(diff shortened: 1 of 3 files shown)


fpagnoux:
Anyone ? This is a nasty bug 🐛

setup.py
```diff
@@ -7,7 +7,7 @@
 
 setup(
     name = 'OpenFisca-Core',
-    version = '22.0.8',
+    version = '22.0.9',
```
MattiSG: > The impact of the change is large enough that this could have warranted a minor bump IMO: it's easier to say my tax and benefit system relies on `v22.1+` rather than `v22.0.9+` 🙂

openfisca_core/taxbenefitsystems.py
```diff
@@ -137,7 +137,7 @@ def add_variables_from_file(self, file_path):
         """
         try:
             file_name = path.splitext(path.basename(file_path))[0]
-            module_name = '{}_{}'.format(id(self), file_name)  # If two tax and benefit systems load the same module, the second one should not replace the first one. Hence this unique module name.
+            module_name = '{}_{}_{}'.format(id(self), hash(path.abspath(file_path)), file_name)  # If two tax and benefit systems load the same module, the second one should not replace the first one. Hence this unique module name.
```
guillett: Can you add a comment that justify such a definition for `module_name`?

Indeed, previously the `module_name` was only match with `endsWith` but you updated this with `==`.
My quick online investigation didn't provide me with the format of `__module__`.
Format that you explicit here with `module_name = '{}_{}_{}'.format(id(self), hash(path.abspath(file_path)), file_name)`

fpagnoux: In the regular python world, `__module__` looks like `'openfisca_france.model.revenus.activite.salarie'`, so our name is not standard.

We are doing something quite specific when loading the legislation. We dynamically load many variables files as python modules. In this context, we don't know what the standard module name would be. There may even not be one: we can load _any_ file, including files that would not be accessible via `from ... import ...`, because they are not in an installed package.

We thus need to choose a name for the module we import.

As Python remembers all loaded modules by their name, we need to take some precautions to avoid collisions:

- If we load two different files with the same file name, they must have different module names. This is the issue you reported. Hence the absolute file path hash in the module name.
- If we load the same files for two different tax and benefit system, we also want them to be separate modules, for more tricky reasons. This is #612. Hence the `id(self)` in the module name.

The `endsWith` did not make a lot of sense in the context, I think it was a remaining piece of a previous implementation.

guillett: OK, thanks for the clarification (it is matching because of `load_module(module_name,…)` where `module_name` is of our own format.)

fpagnoux:
@guillett I added a comment in the code, let me know if it's clear enough.

maukoquiroga:
Thanks @fpagnoux 😃 (I see you reused the `id(self)` trick).

PR #654, opened 2018-04-19
Fix enum dtype

### LINKED ISSUE #652: Fix EnumArray dtype type (should be np.dtype)
Hi there!

### Here is what I did:
````
    month = '2005-01'
    simulation = new_simulation(tax_benefit_system, month)
    var = simulation.calculate('housing_occupancy_status', month)
    print(var.dtype.kind)
````

No error

AttributeError: type object 'numpy.int16' has no attribute 'kind'

## Context

Bug appeared when using [`create_data_frame_by_entity`](https://github.com/openfisca/openfisca-survey-manager/blob/master/openfisca_survey_manager/scenarios.py#L302) with Enum variables.

It is linked to a wrong use of dtype in [EnumArray](https://github.com/openfisca/openfisca-core/blob/e97a3ed33bc3980a8bc27fbe81d17bdeeff9f999/openfisca_core/indexed_enums.py#L59)
wich is numpy type and not of class numpy.dtype, as you can see in the following example:
 ````
print(type(np.dtype(np.int16)))
print(np.dtype(np.int16).kind)
print(type(np.int16))
print(np.int16.kind)
````

I identify more as a:

fpagnoux:
Fixes #652 

* Fix `dtype` attribute for `EnumArray`s (returned when calculating a variable of `value_type` `Enum`):
  - It was the type `np.int16` and not the dtype instance `np.dtype(np.int16)`
  - This caused issue when trying to export an `EnumArray` with `pandas`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+### 22.0.10 [#654](https://github.com/openfisca/openfisca-core/pull/654)
+
+* Fix `dtype` attribute for `EnumArray`s (returned when calculating a variable of `value_type` `Enum`):
+  - It was the type `np.int16` and not the dtype instance `np.dtype(np.int16)`
+  - This caused issue when trying to export an `EnumArray` with `pandas`
+
 ### 22.0.9 [#650](https://github.com/openfisca/openfisca-core/pull/5O)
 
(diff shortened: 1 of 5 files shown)


benjello:
@bfabre01 : could you test it on your run ? And tell me if it works for you ?

bfabre01:
@fpagnoux @benjello : ça marche ! Merci @fpagnoux !

PR #660, opened 2018-05-12
Make package compatible with Python 3

fpagnoux:
Once this is merged, we will also need to automatically publish python 3 versions of our packages on pypi.

### DIFF
--- a/openfisca_core/parameters.py
+++ b/openfisca_core/parameters.py
@@ -5,6 +5,7 @@
 
 
 import os
+import sys
 import logging
 import traceback
 
@@ -57,7 +58,7 @@ class ParameterNotFound(AttributeError):
         message += (
             u" was not found in the {} tax and benefit system."
             ).format(instant_str)
-        super(ParameterNotFound, self).__init__(message)
+        super(ParameterNotFound, self).__init__(message.encode('utf-8'))
 
 
 class ParameterParsingError(Exception):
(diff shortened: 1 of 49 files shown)


openfisca_core/parameters.py
```diff
@@ -263,20 +263,24 @@ def validate(self, data):
...
                 )
 
     def __eq__(self, other):
         return (self.name == other.name) and (self.instant_str == other.instant_str) and (self.value == other.value)
 
     def __repr__(self):
-        return "ParameterAtInstant({})".format({self.instant_str: self.value}).encode('utf-8')
+        result = "ParameterAtInstant({})".format({self.instant_str: self.value})
+        # repr output must be encoded in Python 2, but not in Python 3
+        if sys.version_info < (3, 0):
```
Anna-Livia: Coult this be a `to_utf8` function ?It seem to happen several times in the code

fpagnoux: In both Python 2 and 3, `__repr__` must return a `str`. But `str` has a different meaning in Python 2 and 3 😒: it is a `byte` in Python 2, but a `unicode` in Python 3.

We only want to encode the string in `UTF-8` in Python 2, as in Python 3 we want to return a unicode. So if if we extracted the function, that would be `to_utf8_if_python2`, which doesn't seem great.

If all this encoding business is not clear for you, as it was for me a few days ago,  I'd recommend [this article[FR]](http://sametmax.com/lencoding-en-python-une-bonne-fois-pour-toute/). I can't pretend I understand everything now, but it's more clear 🙂.

maukoquiroga:
👏

PR #662, opened 2018-05-15
Fixup py3

fpagnoux:
Relates to #630

- Fixup Python 3 compatibility
- Automatically Publish Python 3 versions in the CI

Publishing works both for [Python 3](https://circleci.com/gh/openfisca/openfisca-core/1609) and [Python 2](https://circleci.com/gh/openfisca/openfisca-core/1610).
The Python 2 deploy script is in charge of publishing the tag and deploying the API, as this should be done only once.

PR #666, opened 2018-05-18
Fix CircleCI configuration

maukoquiroga:
Relates to #630

As discussed in #645 in [this comment](https://github.com/openfisca/openfisca-core/pull/645#issuecomment-390194875), this pull request fixes:

- The overlapping of caches between different python versions
- CircleCI workflow's job dependencies

Please note this is a small subset of #662

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,10 +1,11 @@
 version: 2
 
 jobs:
-  checkout:
+  build-2.7: &build
     working_directory: ~/openfisca-core
+
     docker:
-      - image: python:2.7.14
+      - image: python:2.7
 
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


fpagnoux:
Nice strategy to avoid duplication! Separating the caches for Python 2 and 3 was tricky, as Circle doesn't let use your own environment variable in the cache key 😞.

As discussed in Slack, and given that we seem to have[ something that seems to work right now](https://github.com/openfisca/country-template/pull/41#issuecomment-390288483), I'm in favor of delivering first, and refactoring after, as we don't know for sure what issues we'll face when we try to factorize the deployment step

MattiSG:
I believe it's worth waiting for the dust to settle on https://github.com/openfisca/country-template/pull/41 and then bring back knowledge to here.

PR #672, opened 2018-05-23
Make exit code more semantic for detect-functional-change

fpagnoux:
Mirroring https://github.com/openfisca/country-template/pull/46


maukoquiroga:
@fpagnoux Agree with @MattiSG, you need to change `detect` to `has`

fpagnoux:
> @fpagnoux Agree with @MattiSG, you need to change detect to has

To be honest and direct, I'd like to add this to the list of issues I've had with the review process recently. There are tons of PRs waiting for reviews and approval. I don't understand why I get a notification just to tell me that the review I got 3 days ago is accurate. To me, this is only noise and I'm sure ther

fpagnoux:
> @fpagnoux Agree with @MattiSG, you need to change detect to has

To be honest and direct, I'd like to add this to the list of issues I've had with the review process recently. There are tons of PRs waiting for reviews and approval. I don't understand why I get a notification just to tell me that the review I got 3 days ago is accurate. To me, this is only noise and doesn't help me. Sorry if it sounds abrasive, hope we can discuss more about that all soon 🙂.

fpagnoux:
Ok. I'm a little skeptical about the fact that this PR needed a second review approving the one I received 3 days ago though 😕.

PR #675, opened 2018-06-04
Add performance testing for the test-runner

Anna-Livia:
Connected to openfisca/openfisca-france/issues/1002

This PR allows users to check how test-running performance can be affected by the changes they made on their branch.

#### New features
- `pip install -e .[performance]` installs the  necessary dependencies
- `make performance` runs the script

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 ## 23.2.0 [#675](https://github.com/openfisca/openfisca-core/pull/675)
 
 - Add performance testing for the test-runner 
-  - `pip install -e .[performance]` installs the necessary dependencies
+  - Dependencies : 
+    - `pip install nose`
+    - `pip install --editable git+https://github.com/openfisca/openfisca-france.git@master#egg=OpenFisca-France`
   - `make performance` runs the script
 
 ### 23.1.2 [#671](https://github.com/openfisca/openfisca-core/pull/671)
--- a/setup.py
+++ b/setup.py
@@ -37,10 +37,6 @@ setup(
             ],
         'tracker': [
             'openfisca-tracker == 0.2.0',
-            ],
-        'performance': [
-            'openfisca-france >= 21.10.5, < 22.0.0',
-            'nose'
             ]
         },
     include_package_data = True,  # Will read MANIFEST.in


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 23.2.0 [#675](https://github.com/openfisca/openfisca-core/pull/675)
+
+- Add performance testing for the test-runner 
+  - `pip install -e .[performance]` installs the necessary dependencies
```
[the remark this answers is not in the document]
fpagnoux: Having said that, this makes me wonder if this performance script would not be better located in `openfisca-france`. 

I understand why we thought about putting it there, and I totally agree that France tells us a lot about Core performance, as it's the only model where we push the limits in term of model complexity.

However:

- The fact that we have to go out of our way to access these files smells like we are introducing a dependencies that is not very natural.
- We are not _really_ testing the performance of the test runner. The test runner is only a small cog in the machinery. We are testing the performance of many small (in term of number of people) OpenFisca-France simulations by using the test runner. So it's not clear, when looking at the performance result, what part is due to OpenFisca-Core in general, and what part is due to implementation choices in OpenFisca-France.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 23.2.0 [#675](https://github.com/openfisca/openfisca-core/pull/675)
```
fpagnoux: As a reuser of Core, do I really get a new feature out of this change? It seems that this feature is only targeted to Core developpers, so I'd go for `23.1.3`.

setup.py
```diff
@@ -37,6 +37,11 @@
             ],
         'tracker': [
             'openfisca-tracker == 0.2.0',
+            ],
+        'performance': [
+            'openfisca-france >= 21.10.5, < 22.0.0',
+            'snakeviz',
```
fpagnoux: You're not using snakeviz in the script.

tests/core/performance/test_perf_tests.py
```diff
@@ -0,0 +1,28 @@
...
+start_time = time.time()
+tbs = CountryTaxBenefitSystem()
+time_spent = time.time() - start_time
+print("Generate TBS: --- %s seconds ---" % (time_spent))
+
+start_time = time.time()
+openfisca_france_dir = pkg_resources.get_distribution('OpenFisca-France').location
+yaml_tests_dir = os.path.join(openfisca_france_dir, 'tests', 'mes-aides.gouv.fr')
+time_spent = time.time() - start_time
+print("Locate directories: --- %s seconds ---" % (time_spent))
```
fpagnoux: When running the script on my machine, this part is insignificant compared to the two other ones (2000 less time than the Tax Benefit System initialization). What is the point of printing its duration? I'm afraid it may bring more noise than exploitable data.

tests/core/performance/test_perf_tests.py
```diff
@@ -0,0 +1,28 @@
...
+import time
+import pkg_resources
+
+from openfisca_core.tools.test_runner import run_tests
+from openfisca_france import CountryTaxBenefitSystem
+
+start_time = time.time()
+tbs = CountryTaxBenefitSystem()
+time_spent = time.time() - start_time
+print("Generate TBS: --- %s seconds ---" % (time_spent))
```
fpagnoux: What about 
```py
print("Generate Tax Benefit System: --- {}s seconds ---".format(time_spent))
```
?
(`format` is the standard way to interpolate variables in strings since `2.6` 🙂 )

tests/core/performance/test_perf_tests.py
```diff
@@ -0,0 +1,28 @@
...
+start_time = time.time()
+openfisca_france_dir = pkg_resources.get_distribution('OpenFisca-France').location
+yaml_tests_dir = os.path.join(openfisca_france_dir, 'tests', 'mes-aides.gouv.fr')
+time_spent = time.time() - start_time
+print("Locate directories: --- %s seconds ---" % (time_spent))
+
+start_time = time.time()
+run_tests(tbs, yaml_tests_dir)
+time_spent = time.time() - start_time
+print("Pass Mes-aides tests: --- %s seconds ---" % (time_spent))
```
fpagnoux: Use format

tests/web_api/basic_case/test_variables.py
```diff
@@ -140,8 +140,8 @@ def test_dated_variable_formulas_dates():
...
+    formula_code_2016 = "def formula_2016_12(person, period, parameters):\n    age_condition = person('age', period) >= parameters(period).general.age_of_majority\n    return age_condition * parameters(period).benefits.basic_income  # This '*' is a vectorial 'if'. See https://openfisca.org/doc/coding-the-legislation/30_case_disjunction.html#simple-multiplication\n"
+    formula_code_2015 = "def formula_2015_12(person, period, parameters):\n    age_condition = person('age', period) >= parameters(period).general.age_of_majority\n    salary_condition = person('salary', period) == 0\n    return age_condition * salary_condition * parameters(period).benefits.basic_income  # The '*' is also used as a vectorial 'and'. See https://openfisca.org/doc/coding-the-legislation/25_vectorial_computing.html#forbidden-operations-and-alternatives\n"
```
fpagnoux: If this fix is due to the release of `3.1.3` for the country template, then the minimum version of `openfisca_country_template` should be updated in `setup.py`

tests/core/performance/test_perf_tests.py
```diff
@@ -0,0 +1,28 @@
+"""
+This files tests the performance of the test runner (openfisca-run-test).
+It uses the existing tests from openfisca-france because it is the largest set we currently have.
+
```
fpagnoux: Unnecessary line break

tests/core/performance/test_perf_tests.py
```diff
@@ -0,0 +1,28 @@
+"""
```
fpagnoux: Not sure about putting this file here.

It's not exactly a test of the same nature that all the unit tests that live around here, and we definitely don't want to run this in the CI.

CHANGELOG.md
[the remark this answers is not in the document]
MattiSG: Would love to see that land in Core if it was possible to add the measurement part in it, then letting each country use the tooling to assess their implementation choices  🙂

CHANGELOG.md
MattiSG: SemVer is not strict on that: the engagement is only that **if** there is a breaking change, you'll bump major. You're always free to bump minor for a patch change. In this case, if a country developer can now measure performance, I'd say it's definitely a new feature 🙂

CHANGELOG.md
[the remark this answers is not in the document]
fpagnoux: Not sure I understand. Could you give a concrete example of what parameters you would like to give (test files?), and what results you'd expect?

FI, the test runner already gives you the total time use for tests:

image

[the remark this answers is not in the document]
MattiSG: Sorry, you're right, I reacted from the notifications and did not have a clear vision of what this changeset brings in terms of functionality. What I meant was: if there is tooling to measure performance impact of changes, I'd love to have it brought to Core.

Looking at the diff, I indeed don't think this belongs to Core, due to the dependency to France but also to the lack of any machine-consumable report.

I would see the point of a performance metrics creation tool to follow the impact of changes in Core. As such though, without CI integration nor country-independent report creation, I don't see the point of sharing this in the codebase.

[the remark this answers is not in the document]
Anna-Livia: OK so two options I see are to  use the `country-template` and to put this script in `openfisca-france`.
I like the second option because right now, this is were the issues are. 
@MattiSG about the machine consumable perf testing, @sandra found the circle API that shares some interesting meta-data on testing. my first inverstigations on the subject are [here](https://github.com/openfisca/openfisca-france/issues/1002)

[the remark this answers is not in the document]
fpagnoux: I'd also support the 2nd option :)

tests/core/performance/test_perf_tests.py
fpagnoux: Seems like we are more leaning towards moving everything to France anyway.

But to answer your question, I'm not sure. I like the idea of `tests` behind executable tests that run in the CI, and crash if there is an issue. Which the performance test is not, at least in its current state. @MattiSG, any insight on good practice about where to put a performance measurement script?

MattiSG: This is called _continuous inspection_ and is usually driven by metadata derived from test runs. Other metadata that is usually extracted are coverage and churn.

MattiSG:
I don't really understand what has happened. There is still France-dependant code in Core, but now that dependency is not even listed in `setup.py`. This is worse for me 😞

PR #680, opened 2018-06-25
Fix test_variables.py failing test

maukoquiroga:
#### Fix failing tests

* Failing test at `tests/web_api/basic_case/test_variables.py:146`
  - Changed `http` to `https`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 23.1.3 [#680](https://github.com/openfisca/openfisca-core/pull/680)
+
+Fix test that was failing due to migration to HTTPS
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+### 23.1.3 [#680](https://github.com/openfisca/openfisca-core/pull/680)
+
+* Failing test at `tests/web_api/basic_case/test_variables.py:146`
```
fpagnoux: Changelog should always be written from the point of views of users. Prefer active sentences, and describe what has changed:

For instance:
> Fix test that was failing due to migration to HTTPS
or 
> Use HTTPS URL in tests

Also, the exact file and line number of the change is rarely relevant, as this information is easily accessible through the diff.

PR #686, opened 2018-07-10
Update psutil version to 5.4.6

maukoquiroga:
#### Technical changes

- Require `psutil` version 5.4.6
- Installation fails with Python 3.7.0 on Windows

See https://github.com/openfisca/openfisca-senegal/pull/10#discussion_r201389882

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+### 23.1.7 [#686](https://github.com/openfisca/openfisca-core/pull/686)
+
+* Fix installation on Windows with Python 3.7
+  - Require `psutil` version `5.4.6`, as `5.4.2` is incompatible with that environment.
+
 ### 23.1.6 [#688](https://github.com/openfisca/openfisca-core/pull/688)
 
 * In the error message sent to a user trying to set a variable without specifying for which period, add an example for a variable defined for `ETERNITY`.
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -49,7 +49,7 @@
         'flask-cors == 3.0.2',
         'gunicorn >= 19.7.1',
         'numpy >= 1.11, < 1.15',
-        'psutil == 5.4.2',
+        'psutil >= 5.4.6, < 6.0.0',
```
fpagnoux: Are we sure that `psutil` applies semantic versioning?

maukoquiroga: I'm not sure, but I believe so, i.e from the changelog it seems API changes are grouped by majors:

- https://github.com/giampaolo/psutil/blob/master/HISTORY.rst#100
- https://github.com/giampaolo/psutil/blob/master/HISTORY.rst#200


But there's no reference to semantic versioning in the doc.

Safer option: `==5.4.6` ?

fpagnoux: I'll leave it up to your judgment, reading the other dependencies, it seems like we don't have a clear policy about that 🙂.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+### 23.1.5 [#686](https://github.com/openfisca/openfisca-core/pull/686)
+
+* Require at least `psutil` version 5.4.6
```
fpagnoux: > To make the Changelog more user-centric, I'd prefer something like:
> 
> * Fix installation on Windows with Python 3.7 
>   - Require `psutil` version `5.4.6`, as `5.4.2` was incompatible with this environment

Issue #692, opened 2018-07-23
Failure on Core 23.3.0 make test with python 3.5.5


sandcha:
Hi there!

I really enjoy OpenFisca, but I recently encountered an issue.

### Here is what I did:
On openfisca-core v.`23.3.0` master branch, I ran:
`make test`

in a virtualenv where I did:
```
pip install -U pip wheel
pip install -e .[test]
```

### Here is what I expected to happen:
```
[openfisca-core]$ make test
flake8
nosetests
....................................................................................................................................................................................................................................................................................................................................
----------------------------------------------------------------------
Ran 324 tests in 5.625s

OK
```

### Here is what actually happened:
```
c690-py35[openfisca-core]$ make test
flake8
nosetests
...............................................................................................................................................................................................................................................EEEEEEEEEEEEEEEEEEEEEEEEEEE.......E..EEE.........EEEEE...
======================================================================
ERROR: web_api.basic_case.test_calculate.test_responses('{"a" : "x", "b"}', <HTTPStatus.BAD_REQUEST: 400>, 'error', 'Invalid JSON')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/sch/.local/share/virtualenvs/c690-py35/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/sch/Dev/github/openfisca-core/tests/web_api/basic_case/test_calculate.py", line 26, in check_response
    json_response = json.loads(response.data)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'bytes'
```
...
```
======================================================================
ERROR: Failure: TypeError (the JSON object must be str, not 'bytes')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/sch/.local/share/virtualenvs/c690-py35/lib/python3.5/site-packages/nose/failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "/Users/sch/.local/share/virtualenvs/c690-py35/lib/python3.5/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/Users/sch/.local/share/virtualenvs/c690-py35/lib/python3.5/site-packages/nose/importer.py", line 47, in importFromPath
[...]

PR #698, opened 2018-07-24
Fix encoding error

fpagnoux:
Fixes https://github.com/openfisca/openfisca-core/issues/692

@sandcha can you confirm it solves the issue? I don't have python `3.5` 🙂

### DIFF
--- a/tests/web_api/basic_case/test_calculate.py
+++ b/tests/web_api/basic_case/test_calculate.py
@@ -23,7 +23,7 @@ def post_json(data = None, file = None):
 def check_response(data, expected_error_code, path_to_check, content_to_check):
     response = post_json(data)
     assert_equal(response.status_code, expected_error_code)
-    json_response = json.loads(response.data)
+    json_response = json.loads(response.data.decode('utf-8'))
     if path_to_check:
         content = dpath.util.get(json_response, path_to_check)
         assert_in(content_to_check, content)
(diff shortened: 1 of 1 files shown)


sandcha:
Solved 27 errors out of 36 🎉 
I'm checking last ones.

.circleci/has-functional-changes.sh
```diff
@@ -1,6 +1,6 @@
 #! /usr/bin/env bash
 
-IGNORE_DIFF_ON="README.md CONTRIBUTING.md .gitignore .circleci/* .github/*"
+IGNORE_DIFF_ON="README.md CONTRIBUTING.md .gitignore .circleci/* .github/* tests/*"
```
sandcha: 👍

guillett: Are we sure we want to ignore `tests` in the functional change detection ?

I understand this is not convenient that that specific PR but IMHO a change in the expected behavior requires some form of documentation in the changelog.

My 2 cents.

fpagnoux: I feel like a "change in the expected behavior" would come with a change in the core codebase, and therefore would have to me documented in the changelog.

If there is _only_ changes in `tests`, and none in the real packages, can there really be a  change? Do you have a concrete example in mind?

fpagnoux: Also before migrating to CircleCi 2 et changing the conf, `tests` was already ignored.

guillett: I thought about test refactorings and change in test accuracy.

Issue #683, opened 2018-07-09
Track metadata in Piwik


maukoquiroga:
Relates to openfisca/openfisca-core#685
Requires openfisca/tracker/issues/4

**For** maintainers.

**Story:**

When deploying new API versions,
I want to know the detailed usage metrics for each of the currently supported versions,
so I can provide a better overall experience to our users.

**Definition of done:**

In order to have a better view of the API's usage, track :

- [ ] ~Core version~
- [x] Country package name
- [x] Country package version

MattiSG:
Looks like a duplicate of https://github.com/openfisca/openfisca-core/issues/683.

Closing this for now, feel free to reopen if I misqualified 🙂

fpagnoux:
I think we already have the name and country package version:

![image](https://user-images.githubusercontent.com/11834997/44536684-01964f80-a6cb-11e8-9026-38d0b03effb7.png)

I'm however not sure what the core version would bring. @maukoquiroga in which case would this piece of information help you?

maukoquiroga:
Hi @fpagnoux, effectively the current information available is enough. I've updated description.

PR #700, opened 2018-07-27
Track user ip through proxy

Anna-Livia:
connected to core #683 

This PR fixes the web API tracking so that the user's IP will be sent to the tracker even when the requests goes through a proxy (nginx)

#### Technical changes

- Adds a condition to send the first IP of the  `'X-Forwarded-For'` variable set by nginx, as the user's IP.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### 23.4.1 [#700](https://github.com/openfisca/openfisca-core/pull/700)
 
-* Send the web API user's IP address to the tracker even when it goes through a proxy.
+* Fix API source IP detection through proxies.
 
 ## 23.4.0 [#694](https://github.com/openfisca/openfisca-core/pull/694)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 23.4.1 [#700](https://github.com/openfisca/openfisca-core/pull/700)
+
+* Send the web API user's IP address to the tracker even when it goes through a proxy.
+
 ## 23.4.0 [#694](https://github.com/openfisca/openfisca-core/pull/694)
 
 * Use `/` rather than `.` in the path to access a parameter:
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+### 23.3.2 [#700](https://github.com/openfisca/openfisca-core/pull/700)
+
+* Send the web API user's IP address to the tracker even when it goes through a proxy.
```
MattiSG: I feel this changelog reads very ominous and gives a wrong idea about how tracking is done.
Suggestion: `Fix API source IP detection through proxies`.

openfisca_web_api_preview/app.py
```diff
@@ -167,12 +167,18 @@ def apply_headers(response):
 
     @app.after_request
     def track_requests(response):
+
         if tracker:
+            api_version = "{}-{}".format(data['country_package_metadata']['name'],
+                                         data['country_package_metadata']['version'])
             if request.headers.get('dnt'):
-                tracker.track(request.url)
+                user_ip = ""
```
MattiSG: > `source_ip` or even just `ip` would read clearer than `user_ip` to me, as it might not be a human user.

openfisca_web_api_preview/app.py
```diff
@@ -167,12 +167,18 @@ def apply_headers(response):
 
     @app.after_request
     def track_requests(response):
+
         if tracker:
+            api_version = "{}-{}".format(data['country_package_metadata']['name'],
```
sandcha: Move this line next to `api_version` call (before `tracker.track(...)`?

Anna-Livia:
@MattiSG Is is good for you ?

openfisca_web_api_preview/app.py
```diff
@@ -167,12 +167,20 @@ def apply_headers(response):
 
     @app.after_request
     def track_requests(response):
+
         if tracker:
+
```
MattiSG: These empty lines read weird to me 😉

PR #704, opened 2018-08-03
Actually take tracker token into account

fpagnoux:
Not sure exactly what this token does, but it was certainly not having any effect 😅 

I'll leave the changelog writing to someone who knows a little better what this is all about 🙂


Anna-Livia:
Good catch ! :) 
However, the tracking is working pretty well without it.... Do you have some idea of what adding this will bring ?

fpagnoux:
This token was introduced to track user IP... I've no idea why everything seemed to have work without it...

If we don't have any explanation to give, maybe keeping this change in #705 (which needs it) is fine...

fpagnoux:
Superseded by #705

Issue #615, opened 2018-01-25
Make the API discoverable from the API root


maukoquiroga:
Related to #512

**For** developers.

**Stories:**

When I’m considering working with the Web API,
I want to know if I'm in the right place,
so I can continue working without being frustrated.

When I'm working with the Web API,
I want to know the most important endpoints,
and I want to be able to navigate them,
so I can build my app more efficiently.

**Definition of done:**

The idea is to render a helpful response to :

- [ ] Let people know they're in the right place
- [ ] Let people request the root of API to get a list of the most important endpoints

Anna-Livia:
Proposition  when accessing the root of the API: 
{
      "openfisca-country-package" : "country-template-3.12.3",
      links: {
                   "list of all the parameters" : "//parameters"
                   "list of all the variables" : "//variables"
                    "description of a parameter": "//parameter/{name of the parameter}"
                   "description of a variable": "//variable/{name of the variable}"
                   "Calculate variables for a given situation": "//calculate"
                    "OpenAPI specifications": "//specs/"
      }
}

maukoquiroga:
Love the idea (looks a lot like https://api.github.com/).

I would:

- Use the full path, otherwise it won't be truly discoverable
- Only enumerate resources (i.e. `parameters` is a resource, `list` or `calculate` is an operation)

Example:

```json
{
  "info": {
    "country-package": {
      "title": "country-template",
      "version": "3.12.3"
    }
  },
  "links": [
    {
      "rel": "self",
      "description": null,
      "href": "https://fr.openfisca.org/api/v21"
    }, {
      "rel": "parameters",
      "description": null,
      "href": "https://fr.openfisca.org/api/v21/parameters"
    }, {
      "rel": "parameter",
      "description": null,
      "href": "https://fr.openfisca.org/api/v21/parameter/{parameter_id}"
    }, {
      "rel": "variables",
      "description": null,
      "href": "https://fr.openfisca.org/api/v21/variables"
    }, {
      "rel": "variable",
      "description": null,
      "href": "https://fr.openfisca.org/api/v21/variable/{variable_id}"
    }, {
      "rel": "spec",
      "description": "OpenAPI specifications",
      "href": "https://fr.openfisca.org/api/v21/specs"
    }
  ]
}
```

Anna-Livia:
I feel that the `github model` is more readable.

> Only enumerate resources (i.e. parameters is a resource, list or calculate is an operation)
[...]

PR #705, opened 2018-08-03
Make API discoverable from the root

fpagnoux:
Connected to #615 

* On the Web API, expose a welcome message (with a 300 code) on `/` instead of a 404 error.

For instance, `curl -i localhost:5000` gives:

```
HTTP/1.1 300 MULTIPLE CHOICES
(...)

{
  "welcome": "This is the root of an OpenFisca Web API. To learn how to use it, check the general documentation (https://openfisca.org/doc/api) and the OpenAPI specification of this instance (http://localhost:5000/spec)."
}
```

* This message can be customized:

If the Web API is started with `openfisca serve -p 3000 --welcome-message "Welcome to the OpenFisca-France Web API. To learn how to use it, check our interactive swagger documentation: https://fr.openfisca.org/legislation/swagger."`

Then `curl -i localhost:5000` gives:

```
HTTP/1.1 300 MULTIPLE CHOICES
(...)

{
  "welcome": "Welcome to the OpenFisca-France Web API. To learn how to use it, check our interactive swagger documenation: https://fr.openfisca.org/legislation/swagger"
}
```

(Like other configuration variables, this custom message can also be defined in a configuration file. Check the [openfisca serve documentation](https://openfisca.readthedocs.io/en/latest/openfisca_serve.html))

### DIFF
--- a/openfisca_web_api_preview/app.py
+++ b/openfisca_web_api_preview/app.py
@@ -58,6 +58,22 @@ def create_app(tax_benefit_system,
 
     data = build_data(tax_benefit_system)
 
+    def resolve_host(request):
+        if data['host'] is not None:
+            return data['host']
+        host = request.host_url
+        data['host'] = host
+        data['openAPI_spec']['host'] = host
+        return host
+
+    DEFAULT_WELCOME_MESSAGE = "This is the root of an OpenFisca Web API. To learn how to use it, check the general documentation (https://openfisca.org/doc/api) and the OpenAPI specification of this instance ({}spec)."
+
+    @app.route('/')
+    def get_root():
+        return jsonify({
+            'welcome': DEFAULT_WELCOME_MESSAGE.format(resolve_host(request))
+            }), 300
+
     @app.route('/parameters')
     def get_parameters():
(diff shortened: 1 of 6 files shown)


openfisca_web_api_preview/app.py
```diff
@@ -81,6 +98,8 @@ def get_variable(id):
 
     @app.route('/spec')
     def get_spec():
+        if data['host'] is None:
```
Anna-Livia: This condition is checked in `def resolve_host`. Isn't it redundant to have it here also ?

fpagnoux: The `spec` route can be called before the `/` one. In that case, `data['host']` is still `None` when we get here, so we need to resolve the host.

fpagnoux: However, I was just told that re-using the host from one request to another was risky (one local request, and everyone is seeing a `localhost:3000` as a host) and not very stateless, so we need a different approach 😞

fpagnoux: Ok, adapted the code :)

PR #712, opened 2018-08-21
Add dump & restore to holders & simulations.

### LINKED ISSUE #661: Provide a way to dump to/restore fom disk the state of a simulation
Hi there!

### Here is what I want to do:

I would like to save the state of a simulation  to be able to restart from that exact point several times.
I should thus be able to dump it and reload it.
I think that a good starting point could be the work done by @fpagnoux to allow out-of-memory computations with variables saved on disk.

## Context

I identify more as a:

I want to build simulation on huge data and save time at initialisation by always loading the same initial state.
I want to be able to save individual states to run multiple reforms on them.

eraviart:
#### New features

- Add dump & restore to holders & simulations (fix #661).

-------------------

EDIT (@fpagnoux)

Dump:

```py
from openfisca_core.tools.simulation_dumper import  dump_simulation, 
dump_simulation(simulation, '/path/to/directory')
```

Restore:

```py
from openfisca_core.tools.simulation_dumper import restore_simulation
simulation_2 = restore_simulation('/path/to/directory', tax_benefit_system)
```

### DIFF
--- a/tests/core/test_dump_restore.py
+++ b/tests/core/test_dump_restore.py
@@ -1,141 +1,43 @@
 # -*- coding: utf-8 -*-
 
-
 from __future__ import unicode_literals, print_function, division, absolute_import
 
 import shutil
 import tempfile
 
-import numpy as np
+from numpy.testing import assert_array_equal
 
-from openfisca_core.variables import Variable
-from openfisca_core.periods import MONTH
-from openfisca_country_template import CountryTaxBenefitSystem
-from openfisca_country_template.entities import Person
+from openfisca_core.simulations import Simulation
+from openfisca_country_template.situation_examples import couple
+from openfisca_core.tools.simulation_dumper import dump_simulation, restore_simulation
 
-
-class input(Variable):
-    value_type = int
-    entity = Person
-    label = "Input variable"
-    definition_period = MONTH
+from .test_countries import tax_benefit_system
 
 
-class intermediate(Variable):
-    value_type = int
-    entity = Person
-    label = "Intermediate value"
-    definition_period = MONTH
+def test_dump():
+    directory = tempfile.mkdtemp(prefix = "openfisca_")
+    simulation = Simulation(tax_benefit_system = tax_benefit_system, simulation_json = couple)
+    calculated_value = simulation.calculate('disposable_income', '2018-01')
+    dump_simulation(simulation, directory)
 
-    def formula(person, period):
-        return person('input', period)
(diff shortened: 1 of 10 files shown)


benjello:
@fpagnoux : I took the liberty to ask you to review this PR.
@eraviart coded this during a session at IPP.
We need your input before we go forward and we add higher level methods to `SurveyScenario`.

fpagnoux:
Thanks for the PR @eraviart!

A few general questions @benjello before looking more into details:

- With this implem, after we restore the simulation, nothing is loaded in memory and everything is kept on disk. Is that on purpose? This could significantly slow down calculations after restoring. 
- Entities compositions and roles are not stored, is that an issue?

benjello:
@fpagnoux : I knew that your input would be valuable !

- You first point is a very good one. For performance issue we may want that  the variables are kept in RAM when read.
- The second point is another good one, we definitively want to keep those if we want to compute more  variables involving several entities after restoring. I completely forgot that when explaing the problem to @eraviart.

fpagnoux:
Ok, I will try adjust the implem to take these two points into account :)

fpagnoux:
Tested on France calculating `revenu_disponible` for 40000 persons, dumping and restoring takes a reasonable time (<10s)

Mahdi, do you want to try the branch sees if it works for you, or should we go ahead and ask for a review?

benjello:
Thanks @fpagnoux !

Do you confirm that when restoring a simulation, the variables are kept back in RAM/disk according to the new `memroy_config` only when recalled and not restored at once ? 

I think we should ask for a review so I can go on and propagate to downstream packages. 
I could definitively come back and ask for a few changes if needed by a use case.

tests/core/test_dump_restore.py
```diff
@@ -1,141 +1,43 @@
...
-    shutil.rmtree(data_storage_dir)
+from numpy.testing import assert_array_equal
 
+from openfisca_core.simulations import Simulation
+from openfisca_country_template.situation_examples import couple
+from openfisca_core.tools.simulation_dumper import dump_simulation, restore_simulation
 
-def test_dump_restore_different_simulations_different_variables():
-    data_storage_dir = tempfile.mkdtemp(prefix = "openfisca_")
+from .test_countries import tax_benefit_system
```
sandcha: Use absolute import?

fpagnoux: We can't use absolute imports as the `tests` are not packaged in `openfisca_core`.

benjello:
Looks good to merge then @fpagnoux !

PR #714, opened 2018-08-23
 Add a `/entities` endpoint to the API

### LINKED ISSUE #561: Add a `/entities` endpoint to the API
Returns a JSON object listing all modelled entities where the property names is the `key` property of the entity and where the value is an Object with the following properties:

- String `plural`. Mandatory.
- String `documentation`, coming from the entity docstring. Defaults to `""`.
- Object `roles` if the entity is not a person (in case the entity is a person, `roles` must be falsy), where the name is the `key` property and where the value is an object with the following properties:
  - String `plural`. Mandatory.
  - Boolean `mandatory`. Defaults to `false`.
  - Number `max` (if one is defined).

`subroles` are not to be exposed yet.

Related to https://github.com/openfisca/legislation-explorer/issues/80.

Anna-Livia:
Fixes #561 
#### New features
- Get information on a country package's entities, and their roles
- Introduce the `/entities` endpoint for the Web API
```json
{
            "description": "[...]",
            "plural": "households",
            "roles": {
                "parent": {
                    "description": "The one or two adults in charge of the household.",
                    "mandatory": "true",
                    "max": 2,
                    "plural": "parents"
                    },
                "child": {
                    "description": "Other individuals living in the household.",
                    "mandatory": "false",
                    "plural": "children"
                    }
                }
            }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,28 @@
 # Changelog
 
+### 24.3.0 [#714](https://github.com/openfisca/openfisca-core/pull/714)
+
+- Introduce the `/entities` endpoint for the Web API.
+  - Expose information about the country package's entities, and their roles.
+```json
+{
+            "description": "Household",
+            "documentation": "Household is an example of a group entity. A group entity contains one or more individual·s.[...]",
+            "plural": "households",
+            "roles": {
+                "parent": {
+                    "description": "The one or two adults in charge of the household.",
+                    "max": 2,
+                    "plural": "parents"
+                    },
+                "child": {
+                    "description": "Other individuals living in the household.",
+                    "plural": "children"
+                    }
+                }
+            }
+```
+
 ## 24.2.0 [#712](https://github.com/openfisca/openfisca-core/pull/712)
 
 - Allow to dump and restore a simulation in a directory
--- a/openfisca_web_api/loader/entities.py
+++ b/openfisca_web_api/loader/entities.py
@@ -5,7 +5,8 @@ from __future__ import unicode_literals, print_function, division, absolute_impo
 def build_entities(tax_benefit_system):
     entities = {
         entity.key: {
-            "plural": entity.plural
+            "plural": entity.plural,
+            "description": entity.doc
         }
         for entity in tax_benefit_system.entities
     }
--- a/openfisca_web_api/loader/entities.py
+++ b/openfisca_web_api/loader/entities.py
@@ -1,0 +1,14 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals, print_function, division, absolute_import
+
+def build_entities(tax_benefit_system):
+    entities = {
+        entity.key: {
+            "plural": entity.plural
+        }
+        for entity in tax_benefit_system.entities
+    }
+    return entities
+
+
--- a/openfisca_web_api/loader/entities.py
+++ b/openfisca_web_api/loader/entities.py
@@ -20,7 +20,7 @@ def build_entity(entity):
         'description': to_unicode(entity.label),
         'documentation': formatted_doc
         }
-    if hasattr(entity, 'roles'):
+    if not entity.is_person:
         formatted_entity['roles'] = {
             role.key: build_role(role)
             for role in entity.roles
@@ -39,5 +39,4 @@ def build_role(role):
     if role.subroles:
         formatted_role['max'] = len(role.subroles)
 
-    formatted_role['mandatory'] = True if formatted_role.get('max') else False
     return formatted_role
--- a/openfisca_web_api/loader/entities.py
+++ b/openfisca_web_api/loader/entities.py
@@ -14,29 +14,29 @@ def build_entities(tax_benefit_system):
 
 def build_entity(entity):
 
-    entity_formated = {
+    formatted_entity = {
         'plural': entity.plural,
-        'description': to_unicode(entity.doc)
+        'description': to_unicode(entity.label),
+        'documentation': to_unicode(entity.doc)
         }
     if hasattr(entity, 'roles'):
-        entity_formated['roles'] = \
-            {
+        formatted_entity['roles'] = {
             role.key: build_role(role)
             for role in entity.roles
             }
-    return entity_formated
+    return formatted_entity
 
 
 def build_role(role):
-    role_formated = {
+    formatted_role = {
         'plural': role.plural,
         'description': role.doc
         }
 
     if role.max:
-        role_formated['max'] = role.max
+        formatted_role['max'] = role.max
     if role.subroles:
-        role_formated['max'] = len(role.subroles)
+        formatted_role['max'] = len(role.subroles)
 
-    role_formated['mandatory'] = True if role_formated.get('max') else False
-    return role_formated
+    formatted_role['mandatory'] = True if formatted_role.get('max') else False
+    return formatted_role
--- a/openfisca_web_api/loader/entities.py
+++ b/openfisca_web_api/loader/entities.py
@@ -2,14 +2,40 @@
 
 from __future__ import unicode_literals, print_function, division, absolute_import
 
+
 def build_entities(tax_benefit_system):
     entities = {
-        entity.key: {
-            "plural": entity.plural,
-            "description": entity.doc
-        }
+        entity.key: build_entity(entity)
         for entity in tax_benefit_system.entities
-    }
+        }
     return entities
 
 
+def build_entity(entity):
+
+    entity_formated = {
+        "description": entity.doc,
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
 # Changelog
 
+### 24.1.0 [#714](https://github.com/openfisca/openfisca-core/pull/714)
+
+- Get information on a country package's entities, and their roles
+- Introduce the `/entities` endpoint for the Web API
+```json
+{
+            "description": "[...]",
```
fpagnoux: Could we make up a simple description? Examples will real values are always easier to grasp 🙂

openfisca_web_api/loader/entities.py
```diff
@@ -0,0 +1,42 @@
...
+        for entity in tax_benefit_system.entities
+        }
+    return entities
+
+
+def build_entity(entity):
+
+    entity_formated = {
+        'plural': entity.plural,
+        'description': to_unicode(entity.doc)
```
fpagnoux: #561 mentioned a `documentation` key containing the docstring. 
If we are to have a `description` key, it should probably contain `entity.label`

openfisca_web_api/loader/entities.py
```diff
@@ -0,0 +1,42 @@
...
+        }
+    if hasattr(entity, 'roles'):
+        entity_formated['roles'] = \
```
fpagnoux: Avoid `\` for line continuation: you can rely on implicit continuation, as you've done everywhere else 🙂

openfisca_web_api/loader/entities.py
```diff
@@ -0,0 +1,42 @@
...
+    return entities
+
+
+def build_entity(entity):
+
+    entity_formated = {
+        'plural': entity.plural,
+        'description': to_unicode(entity.doc)
+        }
+    if hasattr(entity, 'roles'):
```
fpagnoux: We usually use:

 ```py
if entity.is_person:
```

File: openfisca_web_api/loader/entities.py [hunk 5]
```diff
@@ -0,0 +1,42 @@
...
+def build_role(role):
+    role_formated = {
+        'plural': role.plural,
+        'description': role.doc
+        }
+
+    if role.max:
+        role_formated['max'] = role.max
+    if role.subroles:
+        role_formated['max'] = len(role.subroles)
```
fpagnoux: 👍

openfisca_web_api/loader/entities.py
```diff
@@ -0,0 +1,42 @@
...
+
+    role_formated['mandatory'] = True if role_formated.get('max') else False
```
fpagnoux: No, there is no good reason to assume than a role with a max is a mandatory role. 

Right now, we don't handle mandatory role in OpenFisca Core. I see two solutions to deal with that:
1. Always set `role_formated['mandatory']` to `False` (the Web API would be "anticipating" the feature)
1. Open another PR in Core introducing this `mandatory` attribute in Python


1 allows for a quick delivery, but leads to a relatively inconsistent state of the codebase.

openfisca_web_api/openAPI.yml
```diff
@@ -16,13 +16,18 @@ tags:
   - name: "Parameters"
     description: "A parameter is a numeric property of the legislation that can evolve over time."
     externalDocs:
-      description: "Parameter official documentation"
+      description: "Official Parameter documentation"
```
fpagnoux: > Not we sure what we mean by "official". Never seen an fan-based documentation so far, nor an off-brand one ;)

openfisca_web_api/loader/entities.py
```diff
@@ -0,0 +1,42 @@
...
+    entities = {
+        entity.key: build_entity(entity)
+        for entity in tax_benefit_system.entities
+        }
+    return entities
+
+
+def build_entity(entity):
+
+    entity_formated = {
```
fpagnoux: `formatted_entity` ?

tests/web_api/test_entities.py
```diff
@@ -0,0 +1,54 @@
...
+# /entities
+
+
+def test_return_code():
+    assert_equal(entities_response.status_code, OK)
+
+
+def test_response_data():
+    entities = json.loads(entities_response.data.decode('utf-8'))
+    test_description = '''
```
fpagnoux: Copy-pasting the description makes the test really fragile. A minor edition of the doc would break it.
I'd recommend importing `Household` from the country template and getting the doc from `Household.doc`

tests/web_api/test_entities.py
```diff
@@ -0,0 +1,54 @@
...
+'''
+    assert_equal(
+        entities['household'],
+        {
+            'description': test_description,
+            'plural': 'households',
+            'roles': {
+                'parent': {
+                    'description': 'The one or two adults in charge of the household.',
+                    'mandatory': True,
```
fpagnoux: Should probably stick to `False` unless we actually handle mandatory roles.

CHANGELOG.md
fpagnoux: Ok, in my previous comment I thought description was the label, now I understand better why you didn't want to write the full `Household` doc. 
But I still think a one or two-sentence max doc would hep :)

openfisca_web_api/loader/entities.py
Anna-Livia: I think, pycharm outsmarted me. ;)

openfisca_web_api/loader/entities.py
fpagnoux: Fair point. 

We have never used a group entity without roles, and right now it would trigger an ugly error, but you're right, there could be use cases.

However, even if we had a group entity without roles, it would still have an attribute `roles` that would be `[]`, so the code further down would still work. In general, using `hasattr` is pretty rare unless you are doing relatively tricky stuff.

openfisca_web_api/loader/entities.py
Anna-Livia: I implemented 1 and opened #721 to follow 2

openfisca_web_api/loader/entities.py
```diff
@@ -0,0 +1,42 @@
...
+        }
+    return entities
+
+
+def build_entity(entity):
+
+    formatted_entity = {
+        'plural': entity.plural,
+        'description': to_unicode(entity.label),
+        'documentation': to_unicode(entity.doc)
```
fpagnoux: `documentation` seems to often start and end with `\n`, which is not really relevant. We should probably clean these :)

openfisca_web_api/openAPI.yml
```diff
@@ -318,6 +338,14 @@ definitions:
       type: "number"
       format: "float"
 
+  Entities:
+    type: "object"
+    additionalProperties:
+      type: "object"
+      properties:
+        definition:
```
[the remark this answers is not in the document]
Anna-Livia: and `roles` ?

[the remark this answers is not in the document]
fpagnoux: Yep, forgot that one 😓

openfisca_web_api/loader/entities.py
```diff
@@ -0,0 +1,42 @@
...
+        'plural': role.plural,
+        'description': role.doc
+        }
+
+    if role.max:
+        formatted_role['max'] = role.max
+    if role.subroles:
+        formatted_role['max'] = len(role.subroles)
+
+    formatted_role['mandatory'] = False
```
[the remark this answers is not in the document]
Anna-Livia: I totally agree

openfisca_web_api/loader/entities.py
Anna-Livia: I found a way but it is not very elegant.... I'm looking forward to your feedback on it :)

openfisca_web_api/loader/entities.py
```diff
@@ -0,0 +1,45 @@
...
+        for entity in tax_benefit_system.entities
+        }
+    return entities
+
+
+def build_entity(entity):
+    split_doc = to_unicode(entity.doc).split("\n")
+    formatted_start_doc = split_doc[1:] if split_doc[0] == "" else split_doc
+    formatted_end_doc = formatted_start_doc[:-1] if formatted_start_doc[-1] == "" else formatted_start_doc
+    formatted_doc = "\n".join(formatted_end_doc)
```
fpagnoux: I think you can just use `formatted_doc  = to_unicode(entity.doc.strip())`

`strip` removes all trailing white spaces, and `\n` is considered as a particular white space 🎉

Anna-Livia: ... 🤦

PR #717, opened 2018-08-27
Add editorial content on parameter and variable

sandcha:
Connected to openfisca/legislation-explorer#124

#### New features

Allow for multiline editorial content on parameters and variables:
- Introduce `ParameterNode.documentation`, `Parameter.documentation` and `Variable.documentation` attributes
- User can access a `documentation` attribute through Web API
    - on `/parameter` nodes as `/parameter/benefits`
      > = python `ParameterNode.documentation`  
      > = `documentation` string attribute of `index.yaml`
    - on `/parameter` leafs as `/parameter/benefits/housing_allowance`
      > = python `Parameter.documentation`  
      > = yaml `documentation` string attribute
    - on `/variable` leafs as `/variable/housing_allowance`
      > = python `Variable.documentation`
    - on every `/variable` leaf formula 
      > = python `Variable` formula docstring

Here is web API `/variable/housing_allowance` example:

      {
        "description": "Housing allowance",
        "documentation": "This allowance was introduced on the 1st of Jan 1980.\nIt needs the   'rent' value (same month) but doesn't care about the 'housing_occupancy_status'.",
        "entity": "household",
        "formulas": {
          "1980-01-01": {
            "content": "def formula_1980(household, period, parameters):\n    '''This is my specific life statement.'''\n    return     household('rent', period) * parameters(period).benefits.housing_allowance\n",
            "documentation": "This is my specific life statement.",
            "source": "https://github.com/openfisca/country-template/blob/3.3.0/openfisca_country_template/variables/benefits.py#L47-L49"
          },
          "2016-12-01": null
        },
        "id": "housing_allowance",
        (...)
      }

The formula doctoring is in the formula `content` and `documentation` attributes.

### DIFF
--- a/openfisca_core/parameters.py
+++ b/openfisca_core/parameters.py
@@ -91,7 +91,7 @@ class Parameter(object):
         :param string name: Name of the parameter, e.g. "taxes.some_tax.some_param"
         :param dict data: Data loaded from a YAML file.
         :param string file_path: File the parameter was loaded from.
-        :param string documentation: Some documentation describing parameter usage and context.
+        :param string documentation: Documentation describing parameter usage and context.
 
 
         Instantiate a parameter without metadata:
@@ -321,7 +321,7 @@ class ParameterNode(object):
         :param string name: Name of the node, eg "taxes.some_tax".
         :param string directory_path: Directory containing YAML files describing the node.
         :param dict data: Object representing the parameter node. It usually has been extracted from a YAML file.
-        :param string documentation: Some documentation describing parameter node usage and context.
+        :param string documentation: Documentation describing parameter node usage and context.
         :param string file_path: YAML file from which the `data` has been extracted from.
 
 
@@ -367,8 +367,8 @@ class ParameterNode(object):
                     if child_name == 'index':
                         data = _load_yaml_file(child_path)
                         _validate_parameter(self, data, allowed_keys = ['metadata', 'description', 'documentation', 'reference'])
-                        self.description = data.get('description', None)
-                        self.documentation = data.get('documentation', None)
+                        self.description = data.get('description')
+                        self.documentation = data.get('documentation')
                         _set_backward_compatibility_metadata(self, data)
                         self.metadata.update(data.get('metadata', {}))
                     else:
@@ -387,9 +387,8 @@ class ParameterNode(object):
         else:
             self.file_path = file_path
             _validate_parameter(self, data, data_type = dict, allowed_keys = self._allowed_keys)
-            # We allow to set a reference, a description and a documentation for a node.
-            self.description = data.get('description', None)
-            self.documentation = data.get('documentation', None)
+            self.description = data.get('description')
+            self.documentation = data.get('documentation')
             _set_backward_compatibility_metadata(self, data)
             self.metadata.update(data.get('metadata', {}))
             for child_name, child in data.items():
--- a/openfisca_core/parameters.py
+++ b/openfisca_core/parameters.py
@@ -121,18 +121,22 @@ class Parameter(object):
         _validate_parameter(self, data, data_type = dict)
         self.description = None
         self.metadata = {}
+        self.documentation = None
         self.values_history = self  # Only for backward compatibility
 
         # Normal parameter declaration: the values are declared under the 'values' key: parse the description and metadata.
         if data.get('values'):
             # 'unit' and 'reference' are only listed here for backward compatibility
-            _validate_parameter(self, data, allowed_keys = set(['values', 'description', 'metadata', 'unit', 'reference']))
+            _validate_parameter(self, data, allowed_keys = set(['values', 'description', 'metadata', 'unit', 'reference', 'documentation']))
             self.description = data.get('description')
+
             _set_backward_compatibility_metadata(self, data)
             self.metadata.update(data.get('metadata', {}))
 
             _validate_parameter(self, data['values'], data_type = dict)
             values = data['values']
+
+            self.documentation = data.get('documentation')
 
         else:  # Simplified parameter declaration: only values are provided
             values = data
--- a/openfisca_core/parameters.py
+++ b/openfisca_core/parameters.py
@@ -88,9 +88,10 @@ class Parameter(object):
     """
         A parameter of the legislation. Parameters can change over time.
 
-        :param name: name of the parameter, e.g. "taxes.some_tax.some_param"
-        :param data: Data loaded from a YAML file.
-        :param file_path: File the parameter was loaded from.
+        :param string name: Name of the parameter, e.g. "taxes.some_tax.some_param"
+        :param dict data: Data loaded from a YAML file.
+        :param string file_path: File the parameter was loaded from.
+        :param string documentation: Some documentation describing parameter usage and context.
 
 
         Instantiate a parameter without metadata:
@@ -258,9 +259,9 @@ class ParameterAtInstant(object):
 
     def __init__(self, name, instant_str, data = None, file_path = None, metadata = None):
         """
-            :param name: name of the parameter, e.g. "taxes.some_tax.some_param"
-            :param instant_str: Date of the value in the format `YYYY-MM-DD`.
-            :param data: Data, usually loaded from a YAML file.
+            :param string name: name of the parameter, e.g. "taxes.some_tax.some_param"
+            :param string instant_str: Date of the value in the format `YYYY-MM-DD`.
+            :param dict data: Data, usually loaded from a YAML file.
         """
         self.name = name
         self.instant_str = instant_str
@@ -320,6 +321,7 @@ class ParameterNode(object):
         :param string name: Name of the node, eg "taxes.some_tax".
         :param string directory_path: Directory containing YAML files describing the node.
         :param dict data: Object representing the parameter node. It usually has been extracted from a YAML file.
+        :param string documentation: Some documentation describing parameter node usage and context.
         :param string file_path: YAML file from which the `data` has been extracted from.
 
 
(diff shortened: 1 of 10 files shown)


tests/web_api/test_variables.py
```diff
@@ -154,3 +154,11 @@ def test_dated_variable_formulas_content():
 def test_variable_encoding():
     variable_response = subject.get('/variable/pension')
     assert_equal(variable_response.status_code, OK)
+
+
+def test_variable_documentation():
+    response = subject.get('/variable/housing_allowance')
+    variable = json.loads(response.data.decode('utf-8'))
+    assert_equal(variable['documentation'],
+        "\nThis allowance was introduced on the 1st of Jan 1980.\nIt needs the 'rent' value "
```
fpagnoux: The first line break is not relevant, and should be removed somehow. Same for the last one.

openfisca_web_api/loader/variables.py
```diff
@@ -78,6 +78,9 @@ def build_variable(variable, country_package_metadata, tax_benefit_system):
             ),
         }
 
+    if variable.documentation:
```
fpagnoux: Looking at OpenFisca France, a lot of doc string is added at the _formula_ level. This PR is not exposing this doc though the API. We should probably do that too :)

openfisca_core/parameters.py
```diff
@@ -120,19 +120,23 @@ def __init__(self, name, data, file_path = None):
         _validate_parameter(self, data, data_type = dict)
         self.description = None
         self.metadata = {}
+        self.documentation = None
```
fpagnoux: Should we also add this new attributes to parameter nodes? I'm not sure there is a good reason to do it only for leads.

tests/core/test_parameters.py
```diff
@@ -94,3 +94,15 @@ def test_parameter_node_metadata():
...
     assert_equal(parameter_2.description, 'Housing tax')
+
+
+def test_parameter_documentation():
+    parameter = tax_benefit_system.parameters.benefits.housing_allowance
+    assert_equal(parameter.documentation,
+        'A fraction of the rent. \nFrom the 1st of Dec 2016, the housing allowance no longer exists.\n')
+
+
+@raises(AttributeError)
```
fpagnoux: > Even if we chose not to give parameter nodes a documentation, I'm not sure about the relevance of such a test.

>  When reading this test, it feels like we are testing that Python sends an `AttributeError ` when calling an undefined attribute, which doesn't say much about OpenFisca itself

openfisca_core/parameters.py
sandcha: Do you see some use case for documentation on `ParameterNode`objects? 
As I see it, for now, we wouldn't use it so it would be over-engineering it.

fpagnoux: Yes, imagine something like: `rsa.plafonds.1_enfant`, `rsa.plafonds.2_enfants`, `rsa.plafonds.par_enfant_supp`.

In that case, it's probably more interesting to document the `rsa.plafonds` rather than the specific values.

In many case, parameters work by group and are not fully independent. In that case, the nodes may be the thing we want to document.

tests/web_api/test_variables.py
fpagnoux: See https://github.com/openfisca/openfisca-core/pull/714#discussion_r214879580 for an easy implem

tests/core/test_parameters.py
sandcha: Removed

> It was there to define the expected behaviour before coding but is irrelevant now.

openfisca_core/parameters.py
```diff
@@ -87,9 +87,10 @@ class Parameter(object):
     """
         A parameter of the legislation. Parameters can change over time.
 
-        :param name: name of the parameter, e.g. "taxes.some_tax.some_param"
-        :param data: Data loaded from a YAML file.
-        :param file_path: File the parameter was loaded from.
+        :param string name: Name of the parameter, e.g. "taxes.some_tax.some_param"
+        :param dict data: Data loaded from a YAML file.
+        :param string file_path: File the parameter was loaded from.
+        :param string documentation: Some documentation describing parameter usage and context.
```
fpagnoux: > Some documentation -> Documentation

openfisca_core/parameters.py
```diff
@@ -315,6 +320,7 @@ def __init__(self, name = "", directory_path = None, data = None, file_path = No
         :param string name: Name of the node, eg "taxes.some_tax".
         :param string directory_path: Directory containing YAML files describing the node.
         :param dict data: Object representing the parameter node. It usually has been extracted from a YAML file.
+        :param string documentation: Some documentation describing parameter node usage and context.
```
fpagnoux: > Some documentation -> Documentation

openfisca_core/parameters.py
```diff
@@ -358,8 +365,9 @@ def __init__(self, name = "", directory_path = None, data = None, file_path = No
 
                     if child_name == 'index':
                         data = _load_yaml_file(child_path)
-                        _validate_parameter(self, data, allowed_keys = ['metadata', 'description', 'reference'])
+                        _validate_parameter(self, data, allowed_keys = ['metadata', 'description', 'documentation', 'reference'])
                         self.description = data.get('description', None)
+                        self.documentation = data.get('documentation', None)
```
fpagnoux: `None` is already the default 2nd argument of get, specifying is redundant.

openfisca_core/parameters.py
```diff
@@ -378,8 +386,9 @@ def __init__(self, name = "", directory_path = None, data = None, file_path = No
         else:
             self.file_path = file_path
             _validate_parameter(self, data, data_type = dict, allowed_keys = self._allowed_keys)
-            # We allow to set a reference and a description for a node.
+            # We allow to set a reference, a description and a documentation for a node.
             self.description = data.get('description', None)
+            self.documentation = data.get('documentation', None)
```
fpagnoux: Unnecessary 2nd argument

openfisca_core/parameters.py
```diff
@@ -378,8 +386,9 @@ def __init__(self, name = "", directory_path = None, data = None, file_path = No
         else:
             self.file_path = file_path
             _validate_parameter(self, data, data_type = dict, allowed_keys = self._allowed_keys)
-            # We allow to set a reference and a description for a node.
```
fpagnoux: This comment seems outdated (we don't set a reference around it). I'd just remove it 🙂

openfisca_web_api/loader/variables.py
```diff
@@ -78,6 +82,9 @@ def build_variable(variable, country_package_metadata, tax_benefit_system):
...
```
fpagnoux: The way we handle unspecified documentation in the Web API is currently inconsistent:
- `documentation: null for parameters`
- `documentation: "" for formulas`
- no `documentation` key for variables

I'd be in favor of generalizing the 3rd option, as it makes the JSON more readable, lighter, and doesn't change anything for JS clients.

tests/core/test_parameters.py
```diff
@@ -94,3 +94,9 @@ def test_parameter_node_metadata():
 
     parameter_2 = tax_benefit_system.parameters.taxes.housing_tax
     assert_equal(parameter_2.description, 'Housing tax')
+
+
+def test_parameter_documentation():
+    parameter = tax_benefit_system.parameters.benefits.housing_allowance
+    assert_equal(parameter.documentation,
+        'A fraction of the rent. \nFrom the 1st of Dec 2016, the housing allowance no longer exists.\n')
```
fpagnoux: > This last line break is not really relevant. Cleaning the trailing line breaks with a `strip` in parameters parsing would be nice.

tests/web_api/test_variables.py
```diff
@@ -154,3 +154,14 @@ def test_dated_variable_formulas_content():
...
     variable_response = subject.get('/variable/pension')
     assert_equal(variable_response.status_code, OK)
+
+
+def test_variable_documentation():
+    response = subject.get('/variable/housing_allowance')
+    variable = json.loads(response.data.decode('utf-8'))
+    assert_equal(variable['documentation'],
+        "This allowance was introduced on the 1st of Jan 1980.\nIt needs the 'rent' value "
+        + "(same month) but doesn't care about the 'housing_occupancy_status'.")
```
fpagnoux: The `+` is not necessary, Python automatically concatenates strings.

tests/core/test_parameters.py
sandcha: Done for web API (not for python API).

openfisca_web_api/loader/variables.py
sandcha: Fixed for Web API : a documentation attribute is exposed only when it exists and its value is cleaned from start and end spaces.

> On Python API : the documentation attribute value is not modified at all.
For variable formulas, there is no specific `documentation` attribute as the user can already get  formula docstring with `formula.__doc__`.

tests/web_api/test_parameters.py
```diff
@@ -68,7 +68,11 @@ def test_parameter_node():
     assert_equal(response.status_code, OK)
     parameter = json.loads(response.data)
     assert_equal(sorted(list(parameter.keys())), ['description', 'documentation', 'id', 'metadata', 'source', 'subparams'])
-    assert_equal(parameter['documentation'], 'This (optional) file defines metadata for the node parameters.benefits in the parameter tree.')
+    assert_equal(parameter['documentation'],
```
fpagnoux: > For test robustness sake, I'd just check that `parameter['documentation']` contains a key expression, like "Government support for the citizens and residents".

sandcha:
@benjello Including one attribute of primitive type for all the variables won't have significant effect on tax benefit system instantiation. For now, documentation attributes will mainly gather existing comments and have very short contents.

PR #719, opened 2018-08-30
Experimental computation of dependencies based on AST parsing

Morendil:
#### New features

- The `/variable/:name` endpoint in Web API now provides a list of variables depended on
  - This is on a per-formula basis and the field is named `dependencies`

#### Status

Experimental; feedback sought on the feature and its possible extensions (see below). For now, consider this PR a prompt for a discussion of whether this approach would be valuable, to whom and under what conditions.

#### Possible improvements
- [ ] Support reverse dependency computation
- [ ] Provide support for static circular dependency detection
- [ ] Handle the tough case - dependencies via extracted (non-formula) methods
- [x] Provide a CLI interface to output e.g. Graphviz to dump the entire graph
- [ ] Parse and represent period shifts (e.g. `dependencies: [['income','Y-2']]`)
- [ ] Provide some error handling ?
- [ ] Replace the code in legislation explorer that regexes formula source code to link to dependencies
- [ ] be able to answer "what link is there between `revenu_disponible` and `csg_imposable_salaire`"


Morendil:
Closing this for now, as it didn't find a use in the `/trace` endpoint as I had anticipated at some point, and we do not have an immediate use case for it (such as a well-thought-out idea for visualizing the dependency graph).

PR #723, opened 2018-09-04
Small fixes in /spec route

### LINKED ISSUE #555: Use legit JSON types in the Open API specification
As a client of the OpenFisca web API,
I can get a proper JSON schema, with legit JSON types, of the request to send,
So that I can understand it and use it with standard tools

---------

`Integer` is not a valid JSON schema.

```JSON
        "af_age_aine": {
          "additionalProperties": {
            "type": "Integer"
          }, 
          "type": "object"
        }, 
```

-------------

Implementation suggestion:
 
- Pursue the work done on  #551

fpagnoux:
Fixes #555 

- Fix small issues in the `/spec` route of the Web API
  - Use proper JSON Schema type to describe input types
  - Fix property name in the description of `/parameters` and `/variables`


openfisca_core/variables.py
```diff
@@ -27,36 +27,42 @@
     bool: {
         'dtype': np.bool,
         'default': False,
-        'json_type': 'Boolean',
+        'json_type': 'boolean',
+        'formatted_value_type': 'Boolean',
```
fpagnoux: > The legislation explorer relies and these values, so I kept them. Swithing to something a little more standard would be probably better but this would introduce a breaking change, and would not add much value

Anna-Livia: > would `python_type` make more sense ? I explains why we display this in the legexp.

fpagnoux: I'd expect something called `python_type` to be `'float'`, `'int'` or `'bool'`.

sandcha:
Rebase en cours.

Issue #693, opened 2018-07-23
Deprecate Python 2.x


maukoquiroga:
As per python/devguide#344, Python 2 will be sunset come Jan 1st, 2020.

Additionally, a number of projects used in conjunction with OpenFisca [have pledged to stop supporting Python 2 soon](http://python3statement.org/) :

- pandas
- Matplotlib
- NumPy
- Jupyter notebook

Notably, NumPy's Python 2 maintenance mode will start come 2019.

Let's open the discussion and tackle this effort together.

To-do:

- [x] Change CircleCI configs
  - [x] Core uses 3.7
- [x] Change READMEs
- [x] Edit doc
- [x] Shared in Slack #general (YouTube, future, ...)
- [x] Tweeted
- [x] Sent by email

---------
_EDIT by @fpagnoux_:

Todo on January 1st:
- [ ] Remove all the `from __future__`, `from builtins` and `from future` imports

benjello:
At IPP, we will be happy to participate to the big switch, but we will be ready to do that around next december ...

fpagnoux:
Shall we announce the end of Python 2 compatibility for January 1st 2019 then? 

It seems a reasonable amount time for re-users to adapt their code?

benjello:
@fpagnoux : I am not against having a strong deadline. 

The best way to keep every contributors happy to reach it might be helping them to go through the necessary steps to transition to Python 3.

One point thought, using Python 2 until jan 2019 should be possible. And so the best way is having code supporting both Python until then.

Since @Anna-Livia and you went through the process, could you list somewhere if it is not to cumbersome the necessary steps to go through and I will definitively try to use that info to update IPP codebase and other packages like openfisca-survey-manager and openfisca-france-indirect-taxation (wich is BTW laggard in core too)

guillett:
@jmdallais, @Lashchuk , @ThibaultCCMSA , @SChiusa , @JenniferTelep , @LacombeEric, @frtomas , @mtifarine , @marchand-laetitia-msa, would you be able to move PNDS servers to Python 3 before early 2019 ?

jmdallais:
@guillett thanks for sharing.
@tchabaud, could you answer to @guillett please ? Strong prerequisites or constraints for us ? 
Give me a call tomorrow please.

fpagnoux:
> One point thought, using Python 2 until jan 2019 should be possible. And so the best way is having code supporting both Python until then.

[...]

PR #727, opened 2018-09-20
Auto-fix the code formatting

maukoquiroga:
Child of openfisca/openfisca-france#372
Connected to openfisca/openfisca-france#1114

#### Technical changes

- Add a style formatter that follows community code conventions
- Auto-fix code formatting

What do you folks think?

PR #728, opened 2018-09-26
Use Python 3.7

fpagnoux:
Connected to #693 

As suggested by @tchabaud, Python 3.7 is the new stable version of Python 3. Let's use it :)


fpagnoux:
Changes included in #727.

PR #734, opened 2018-10-03
Correct style ignore: W503 rather than W504

maukoquiroga:
Ignore W503 to enforce Knuth's style.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 24.5.3 [#734](https://github.com/openfisca/openfisca-core/pull/734)
+
+- Ignore W503 to enforce Knuth's style (W504)
+- Fix failing entities test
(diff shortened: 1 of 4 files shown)


fpagnoux:
Any example of what this style enforces?

sandcha:
@fpagnoux [This post](https://github.com/OCA/maintainer-quality-tools/issues/545#issue-316882979) gives an example.

Morendil:
I've just merged a corresponding change in setup.cfg of openfisca-france, but including both W504 & W503 - should it be W503 only ?

maukoquiroga:
@fpagnoux @Morendil It should be W504 only. 

It is not yet implemented in `pycodestyle`, but due to the next release.

It would enforce the style once we upgrade version.

maukoquiroga:
> Any example of what this style enforces?

#735

PR #735, opened 2018-10-06
Apply PEP8 W504 (Knuth's Style)

maukoquiroga:
#### Technical changes

- Apply PEP8 W504: break line before binary operator.

### DIFF
--- a/openfisca_core/periods.py
+++ b/openfisca_core/periods.py
@@ -403,8 +403,12 @@ class Period(tuple):
             return self.__class__((
                 'month',
                 intersection_start,
-                ((intersection_stop.year - intersection_start.year) * 12 + intersection_stop.month -
-                    intersection_start.month + 1),
+                (
+                    (intersection_stop.year - intersection_start.year) * 12
+                    + intersection_stop.month
+                    - intersection_start.month
+                    + 1
+                    ),
                 ))
         return self.__class__((
             'day',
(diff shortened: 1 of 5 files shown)


openfisca_core/periods.py
```diff
@@ -403,8 +403,13 @@ def intersection(self, start, stop):
             return self.__class__((
                 'month',
                 intersection_start,
-                ((intersection_stop.year - intersection_start.year) * 12 + intersection_stop.month -
-                    intersection_start.month + 1),
+                (
+                    (intersection_stop.year - intersection_start.year)
+                    * 12
```
fpagnoux: Not a fan of this `* 12` alone, I find it less easy to understand than `(intersection_stop.year - intersection_start.year) * 12`. 
Following @Morendil suggestion of suggesting a commit rather than requesting changes, I'm committing a small change, let me know if the new style doesn't work for you :)

maukoquiroga:
@fpagnoux They're enforced in the latest version of `pycodestyle`, so should be enforced as soon as flake8 and autopep8 integrate it (I've been testing with the master branches and works like a charm).

It's just a matter of time.

fpagnoux:
> @fpagnoux They're enforced in the latest version of pycodestyle, so should be enforced as soon as flake8 and autopep8 integrate it (I've been testing with the master branches and works like a charm).

Ok, thanks for the details 🙂 !

Issue #701, opened 2018-07-31
Add parameter values used in computations in tracer output


guillett:
Hi there!

I really enjoy OpenFisca, but here is a feature I would appreciate.

I investigate OpenFisca computations with http://localhost:9000/trace?situationId=588752fc3d4bb9bf7a656922

And I would love to have parameter values, in addition to variables values.

## Context

I identify more as a:

- Developer _(I create tools that use the existing OpenFisca code)_.

sandcha:
Today's `/trace` answer for country-template [single.json](https://github.com/openfisca/country-template/blob/8c2fc77e55da814cab7732c479356f9a6bafce22/openfisca_country_template/situation_examples/single.json) situation example is:
```
{
  "entitiesDescription": {
    "households": [
      "_"
    ], 
    "persons": [
      "Alicia"
    ]
  }, 
  "requestedCalculations": [
    "disposable_income<2017-01>", 
    "birth<2017-01>"
  ], 
  "trace": {
    "age<2017-01>": {
      "dependencies": [
        "birth<2017-01>"
      ], 
      "value": [
        47
      ]
    }, 
    "basic_income<2017-01>": {
      "dependencies": [
        "age<2017-01>"
      ], 
      "value": [
        600.0
      ]
    }, 
    "birth<2017-01>": {
      "dependencies": [], 
      "value": [
        "Thu, 01 Jan 1970 00:00:00 GMT"
      ]
    }, 
    "disposable_income<2017-01>": {
      "dependencies": [
        "salary<2017-01>", 
        "basic_income<2017-01>", 
        "income_tax<2017-01>", 
        "social_security_contribution<2017-01>"
      ], 
      "value": [
        600.0
      ]
    }, 
    "income_tax<2017-01>": {
      "dependencies": [
        "salary<2017-01>"
      ], 
      "value": [
        0.0
      ]
    }, 
    "salary<2017-01>": {
      "dependencies": [], 
      "value": [
        0.0
      ]
    }, 
    "social_security_contribution<2017-01>": {
      "dependencies": [
        "salary<2017-01>"
      ], 
      "value": [
        0.0
      ]
    }
  }
}
```
Used command: `curl -X POST http://localhost:5000/trace -H 'Content-Type: application/json' -d @single.json`

guillett:
[...]

PR #737, opened 2018-10-09
Fix type value for string variables

sandcha:
Relates to #701 
 
Bug detected by sending `https://fr.openfisca.org/legislation/swagger` example for `/trace` endpoint to local web api. The web api answer was:
`{"error":"Internal server error: Object of type bytes is not JSON serializable"}`

---

#### Technical changes

- Fix an encoding bug of variables with value type str
- Details:
  * `str` corresponds to `unicode` in Python 2.7 and `bytes` in Python 3.7
  * Set encoding with `numpy.unicode_` to dynamically cast to `unicode` both in Python 2.7 and Python 3.7

---

*Note:*

A numpy array containing values for a variable with `value_type = str`, automatically converts the values to:
* `numpy.str_` in Python 2.7
* `numpy.bytes_` in Python 3.7

In this PR, we update such variables `dtype` in `openfisca_core/variables.py` and this dtype gives:
*  `numpy.bytes_` for `|S` dtype value
* `numpy.unicode_` for `|U` or `unicode_` dtype value

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 
-### 24.5.3 [#734](https://github.com/openfisca/openfisca-core/pull/734)
+### 24.5.3 [#737](https://github.com/openfisca/openfisca-core/pull/737)
+
+- Fix an encoding bug of variables with value type str
+- Details:
+  * `str` corresponds to `unicode` in Python 2.7 and `bytes` in Python 3.7
+  * Set encoding with `numpy.unicode_` to dynamically cast to `unicode` both in Python 2.7 and Python 3.7
+
+### 24.5.2 [#734](https://github.com/openfisca/openfisca-core/pull/734)
 
 - Ignore W503 to enforce Knuth's style (W504)
 - Fix failing entities test
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,12 @@
 # Changelog
 
-## 24.5.3 [#734](https://github.com/openfisca/openfisca-core/pull/734)
+### 24.5.3 [#734](https://github.com/openfisca/openfisca-core/pull/734)
 
 - Ignore W503 to enforce Knuth's style (W504)
 - Fix failing entities test
   - Household description was outdated
 
-## 24.5.1 [#732](https://github.com/openfisca/openfisca-core/pull/732)
+### 24.5.1 [#732](https://github.com/openfisca/openfisca-core/pull/732)
 
 - Further adopt simplified simulation initialisation
   - See [#729](https://github.com/openfisca/openfisca-core/pull/729)
@@ -58,7 +58,7 @@ simulation.set_input('salary', '2018-08', [4000])
   - Use proper JSON Schema type to describe input types
   - Fix property name in the description of `/parameters` and `/variables`
 
-### 24.3.0 [#714](https://github.com/openfisca/openfisca-core/pull/714)
+## 24.3.0 [#714](https://github.com/openfisca/openfisca-core/pull/714)
 
 - Introduce the `/entities` endpoint for the Web API.
   - Expose information about the country package's entities, and their roles.
@@ -99,7 +99,7 @@ from openfisca_core.tools.simulation_dumper import restore_simulation
 simulation = restore_simulation('/path/to/directory', tax_benefit_system)
 ```
 
-### 24.1.0 [#713](https://github.com/openfisca/openfisca-core/pull/713)
+## 24.1.0 [#713](https://github.com/openfisca/openfisca-core/pull/713)
 
 - Enhance navigation within the Openfisca Web API.
 - Provides a direct link to individual parameters and variables from the `/parameters` and `/variables` routes.
--- a/openfisca_core/variables.py
+++ b/openfisca_core/variables.py
@@ -167,7 +167,7 @@ class Variable(object):
         if self.value_type == str:
             self.max_length = self.set(attr, 'max_length', allowed_type = int)
             if self.max_length:
-                self.dtype = '|S{}'.format(self.max_length)
+                self.dtype = '|U{}'.format(self.max_length)
         if self.value_type == Enum:
             self.default_value = self.set(attr, 'default_value', allowed_type = self.possible_values, required = True)
         else:
--- a/openfisca_core/variables.py
+++ b/openfisca_core/variables.py
@@ -162,12 +162,11 @@ class Variable(object):
         self.value_type = self.set(attr, 'value_type', required = True, allowed_values = VALUE_TYPES.keys())
         self.dtype = VALUE_TYPES[self.value_type]['dtype']
         self.json_type = VALUE_TYPES[self.value_type]['json_type']
-        if self.value_type == Enum:
-            self.possible_values = self.set(attr, 'possible_values', required = True, setter = self.set_possible_values)
         if self.value_type == str:
             self.max_length = self.set(attr, 'max_length', allowed_type = int)
             self.dtype = '|U'
         if self.value_type == Enum:
+            self.possible_values = self.set(attr, 'possible_values', required = True, setter = self.set_possible_values)
             self.default_value = self.set(attr, 'default_value', allowed_type = self.possible_values, required = True)
         else:
             self.default_value = self.set(attr, 'default_value', allowed_type = self.value_type, default = VALUE_TYPES[self.value_type].get('default'))
--- a/openfisca_core/variables.py
+++ b/openfisca_core/variables.py
@@ -164,7 +164,7 @@ class Variable(object):
         self.json_type = VALUE_TYPES[self.value_type]['json_type']
         if self.value_type == str:
             self.max_length = self.set(attr, 'max_length', allowed_type = int)
-            self.dtype = '|U'
+            self.dtype = 'unicode_'
         if self.value_type == Enum:
             self.possible_values = self.set(attr, 'possible_values', required = True, setter = self.set_possible_values)
             self.default_value = self.set(attr, 'default_value', allowed_type = self.possible_values, required = True)
--- a/openfisca_core/variables.py
+++ b/openfisca_core/variables.py
@@ -166,8 +166,7 @@ class Variable(object):
             self.possible_values = self.set(attr, 'possible_values', required = True, setter = self.set_possible_values)
         if self.value_type == str:
             self.max_length = self.set(attr, 'max_length', allowed_type = int)
-            if self.max_length:
-                self.dtype = '|U{}'.format(self.max_length)
+            self.dtype = '|U'
         if self.value_type == Enum:
             self.default_value = self.set(attr, 'default_value', allowed_type = self.possible_values, required = True)
         else:
--- a/tests/core/test_variables.py
+++ b/tests/core/test_variables.py
@@ -2,7 +2,9 @@
 
 from __future__ import unicode_literals, print_function, division, absolute_import
 import datetime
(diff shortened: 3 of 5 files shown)


Morendil:
LOL at [this line](https://github.com/openfisca/openfisca-france/blob/10650449abe2a70709165ac0ae7379323c5d2b4a/openfisca_france/model/prelevements_obligatoires/prelevements_sociaux/contributions_sociales/versement_transport.py#L69).

Morendil:
Breaking in Python 2 with

`AssertionError: <type 'numpy.string_'> != <type 'numpy.unicode_'>`

Awwww. I suggest `foo.type in [numpy.str_, numpy.unicode_]` or something like that. All we care about is getting a string, doesn't matter [what kind of string](http://www.simonandschuster.com/books/Bring-Me-a-Rock!/Daniel-Miyares/9781481446020).

openfisca_core/variables.py
```diff
@@ -162,13 +162,11 @@ def __init__(self, baseline_variable = None):
         self.value_type = self.set(attr, 'value_type', required = True, allowed_values = VALUE_TYPES.keys())
         self.dtype = VALUE_TYPES[self.value_type]['dtype']
         self.json_type = VALUE_TYPES[self.value_type]['json_type']
-        if self.value_type == Enum:
-            self.possible_values = self.set(attr, 'possible_values', required = True, setter = self.set_possible_values)
         if self.value_type == str:
             self.max_length = self.set(attr, 'max_length', allowed_type = int)
-            if self.max_length:
-                self.dtype = '|S{}'.format(self.max_length)
```
sandcha: > @fpagnoux For a value exceeding `max_length`, this line reduced the string to `max_length`. If it was done on purpose, let me know as this behaviour is removed here.

CHANGELOG.md
```diff
@@ -1,12 +1,18 @@
 # Changelog
 
-## 24.5.3 [#734](https://github.com/openfisca/openfisca-core/pull/734)
+### 24.5.4 [#737](https://github.com/openfisca/openfisca-core/pull/737)
+
+- Fix a bug on encoding for OpenFisca variables of string type
+- Details:
+  * Set variable value encoding to `numpy.unicode_` in python 2.7 and python 3.7
```
maukoquiroga: I think an explanation of the situation would be very useful for other contributors.

tests/core/test_variables.py
```diff
@@ -474,3 +476,19 @@ class variable_with_strange_attr(Variable):
...
+        entity = Person
+        definition_period = MONTH
+        label = "String variable of specific max length"
+
+    tax_benefit_system.add_variable(variable__str_with_max)
+    month = '2018-01'
+
+    simulation = new_simulation(tax_benefit_system, month)
+    variable_value = simulation.calculate('variable__str_with_max', month)[0]
+    assert_equal(unicode_, type(variable_value))
```
maukoquiroga: Usually, asserts works like this:

```python
assert actual_value == desired_value
assert_equal(actual_value, desired_value)
```

maukoquiroga: I'll change it quickly to merge this PR

openfisca_core/variables.py
fpagnoux: I believe the point was to make sure a given variable would be limited to _n_ characters, to reduce the memory usage of the variable(typically, a zip code).

Whether this was a good idea or not can be discussed, but removing this along the way in a bug fix PR is problematic IMO. After this PR, the `max_length` variable attribute is totally useless, yet it hasn't been deprecated, and is still documented. This is not a stable situation.

fpagnoux:
I'm also under the impression that this PR make France calculations crash. 
When I try to run `nosetests tests/test_basics.py` in `france`, I get:
`TypeError: startswith first arg must be str or a tuple of str, not numpy.bytes_`

https://fr.openfisca.org/legislation/swagger works (both `/calculate` and `/trace`) with version 24.5.2 (before this PR). But I'm afraid it would crash if we upgraded Core.

I'd therefore recommend to revert this PR.

openfisca_core/variables.py
Morendil: We're reversing this, as it turns out to result in a regression undetected by existing unit tests.

As I recall, @sandcha and I were pairing at the time and we had a miscommunication, I assumed truncation to max_length would still be happening elsewhere as a result of line 168, I had not understood we were eliminating the behaviour.

The truncation behaviour, however, was (still is) untested and implicit. Fixing this properly entails at the very least documenting the truncation behaviour.

@sandcha suggests emitting a truncation warning at execution; I think that does not achieve the purpose of avoiding nasty surprises, because whoever is coding an input variable won't exert control on the actual inputs specified, and whoever actually specifies an overlong input is unlikely to see debug logs from the computation engine.

My preferred solution would be to rename max_length to something like truncate_to_length, to make the behaviour explicit; and a unit test for the truncation behaviour would be ideal.

sandcha: Well, I'm just in favour of giving the information to the user somewhere like in a unit test explaining the expected behaviour. A PR is coming on that 🙂

CHANGELOG.md
```diff
@@ -1,12 +1,19 @@
 # Changelog
 
-## 24.5.3 [#734](https://github.com/openfisca/openfisca-core/pull/734)
+### 24.5.3 [#737](https://github.com/openfisca/openfisca-core/pull/737)
+
+- Fix an encoding bug of variables with value type str
```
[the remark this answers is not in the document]
Morendil: Agreed, the fix went beyond what was strictly necessary.

I'm not sure the "thoughts and time" remark is quite justified though, we have systemic contributors to the incident (lack of unit tests or documentation covering this behaviour, Python 2/3 continuing hassles), and both @maukoquiroga and I paired with @sandcha independently so I don't see this as a lack of individual attention or rushing.

I'm not sure I understand your remark on Unicode, the Python 3 default encoding is UTF8 (and this carries over to Numpy via the "U" dtypes according to [the docs](https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.dtypes.html#string-dtype-note)) and while I admit to low expertise in string encodings my understanding is that this "variable-width" encoding does not necessarily use four bytes per character.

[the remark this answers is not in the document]
maukoquiroga: Hi @fpagnoux.

> Use unicode for all string variables

That's definitely better, although I didn't know "string" was a Python concept so that's why I didn't use it here.

> Deprecate max_length variable attribute

Didn't know we were deprecating it, nor that there was a truncation feature.

> Which I think would have deserve maybe a little more thoughts and time.

We were three to work on this specific issue, so more brains weren't available, and invested time was consequent.

Doing the math, we invested ballpark 4h * 3 = 12h of 75% of the team's brains, which is basically around ~10% of our sprint allocated time. I have to agree with @Morendil here:

> we have systemic contributors to the incident (lack of unit tests or documentation covering this behaviour, Python 2/3 continuing hassles)

However, I assume the responsibility as I reviewed, cleaned up, rebased, renamed,  fixed up and merged.

That being said, I don't know exactly what could I have done to avoid the situation.

openfisca_core/variables.py
maukoquiroga: > The truncation behaviour, however, was (still is) untested and implicit. Fixing this properly entails at the very least documenting the truncation behaviour.

👍 

> My preferred solution would be to rename max_length to something like truncate_to_length

👎 Attribute names of a model are preferably _nouns_. `variable.truncate_to_length` sounds like a `function`, because `truncate` is a `verb`.

maukoquiroga:
> I'd therefore recommend to revert this PR.

Fix #738 restores `max_length`, but does not reverts this PR's fix (it actually fixes the fix).

CHANGELOG.md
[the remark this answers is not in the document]
fpagnoux: What I meant by time was more _absolute time_ than _brain time_:

- A review is requested at D0, **2 days ago**. A comment (where I'm mentioned) clearly seem to indicate that the PR openers don't fully understand the purpose of the code they are removing.
- At D0+1, the PR is merged, with this comment still unanswered. 
- Later that day, I check the PR, following the mention. I'm surprised by the wide scope of the PR, and then realize France broke. 
- At D0+2, we revert everything

The pity is that we were very close from avoiding the incident. Delaying the merge a day, until @sandcha's comment was answered, would have been enough😬 .

Having said that, misevaluating the scope of a PR happens. Now we can just:
- extend unit tests to making it less likely
- keep in mind than reverting/unpublishing is the best option it done quickly when it happens 🙂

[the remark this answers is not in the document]
fpagnoux: > I'm not sure I understand your remark on Unicode, the Python 3 default encoding is UTF8 (and this carries over to Numpy via the "U" dtypes according to the docs) and while I admit to low expertise in string encodings my understanding is that this "variable-width" encoding does not necessarily use four bytes per character.

To illustrate my remark:

```py
import numpy as np

postal_codes = ['75012', '75015', '75007']
bytes_array = np.asarray(postal_codes, dtype=np.bytes_)
unicode_array = np.asarray(postal_codes, dtype=np.unicode_)

print(unicode_array.nbytes)
>>> 60  # 3 items x  20 bytes

print(bytes_array.nbytes)
>>> 15 # 3 items x 5 bytes
```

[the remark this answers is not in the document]
fpagnoux: We should probably open a separate issues to discuss all the issues that we have identified.
- Should all string variables be unicode, for convenience?
- Should all string variables be bytes, for memory optimisation?
- Should we allow both? Sounds nice, but is there really a demand for unicode variables?
- Is the "max_length" field really useful? Should it throw errors instead of silently shortening user inputs?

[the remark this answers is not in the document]
Morendil: Ah, the plot thickens. This is apparently not documented but [here's an interesting discussion](https://github.com/PyTables/PyTables/issues/499) pointing out that Numpy uses UTF-32 internally, rather than UTF-8 as I'd inferred from the Numpy docs.

The thread points to an additional option:
```
>>> object_array = np.asarray(postal_codes, dtype="O")
>>> object_array.nbytes
24
>>> type(object_array[0])
<class 'str'>
```

Still some overhead, but limited.

> Should all string variables be unicode, for convenience?

The alternative is that as soon as someone tries to use `value_type = str` for something - anything - that can conceivably contain even one accented character, they're getting set up for a [nasty surprise](https://github.com/betagouv/beta.gouv.fr/pull/1688). I think that goes beyond convenience, it's an egregious violation of the principle of least surprise.

If it's important for performance reasons, then one solution is to encode strings internally as byte arrays before storing them into Numpy arrays and decode them back into proper strings whenever their values is needed. To do that, we'd need to hide Numpy arrays (an implementation detail) behind the Variable/Holder (I'm not sure which) interface protocol. This is possibly a good idea (on information hiding principles) irrespective of the current issue, even if it would admittedly involve a lot of busywork.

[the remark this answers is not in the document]
Morendil: > Delaying the merge a day, until @sandcha's comment was answered, would have been enough 😬.

Can't disagree, but we are analysing this now with the benefit of hindsight. In [Allspaw's words](https://codeascraft.com/2012/05/22/blameless-postmortems/) "the action made sense to the person at the time they took it, because if it hadn’t made sense to them at the time, **they wouldn’t have taken the action in the first place**".

There are any number of additional counterfactual scenarios where merging the PR wouldn't have resulted in a meltdown, including "if we'd had more tests on string-valued variables" and "if we'd limited the fix to the least necessary change" or "if the people pairing on the fix had been more sensitive about memory usage".

PR #739, opened 2018-10-12
Fix str variable test and default_value encoding

sandcha:
Relates to #737  

#### Technical changes

- Make str variable `default_value` of unicode type (Python 2 and Python 3 compatible type)
- Details:
  * Introduce test check on str variable `max_length` effect
  * Explicit str variables data type for Python 2 and Python 3

Note:
Core tests missed checks on `default_value` encoding and `max_length` effect for variables of `value_type = str`. This context was detected by web api `/trace` tests and openfisca-france and led to #737, #738 and this PR.

---

Superseded by #741

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+### 25.5.5 [#739](https://github.com/openfisca/openfisca-core/pull/739)
+
+- Make str variable `default_value` of unicode type for Python 2 backward compatibility
+- Details:
+  * Introduce test check on str variable `max_length` effect
+  * Explicit str variables data type for Python 2 and Python 3
+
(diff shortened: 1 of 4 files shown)


Morendil:
I'm "approve" on this insofar as it adds a new test and a fix that makes it green. And I'm still worried about this:
https://github.com/openfisca/openfisca-core/blob/5ee5f1a8179362a78b2ba7e5beee154d72e53406/openfisca_core/variables.py#L170

…as the Numpy docs [explicitly warn against](https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.dtypes.html#string-dtype-note) the "S" type specifier:

> 'S', 'a' | zero-terminated bytes (not recommended).

We are definitely changing this line again sometime early in 2019, so it's worth asking ourselves, what will we change it to at that point ?

PR #743, opened 2018-10-15
Allow empty parameter node files

Morendil:
#### Technical changes

- Nodes in the parameter hierarchy are now allowed to contain no data at all

Rationale: previously, 11 files in the France parameter hierarchy contained no data and only a single element of metadata, the `reference` field, set to the arbitrary and non-informative value `openfisca`. These were removed by openfisca/openfisca-france#1162 but this breaks parameter validation. Since `reference: openfisca` has zero informational content, I infer the unstated requirement that parameter nodes are allowed to contain no data at all. This introduces a test and a fix for that condition.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 24.5.6 [#743](https://github.com/openfisca/openfisca-core/pull/743)
+
+- When there is an empty `index.yaml` in the parameters, ignore it instead of raising an error.
+
 ## 24.5.5 [#742](https://github.com/openfisca/openfisca-core/pull/742)
 
 - Fix the internal server error that appeared for the  `/trace` and (less frequently) `/calculate` route of the Web API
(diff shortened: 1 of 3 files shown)


sandcha:
What's the difference between an empty `index.yaml` and no `index.yaml` at all (as it is an optional file)? 🤔

Morendil:
None in terms of informational content. However, for me as a France contributor it is quite significant in pragmatic terms.  The test runner in Core does not provide the names of the files that fail parameter validation for this specific reason, so merging this PR would save me from having to seek 11 needles in a haystack of YAML files.

The alternative is to instrument parameter validation to raise an error in the case of an empty YAML file. I could do this as it's a more direct solution to my own problem; however it doesn't do anything for future contributors who might come across the same problem, hence my preference for the solution in this PR.

fpagnoux:
Ignoring empty YAML index files instead of raising an error seems reasonable.

As @Morendil asked for a speedy review, I'll avoid another round trip and write the Changelog.

Morendil:
Thanks ! (I still don't have merge rights on Core - can you merge too, or upgrade my rights ?)

Issue #664, opened 2018-05-18
Obtain web API examples from the provided country package


MattiSG:
The OpenAPI specification exposed on `/spec` is currently pulling parts of the data it exposes from the loaded country package (such as name, variables list, parameters list…).

However, all usage examples are [hardcoded](https://github.com/openfisca/openfisca-core/blob/3d5084a34cebb9274daa9d45717e0424f39f068e/openfisca_web_api_preview/openAPI.yml#L202-L224), which prevents the examples to work with other country packages than `france` (and will also prevent them from working when the `france` package is updated).

The examples should be dynamically created from the loaded country package.

fpagnoux:
Good point, thanks for reporting the issue. 

Adding that to the backlog, but any PR would be welcome :)

fpagnoux:
Concrete impact of this issues:

In http://nz.openfisca.org/swagger, the "Try it out" on the `/calculate` doesn't work with the provided example:

![image](https://user-images.githubusercontent.com/11834997/45742833-282ca500-bbfb-11e8-9085-82ffc69f5082.png)

fpagnoux:
Result of some investigation and discussion with @sandcha:
- Creating a valid simulation input for an arbitrary country package automatically is not trivial
- Without any additional input from the country package maintainer, such an example is likely to be irrelevant (e.g. calculating the `age` with the `salary` as an input)
- It seems more relevant to let each country package define the relevant simulation inputs to be exposed as examples in the OpenAPi spec. The country template actually [already does it](https://github.com/openfisca/country-template/tree/master/openfisca_country_template/situation_examples).

fpagnoux:
## Proposition

Country package maintainers can define which examples they want to use in their OpenAPI spec the following way: https://github.com/openfisca/country-template/compare/define-openapi-ex

If no example has been defined:
    - An arbitrary parameter is used for the `/parameter` route example
    - An arbitrary variable is used for the `/variable` route example
    - No example is shown for the `/calculate` and `/trace` routes. Instead, a text message such as the following is displayed:

![image](https://user-images.githubusercontent.com/11834997/47042257-cfe4b680-d158-11e8-888b-769524c68a2b.png)

What do you think 🙂 ? @openfisca/aotearoa @openfisca/core-contrib ?

PR #744, opened 2018-10-17
Use country package specific examples in the /spec route

fpagnoux:
Connected to #664 

Country package maintainers can define which examples they want to use in their OpenAPI spec the following way: https://github.com/openfisca/country-template/compare/define-openapi-ex

If no example has been defined:
    - An arbitrary parameter is used for the `/parameter` route example
    - An arbitrary variable is used for the `/variable` route example
    - No example is shown for the `/calculate` and `/trace` routes. Instead, a text message such as the following is displayed:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,15 @@
 # Changelog
 
-## 24.5.6 [#743](https://github.com/openfisca/openfisca-core/pull/743)
+## 24.6.0 [#744](https://github.com/openfisca/openfisca-core/pull/744)
+
+- Allow TaxBenefitSystem to define the examples to use in the `/spec` route.
+  - See [docs](http://openfisca.org/doc/openfisca-web-api/config-openapi.html).
+
+### 24.5.6 [#743](https://github.com/openfisca/openfisca-core/pull/743)
 
 - When there is an empty `index.yaml` in the parameters, ignore it instead of raising an error.
 
-## 24.5.5 [#742](https://github.com/openfisca/openfisca-core/pull/742)
+### 24.5.5 [#742](https://github.com/openfisca/openfisca-core/pull/742)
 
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,10 +1,15 @@
 # Changelog
 
-## 24.5.6 [#743](https://github.com/openfisca/openfisca-core/pull/743)
+## 24.6.0 [#744](https://github.com/openfisca/openfisca-core/pull/744)
```
Morendil: Ca ne devrait pas être une MAJOR plutôt ?

fpagnoux: What previous API call / usecase do you think would now break ? I didn't think of any, but maybe I missed something.

Morendil: So, I ssh'd into that build and looked for the `couple.json` file and it is indeed missing, which is the proximal cause of the tests failing.

Does this have to do with Pypi packaging and the JSON file maybe not being bundled up, does it have to be listed in a manifest file or something ?

fpagnoux: Exactly. And this is pretty hard to detect in the CI 🤕

fpagnoux: In the country-template, we avoid the issue by just packaging everything:
https://github.com/openfisca/country-template/blob/master/MANIFEST.in#L1

We may want to do the same in Core, but we should make sure it doesn't increase too much the size of the bundle.

maukoquiroga: I tried to do the same I did in France with core but yaml test fail.

I think we have to refactor openfisca-run-test first, or rewrite the test_yaml.py.

fpagnoux: How do they fail? 
We don't really want to package YAML tests in the wheel anyway (users don't need the tests)

maukoquiroga:
I’ll try to open a PR so we can take a look together

Mauko Quiroga
________________________________

PR #751, opened 2018-10-24
Minor fixes

benjello:

#### Technical changes

- Fix average_rate computation when division by zero are present
- Fix entity.count restoration for a simulation dump

### DIFF
--- a/openfisca_core/rates.py
+++ b/openfisca_core/rates.py
@@ -7,7 +7,7 @@ import numpy
 
 def average_rate(target = None, varying = None):
     # target: numerator, varying: denominator
-    return 1 - target / (varying * (varying != 0) + (varying == 0))
+    return 1 - target / varying
 
 
 def marginal_rate(target = None, varying = None, trim = None):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
-### 24.6.4 [#???](https://github.com/openfisca/openfisca-core/pull/???)
+### 24.6.5 [#751](https://github.com/openfisca/openfisca-core/pull/751)
+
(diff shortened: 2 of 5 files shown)


openfisca_core/rates.py
```diff
@@ -7,7 +7,7 @@
 
 def average_rate(target = None, varying = None):
     # target: numerator, varying: denominator
-    return 1 - target / (varying * (varying != 0) + (varying == 0))
+    return 1 - target / varying
```
Morendil: Where is average_rate used ? I can't find references to this function.

benjello: It is used in many packages like openfisca-matplotlib, notebooks etc. It is so generic that it is better to have it include in core. It is not worth to maintenance cost to have a specific package for that.

Morendil: If it's worth having in Core, it's probably worth having a unit test for it. No unit tests and no obvious callers means zero pressure on maintainers to keep it operational. :)

maukoquiroga: I can add it, but I do not know what it does, example:

```python
import numpy

from openfisca_core.rates import average_rate


def test_average_rate_when_varying_is_zero():
    '''Add description like: it should do this'''
    target = numpy.array([123])
    result = average_rate(target, varying = 0)
    assert numpy.isinf(result)


def test_average_rate_when_varying_is_not_zero():
    '''Add description like: it should do that'''
    target = numpy.array([1, 2, 3])
    result = average_rate(target, varying = 2)
    expected = numpy.array([.5, 0, -.5])
    numpy.testing.assert_equal(result, expected)
```

maukoquiroga: @Morendil see #760

Issue #492, opened 2017-04-20
Merge DummyCountry and boilerplate


MattiSG:
As a _new [Producer](https://github.com/openfisca/openfisca-core/wiki/Taxonomie-des-utilisateurs)_,
I can start coding my tax and benefits system with an always up-to-date boilerplate,
So that I don't learn an outdated syntax

- - - -

Merge https://github.com/openfisca/openfisca-dummy-country and https://github.com/openfisca/country-template/ and use the result as a test base for openfisca-core.

MattiSG:
Import and editorialise the DummyCountry into the Boilerplate, demonstrating standard and advanced uses, adding comments linking to the doc, and suggesting future-proof architecture (i.e. using folders and showing that several files can be used).

MattiSG:
![img_4864](https://cloud.githubusercontent.com/assets/222463/25715686/b0a7695a-30fc-11e7-88c8-3eaf15b3a046.jpg)

MattiSG:
Based on `openfisca-france`:

- [ ] Add `.gitignore` with proper defaults (`.DS_Store`, IDEs, `pyc`…).
- [ ] Add `Makefile` with proper `test` (and `lint`?).
- [ ] Add `CONTRIBUTING.md`.
- [ ] Add `LICENSE.AGPL.txt`.
- [ ] Add `.github/ISSUE_TEMPLATE` (from `openfisca-core`).
- [ ] Add `.github/PULL_REQUEST_TEMPLATE` (translate from `openfisca-france`).
- [ ] Add `README.md`: be clear about that repo does, but expect forks not to modify it properly.
- [ ] Add `CHANGELOG.md`: use the format of `france`.
- [ ] Add `MANIFEST.in`.
- [ ] Add `.circle.yml` and set up sane testing defaults.
- [ ] Add `setup.py`.
- [ ] Move `tests` to `openfisca_country_template/tests` and update the version check code.
- [ ] Move `param.xml` to `parameters/*.xml`.
- [ ] Move `entities.py` to `entities/__init__.py`.
- [ ] Move `scenarios.py` to `entities/relations.py`.
- [ ] Remove `assets`.
- [ ] Remove `conf`.
- [ ] Remove `decompositions`.
- [ ] Remove `extensions`.
- [ ] Remove `i18n`.

PR #753, opened 2018-10-30
Always allow documentation property for parameter nodes

fpagnoux:
- Fixup #717: Always allow `documentation` properties for parameter nodes
  - Due to a bug, this field  was accepted in an `index.yaml` file, but not in a Parameter node located in a single YAML file

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,8 +89,12 @@ jobs:
             - /tmp/venv/openfisca_core
 
       - run:
-          name: Run tests
+          name: Run Core tests
           command: make test
+
+      - run:
+          name: Run Country Template tests
+          command: openfisca-run-test /tmp/venv/openfisca_core/lib/python*/site-packages/openfisca_country_template/tests/
 
   deploy_python3:
     docker:


fpagnoux:
See issue if I [add a `documentation` to a node in the country template](https://github.com/openfisca/country-template/pull/64/files), it [breaks]( https://circleci.com/gh/openfisca/country-template/643)

fpagnoux:
This bug fix has been opened for 7 days, anyone @openfisca/core-contrib ?

Morendil:
I'd approve, but shouldn't we see the tests pass first? Maybe merge #761 first, rebase this one afterwards (will trigger a rebuild), go back to green.

maukoquiroga:
I added a test and rebased.

I had to add a fix as the test failed, please @fpagnoux check if the behaviour is the expected one.

openfisca_core/parameters.py
```diff
@@ -255,7 +259,7 @@ class ParameterAtInstant(object):
 
     _allowed_value_data_types = [int, float, bool, type(None)]
     # 'unit' and 'reference' are only listed here for backward compatibility
-    _allowed_keys = set(['value', 'metadata', 'unit', 'reference'])
+    _allowed_keys = COMMON_KEYS.union({'value'})
```
maukoquiroga: @fpagnoux Is this the expected behaviour?

fpagnoux: No, good point, I don't think we should have `description` and `documentation` for dated values.

fpagnoux:
> I think there's value to render tests independent to another libraries like the country template.

We actually went the opposite directions on purpose 😅: https://github.com/openfisca/openfisca-core/issues/492.

PR #764, opened 2018-11-09
Simplify cycle detection

Morendil:
#### Technical changes

The previous behaviour when detecting "cycles" (which can in fact be spirals) was to re-raise the first `CycleError` encountered until reaching the first computation of a variable that was triggered with a value other than None for `max_nb_cycles`. This PR simplifies this behaviour so that the first formula that results in a "cycle" returns the variable's default value. This removes the inconsistencies observed in production by Mes Aides, but is *definitely* not proven to be a complete solution.

The intent of this PR is to solely allow Mes Aides to move back to the `master` branch of OpenFisca, pending further investigation of the problem.


maukoquiroga:
@benjello Would be nice were you able to test this with some population.

fpagnoux:
> The intent of this PR is to solely allow Mes Aides to move back to the master branch of OpenFisca, pending further investigation of the problem.

I think we should make sure the Core codebase stays in a consistent state. By reading the description of this PR, I'm under the impression that we ditch the `max_nb_cycles ` mechanism. However, we don't remove all parts of the code referencing it. This, and the presence of a non-executed tests, is a strong smell of an inconsistent state of the codebase.

> This removes the inconsistencies observed in production by Mes Aides

And it brings new ones:

```
class variable3(Variable):
    value_type = int
    entity = Person
    definition_period = MONTH

    def formula(person, period):
        return 2 + person('variable4', period.last_month)


class variable4(Variable):
    value_type = int
    entity = Person
    definition_period = MONTH

    def formula(person, period):
        return 3 + person('variable3', period)

def test():
    simulation = tax_benefit_system.new_scenario().init_from_attributes(
        period = reference_period,
        ).new_simulation()
    simulation2 = tax_benefit_system.new_scenario().init_from_attributes(
        period = reference_period,
        ).new_simulation()

    sim1_var_3 = simulation.calculate('variable3', period = reference_period)
    sim1_var_4 = simulation.calculate('variable4', period = reference_period)

    sim2_var_4 = simulation2.calculate('variable4', period = reference_period)
    sim2_var_3 = simulation2.calculate('variable3', period = reference_period)

    assert sim1_var_3 == sim2_var_3  # Breaks as 5 != 2
    assert sim1_var_4 == sim2_var_4  # Breaks as 8 != 5
```
  
The order of the calculations has a result on the outcome.

Morendil:
Closing this, following planning meeting, to take a step back and solve the broader problem, possibly by other means.

PR #765, opened 2018-11-09
Add parameters in trace

### LINKED ISSUE #701: Add parameter values used in computations in tracer output

sandcha:
Fixes #701 

#### New features

- Add parameters in `/trace` web api endpoint

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,28 @@
 # Changelog
 
+## 24.8.0 [#765](https://github.com/openfisca/openfisca-core/pull/765)
+
+- Add called parameters to Web API `/trace` endpoint
+  - For a calculated variable, add `parameters` item next to `dependencies` in `/trace` response
+  - Scale parameter example:
+    ```json
+    # calculated variable period = 2017-01
+    "parameters": {
+      "taxes.social_security_contribution<2017-01-01>": {
+          "0.0": 0.02,
+          "12400.0": 0.12,
+          "6000.0": 0.06
+      }
+    },
+    ```
+  - Simple parameter example:
+    ```json
+    # calculated variable period = 2017-01
+    "parameters": {
+      "taxes.income_tax_rate<2017-01-01>": 0.15
+    },
+    ```
+
 ### 24.7.0 [#756](https://github.com/openfisca/openfisca-core/pull/756)
 
 - Exposes `amount` scales through the Web API
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,22 @@
 
 - Add called parameters to Web API `/trace` endpoint
   - For a calculated variable, add `parameters` item next to `dependencies` in `/trace` response
-  - Scale parameter example:
-    ```json
-    # calculated variable period = 2017-01
+  - For example:
+  ```JSON
+    {
+      "income_tax<2017-01>": {
+        "dependencies": [
+          "salary<2017-01>"
+        ],
+        "parameters": {
+          "taxes.income_tax_rate<2017-01-01>": 0.15
+        },
+        "value": [150]
+      }
+    }
+  ```
+  - Scale parameters are also traced:
+  ```json
     "parameters": {
       "taxes.social_security_contribution<2017-01-01>": {
           "0.0": 0.02,
@@ -14,14 +27,7 @@
           "6000.0": 0.06
       }
     },
-    ```
-  - Simple parameter example:
-    ```json
-    # calculated variable period = 2017-01
-    "parameters": {
-      "taxes.income_tax_rate<2017-01-01>": 0.15
-    },
-    ```
+  ```
 
 ### 24.7.0 [#756](https://github.com/openfisca/openfisca-core/pull/756)
 
--- a/openfisca_core/parameters.py
+++ b/openfisca_core/parameters.py
@@ -466,14 +466,10 @@ class ParameterNodeAtInstant(object):
         self._instant_str = instant_str
         self._children = {}
 
-        # if trace:
-        self._original_children = {}
-
         for child_name, child in node.children.items():
             child_at_instant = child._get_at_instant(instant_str)
             if child_at_instant is not None:
                 self._children[child_name] = child_at_instant
-                self._original_children[child_name] = child
                 setattr(self, child_name, child_at_instant)
 
     def __getattr__(self, key):
--- a/openfisca_core/parameters.py
+++ b/openfisca_core/parameters.py
@@ -465,10 +465,15 @@ class ParameterNodeAtInstant(object):
         self._name = name
         self._instant_str = instant_str
         self._children = {}
+
+        # if trace:
+        self._original_children = {}
+
         for child_name, child in node.children.items():
             child_at_instant = child._get_at_instant(instant_str)
             if child_at_instant is not None:
                 self._children[child_name] = child_at_instant
+                self._original_children[child_name] = child
                 setattr(self, child_name, child_at_instant)
 
     def __getattr__(self, key):
--- a/openfisca_core/parameters.py
+++ b/openfisca_core/parameters.py
@@ -35,6 +35,7 @@ FILE_EXTENSIONS = {'.yaml', '.yml'}
 # 'unit' and 'reference' are only listed here for backward compatibility.
 #  It is now recommended to include them in metadata, until a common consensus emerges.
 COMMON_KEYS = {'description', 'metadata', 'unit', 'reference', 'documentation'}
+ALLOWED_PARAM_TYPES = (float, int, bool, type(None))
 
 
 def date_constructor(loader, node):
@@ -257,7 +258,6 @@ class ParameterAtInstant(object):
         A value of a parameter at a given instant.
     """
 
-    _allowed_value_data_types = [int, float, bool, type(None)]
     # 'unit' and 'reference' are only listed here for backward compatibility
     _allowed_keys = set(['value', 'metadata', 'unit', 'reference'])
 
@@ -273,7 +273,7 @@ class ParameterAtInstant(object):
         self.metadata = {}
 
         # Accept { 2015-01-01: 4000 }
-        if not isinstance(data, dict) and type(data) in self._allowed_value_data_types:
+        if not isinstance(data, dict) and isinstance(data, ALLOWED_PARAM_TYPES):
             self.value = data
             return
 
@@ -294,7 +294,7 @@ class ParameterAtInstant(object):
                 "Missing 'value' property for {}".format(self.name),
                 self.file_path
                 )
-        if type(value) not in self._allowed_value_data_types:
+        if not isinstance(value, ALLOWED_PARAM_TYPES):
             raise ParameterParsingError(
                 "Invalid value in {} : {}".format(self.name, value),
                 self.file_path
--- a/openfisca_core/simulations.py
+++ b/openfisca_core/simulations.py
@@ -11,7 +11,7 @@ import numpy as np
 
 from openfisca_core import periods
 from openfisca_core.commons import empty_clone, stringify_array, basestring_type, to_unicode
-from openfisca_core.tracers import Tracer
+from openfisca_core.tracers import Tracer, TracingParameterNodeAtInstant
 from openfisca_core.indexed_enums import Enum, EnumArray
 
 
@@ -256,11 +256,15 @@ class Simulation(object):
         formula = variable.get_formula(period)
         if formula is None:
             return None
-        parameters_at = self.tax_benefit_system.get_parameters_at_instant
 
         if self.trace:
-            if variable.name == 'income_tax':
-                self.tracer.record_calculation_parameter_access(parameters_at(period).taxes._original_children['income_tax_rate'], period)
+            parameters_at = lambda formula_period: TracingParameterNodeAtInstant(
+                self.tax_benefit_system.get_parameters_at_instant(formula_period),
+                self.tracer
+                )
+            
+        else:
+            parameters_at = self.tax_benefit_system.get_parameters_at_instant
 
         try:
             self._check_for_cycle(variable, period)
--- a/openfisca_core/simulations.py
+++ b/openfisca_core/simulations.py
@@ -257,6 +257,11 @@ class Simulation(object):
         if formula is None:
             return None
         parameters_at = self.tax_benefit_system.get_parameters_at_instant
+
+        if self.trace:
+            if variable.name == 'income_tax':
+                self.tracer.record_calculation_parameter_access(parameters_at(period).taxes._original_children['income_tax_rate'], period)
+
         try:
             self._check_for_cycle(variable, period)
             if formula.__code__.co_argcount == 2:
--- a/openfisca_core/simulations.py
+++ b/openfisca_core/simulations.py
@@ -248,6 +248,12 @@ class Simulation(object):
 
         return variable.calculate_output(self, variable_name, period)
 
+    def trace_parameters_at_instant(self, formula_period):
+        return TracingParameterNodeAtInstant(
+            self.tax_benefit_system.get_parameters_at_instant(formula_period),
+            self.tracer
+            )
+
     def _run_formula(self, variable, entity, period, extra_params, max_nb_cycles):
         """
             Find the ``variable`` formula for the given ``period`` if it exists, and apply it to ``entity``.
@@ -258,10 +264,7 @@ class Simulation(object):
             return None
 
         if self.trace:
-            parameters_at = lambda formula_period: TracingParameterNodeAtInstant(
-                self.tax_benefit_system.get_parameters_at_instant(formula_period),
-                self.tracer
-                )
+            parameters_at = self.trace_parameters_at_instant
         else:
             parameters_at = self.tax_benefit_system.get_parameters_at_instant
 
--- a/openfisca_core/simulations.py
+++ b/openfisca_core/simulations.py
@@ -262,7 +262,6 @@ class Simulation(object):
                 self.tax_benefit_system.get_parameters_at_instant(formula_period),
                 self.tracer
                 )
-            
         else:
             parameters_at = self.tax_benefit_system.get_parameters_at_instant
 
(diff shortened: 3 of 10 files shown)


sandcha:
We have simple parameters (like `taxes.income_tax_rate`) and scale parameters (like `taxes.social_security_contribution`). 

The following `example_2.json`:
```json
{
    "persons": {
        "Alicia": {
            "income_tax": {
                "2017-01": null
            },
            "social_security_contribution": {
                "2017-01": null
            }
        }
    },
    "households": {
        "_": {
            "parents": ["Alicia"]
        }
    }
}
```

Sent with this command line: 
```sh
curl http://localhost:5000/trace -H 'Content-Type: application/json' -d @example_2.json
```

Results in following web api response (containing simple and scale parameters):
```json
{
    "entitiesDescription": {
        "households": [
            "_"
        ],
        "persons": [
            "Alicia"
        ]
    },
    "requestedCalculations": [
        "social_security_contribution<2017-01>",
        "income_tax<2017-01>"
    ],
    "trace": {
        "income_tax<2017-01>": {
            "dependencies": [
                "salary<2017-01>"
            ],
            "parameters": {
                "taxes.income_tax_rate<2017-01-01>": 0.15
            },
            "value": [
                0.0
            ]
        },
        "salary<2017-01>": {
            "dependencies": [],
            "parameters": {},
            "value": [
                0.0
            ]
        },
        "social_security_contribution<2017-01>": {
            "dependencies": [
                "salary<2017-01>"
            ],
            "parameters": {
                "taxes.social_security_contribution<2017-01-01>": {
                    "0.0": 0.02,
                    "12400.0": 0.12,
                    "6000.0": 0.06
                }
            },
            "value": [
                0.0
            ]
        }
    }
}
```

openfisca_core/parameters.py
```diff
@@ -462,10 +462,15 @@ def __init__(self, name, node, instant_str):
         self._name = name
         self._instant_str = instant_str
         self._children = {}
+
+        # if trace:
+        self._original_children = {}
```
guillett: Why are changes in `ParameterNodeAtInstant` required ? I thought a new class would be self-contained.

Morendil: Basically because prior to this change ParameterNodeAtInstant didn't keep track of the actual `ParameterNode`s, `Parameter`s or `Scale`s that are its children, but only of the values attached to these children at its instant (the method `get_at_instant` is improperly named, it should really be `get_value_at_instant`). So if we didn't make the above change we could not get access to a parameter's name in `record_calculation_parameter_access`, only to its value.

Also see [this comment](https://github.com/openfisca/openfisca-core/pull/756#issuecomment-438006702) for more discussion of some design issues in this area.

openfisca_core/tracers.py
```diff
@@ -224,3 +247,25 @@ def print_computation_log(self, aggregate = False):
         """
         for node, depth in self._computation_log:
             self._print_node(node, depth, aggregate)
+
+
+class TracingParameterNodeAtInstant():
```
maukoquiroga: I'd give it a more classy name, depending of what we're doing here. Either

- An actual data structure => `ParameterNodeAtInstantTrace` (noun)
- A command to do something => `TraceParameterNodeAtInstant` (verb, imperative)

openfisca_core/parameters.py
guillett: :thinking: from my point of view (pretty far to be honest this links a production class to a debug one).

But if only the names is required, can't
`record_calculation_parameter_access(self, parameter, period):` be refactored in
`record_calculation_parameter_access(self, parameter_value, parameter_name, period):`
and `parameter_name` be provided from the `key` argument of `def __getattr__(self, key):` ?

tests/core/test_simulations.py
```diff
@@ -14,7 +14,20 @@
...
-    simulation.calculate('disposable_income', "2014-01")
+    simulation.calculate('income_tax', '2017-01')
+
+    salary_trace = simulation.tracer.trace['salary<2017-01>']
+    assert salary_trace['parameters'] == {}
+
+    income_tax_trace = simulation.tracer.trace['income_tax<2017-01>']
+    assert income_tax_trace['parameters']['taxes.income_tax_rate<2017-01-01>'] == 0.15
+
+    # Also trace parameters called with fancy indexing
```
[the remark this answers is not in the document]
Morendil: We're not. That was a confusion (indirect parameter access vs. fancy indexing). We're working on introducing a test that actually does test tracing parameter access while using fancy indexing.

openfisca_core/parameters.py
Morendil: It's a possibility. The original intent was to get at the "effective period" of a parameter (as opposed to the requested period); that proved very difficult and we gave up on that.

tests/core/test_simulations.py
```diff
@@ -14,7 +14,20 @@
 
 def test_calculate_with_trace():
```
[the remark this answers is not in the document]
Morendil: The "housing tax" variant invokes a formula that accesses the leaf parameter in two steps.

openfisca_core/tracers.py
Morendil: "Tracing" here is an adjective, it suggests a modification to the behaviour of the wrapped class. The class is not a data structure (it doesn't hold the trace, the Tracer does), nor is it a command (it doesn't have an `execute()` method or the like). This name is fine, stet.

tests/core/test_simulations.py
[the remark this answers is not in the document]
maukoquiroga: Thanks @Morendil. As examples come I guess from the `country_template`, I think by just giving more descriptive names to the tests will suffice.

openfisca_core/tracers.py
maukoquiroga: Ahhh you're using `tracing` as a noun!

openfisca_core/taxbenefitsystems.py
```diff
@@ -316,7 +316,7 @@ def get_parameters_at_instant(self, instant):
         Get the parameters of the legislation at a given instant
 
         :param instant: string of the format 'YYYY-MM-DD' or `openfisca_core.periods.Instant` instance.
-        :returns: The parameters of the legislation at a given instant.
+        :returns: The parameters of the legislation at a given instant (ParameterNodeAtInstant).
```
maukoquiroga: (ParameterNodeAtInstant) => (`ParameterNodeAtInstant`) ?

openfisca_core/parameters.py
fpagnoux: Strongly support @guillett's point. Adding `parameter_name` to the signature of `record_calculation_parameter_access` seems more elegant.

openfisca_core/tracers.py
```diff
@@ -224,3 +247,33 @@ def print_computation_log(self, aggregate = False):
...
+            child = self.parameter_node_at_instant._original_children[key]
+            period = self.parameter_node_at_instant._instant_str
+
+            if isinstance(child, ParameterNode):
+                return TracingParameterNodeAtInstant(child.get_at_instant(period), self.tracer)
+            else:
+                at_instant_trace = child.get_at_instant(period)
+                self.tracer.record_calculation_parameter_access(child, period)
+                return at_instant_trace
+        except KeyError:
```
fpagnoux: > It's usually preferable to minimise the amount of lines of codes included in the `try`.

openfisca_core/tracers.py
```diff
@@ -224,3 +247,33 @@ def print_computation_log(self, aggregate = False):
...
+            param_name = self.parameter_node_at_instant._compose_name(self.parameter_node_at_instant._name, key)
+            raise ParameterNotFound(param_name, self._instant_str)
+    
+    def __getitem__(self, key):
+        return self.parameter_node_at_instant[key]
+        # If fancy indexing is used, cast to a vectorial node
```
fpagnoux: Tracing parameters call using fancy indexing seems non-trivial.
If fancy indexing is using on a leaf of the parameter tree, like `P.single.owner[zone]`, then your parameter value is just a vector, and you can use your regular call to `record_calculation_parameter_access`.
If it's used on nodes, like `P.single[housing_occupancy_status].z1`, that's harder, as you have a more complex object to insert a tracer into...

openfisca_core/parameters.py
maukoquiroga: I tried to pass the tuple:

```python
        for child in node.children.items():
            child_at_instant = child[1]._get_at_instant(instant_str)
            if child_at_instant is not None:
                self._children[child_name] = child + (child_at_instant,)
                # ...
```

But it wreaks havoc...

openfisca_core/tracers.py
sandcha: Yes. We came to same conclusion with @Morendil: seems non-trivial. 😕 
Thanks a lot @guillett for recommending the test! 

So, as adding fancy indexed parameters in `/trace` will take some more time, we start by delivering this PR without them.

fpagnoux: Seems reasonable 🙂

openfisca_core/tracers.py
```diff
@@ -108,11 +112,30 @@ def record_calculation_start(self, variable_name, period, **parameters):
...
         self.stack.append(key)
         self._computation_log.append((key, len(self.stack)))
         self.usage_stats[variable_name]['nb_requests'] += 1
 
+    def record_calculation_parameter_access(self, parameter, period):
+        at_instant = parameter.get_at_instant(period)
+
+        if isinstance(at_instant, AbstractTaxScale):
+            values = at_instant.amounts if isinstance(at_instant, AmountTaxScale) else at_instant.rates
+            at_instant_trace = {}
```
fpagnoux: Defining a dict by setting the values one by one imperatively is very old school and doesn't read well. Prefer a comprehensive definition like:

```py
        {
            str(threshold): values[index]
            for index, threshold in enumerate(self.thresholds)
            }
```

openfisca_core/tracers.py
guillett: Is there a _real_ need to deliver that PR without _fancy indexing_? What is your definition of done from an end-user (who doesn't know when and where fancy-indexing is used)? For me, it makes sense to give another sprint' time.

tests/core/test_simulations.py
```diff
@@ -1,20 +1,72 @@
...
+    assert income_tax_trace['parameters']['taxes.income_tax_rate<2017-01-01>'] == 0.15
+
+    # Trace parameters called with indirect access
+    simulation.calculate('housing_tax', '2017')
+    housing_tax_trace = simulation.tracer.trace['housing_tax<2017>']
+    assert 'taxes.housing_tax<2017-01-01>' not in housing_tax_trace['parameters']
+    assert housing_tax_trace['parameters']['taxes.housing_tax.rate<2017-01-01>'] == 10
+    assert housing_tax_trace['parameters']['taxes.housing_tax.minimal_amount<2017-01-01>'] == 200
+
+    # Tracing of fancy indexed parameters is not yet supported
```
fpagnoux: If we were to test the tracer with fancy indexing, I think it would be easier to write a unit test on the tracer instead of mutating the test tax benefit system, which causes side-effects (cf. `__pycache__` bug you reported).

But we have not yet implemented it, so I don't really see the point of a test that check that we don't do something we are going to implement later.

openfisca_core/tracers.py
fpagnoux: Trying to find a solution

maukoquiroga: > Is there a real need to deliver that PR without fancy indexing?

YES!, 50 years of computer science and 20 years of agile software development have taught us that _divide and conquer_ is a good practice, and _scope creep_ is a bad one.

What I understand by _we **start** by delivering this PR without them_ is not that the issue is finished from an end-user perspective, only that letting scope creep and the pull request to rot increases the risk of this not being ever delivered.

> What is your definition of done from an end-user (who doesn't know when and where fancy-indexing is used)?

There is no such definition of done.

The proposal is to deliverer this incrementally and iteratively, which is a universal good development practice, not to abandon the effort half-way through, if that's what you're fearing.

> For me, it makes sense to give another sprint' time.

Sure. Which in no way prevent us from deliver.

Let's ship!

openfisca_core/tracers.py
fpagnoux: This part of the code has been refactored, resolving

openfisca_core/tracers.py
```diff
@@ -108,11 +112,30 @@ def record_calculation_start(self, variable_name, period, **parameters):
...
+            self.trace[key] = {'dependencies': [], 'parameters': {}}
+
         self.stack.append(key)
         self._computation_log.append((key, len(self.stack)))
         self.usage_stats[variable_name]['nb_requests'] += 1
 
+    def record_calculation_parameter_access(self, parameter, period):
+        at_instant = parameter.get_at_instant(period)
+
+        if isinstance(at_instant, AbstractTaxScale):
```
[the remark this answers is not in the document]
fpagnoux: Done

tests/core/test_simulations.py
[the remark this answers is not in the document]
fpagnoux: Added some unit tests in `test_tracer.py`. I think it's worth keeping an integration test like this one tough.

openfisca_core/taxbenefitsystems.py
fpagnoux: Done

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
...
+    # calculated variable period = 2017-01
+    "parameters": {
+      "taxes.social_security_contribution<2017-01-01>": {
+          "0.0": 0.02,
+          "12400.0": 0.12,
+          "6000.0": 0.06
+      }
+    },
+    ```
+  - Simple parameter example:
```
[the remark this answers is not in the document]
fpagnoux: Done

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
...
+    ```json
+    # calculated variable period = 2017-01
+    "parameters": {
```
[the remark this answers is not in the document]
fpagnoux: Done

openfisca_core/tracers.py
fpagnoux: Found a solution, we can actually use the same relatively simple mechanism for both cases.

openfisca_core/tracers.py
fpagnoux: Done

tests/core/test_simulations.py
Morendil: The point is that if we hadn't written this test we wouldn't have noticed that the tracing as originally *breaks* fancy indexing; what this unit test specifies is that fancy indexing still works in a simulation with trace on.

openfisca_core/tracers.py
Morendil: > Is there a real need to deliver that PR without fancy indexing?

Our reasoning was that fancy indexing is used in an extremely small fraction of all formulas, therefore the partial implementation delivers most of the value of the full one.

maukoquiroga:
Tested with `example.json`:

```json
{
    "persons": {
        "Alicia": {
            "income_tax": {
                "2017-01": null
            }
        }
    },
    "households": {
        "_": {
            "parents": ["Alicia"]
        }
    }
}
```

And commands:

```
make api
curl http://127.0.0.1:5000/trace -H 'Content-Type: application/json' -d @example.json
```

Gives, as expected:

```json
{
    "entitiesDescription":
    {
        "households": ["_"],
        "persons": ["Alicia"]
    },
    "requestedCalculations": ["income_tax<2017-01>"],
    "trace":
    {
        "income_tax<2017-01>":
        {
            "dependencies": ["salary<2017-01>"],
            "parameters":
            {
                "taxes.income_tax_rate<2017-01-01>": 0.15
            },
            "value": [0.0]
        },
        "salary<2017-01>":
        {
            "dependencies": [],
            "parameters":
            {},
            "value": [0.0]
        }
    }
}
```

openfisca_core/parameters.py
```diff
@@ -30,6 +30,7 @@
     from yaml import Loader
 
 
+ALLOWED_PARAM_TYPES = (float, int, bool, type(None))
```
maukoquiroga: Cool !

openfisca_core/simulations.py
```diff
@@ -248,6 +248,12 @@ def calculate_output(self, variable_name, period):
 
         return variable.calculate_output(self, variable_name, period)
 
+    def trace_parameters_at_instant(self, formula_period):
```
maukoquiroga: I just have a semantic confusion between `trace` and `tracing`. Not blocking.

openfisca_core/taxbenefitsystems.py
```diff
@@ -317,6 +317,7 @@ def get_parameters_at_instant(self, instant):
 
         :param instant: string of the format 'YYYY-MM-DD' or `openfisca_core.periods.Instant` instance.
         :returns: The parameters of the legislation at a given instant.
+        :rtype: :any:`ParameterNodeAtInstant`
```
maukoquiroga: Nice, I didn't know you could do `:any:`

tests/core/test_tracer.py
```diff
@@ -23,3 +27,30 @@ def test_variable_stats():
...
+
+def check_tracing_params(accessor, param_key):
+    tracer = Tracer()
+    tracer.record_calculation_start('A', '2015-01')
+    tracingParams = TracingParameterNodeAtInstant(parameters('2015-01-01'), tracer)
+    param = accessor(tracingParams)
+    assert_near(tracer.trace['A<2015-01>']['parameters'][param_key], param)
+
+
+def test_parameters():
```
maukoquiroga:
```suggestion
# TODO: refactor to use more expressive pytest fixtures
def test_parameters():
```

openfisca_core/taxscales.py
```diff
@@ -58,6 +58,13 @@ def copy(self):
         new.__dict__ = copy.deepcopy(self.__dict__)
         return new
 
+    def to_dict(self):
+        values = self.amounts if isinstance(self, AmountTaxScale) else self.rates
```
maukoquiroga:
```suggestion
        #  TODO: Harmonise 'rates' and 'amounts' protocols, in order to avoid this kind of check.
        values = self.amounts if isinstance(self, AmountTaxScale) else self.rates
```

PR #767, opened 2018-11-14
Introduce ParameterNode.get_descendants

fpagnoux:
- Introduce `ParameterNode.get_descendants`
  - This method returns a generator containing all the parameters and nodes recursively contained in the `ParameterNode`

(I need this helper to parse the IPP parameters, and as it's not the first time we need this, I think a method would is useful)

### DIFF
--- a/openfisca_core/parameters.py
+++ b/openfisca_core/parameters.py
@@ -450,7 +450,7 @@ class ParameterNode(object):
 
     def get_descendants(self):
         """
-            Returns all the parameters and nodes recursively contained in this `ParameterNode`
+            Return a generator containing all the parameters and nodes recursively contained in this `ParameterNode`
         """
         for child in self.children.values():
             yield child
--- a/openfisca_core/parameters.py
+++ b/openfisca_core/parameters.py
@@ -448,6 +448,16 @@ class ParameterNode(object):
             return result.encode('utf-8')
         return result
 
+    def get_descendants(self):
+        """
+            Returns all the parameters and nodes recursively contained in this `ParameterNode`
+        """
+        for child in self.children.values():
+            yield child
+            if isinstance(child, ParameterNode):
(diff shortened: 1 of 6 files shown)


openfisca_core/parameters.py
```diff
@@ -448,6 +448,16 @@ def __repr__(self):
             return result.encode('utf-8')
         return result
 
+    def get_descendants(self):
+        """
+            Return a generator containing all the parameters and nodes recursively contained in this `ParameterNode`
+        """
+        for child in self.children.values():
+            yield child
+            if isinstance(child, ParameterNode):
```
Morendil: Prefer a polymorphic call instead of explicitly switching on type.

fpagnoux: Could you write a code suggestion, or a least more indication about how you would do that? 

Does this mean writing a `get_descendants` that returns an empty iterator for leaves?

Morendil: Yes, and call get_descendents recursively on line 457. I'm not well-versed in Python generators so I'm not sure if there would be a gotcha there though. Rather than an empty iterator I would have imagined a no-op (`pass`) implementation for leaves would be enough, but I'm not quite sure. What would the following require? (Would it work?)

```Python
for child in self.children.values():
    yield child
    yield child.get_descendants()
```

fpagnoux: `yield from child.get_descendants()` would work in Python 3, not in Python 2 (see comment below).

With a no-op method, we would still need to check if the descendants are None, otherwise trying to iterate on it would cause an error:

```py
for child in self.children.values():
  yield child
  child_descendants = child.get_descendants()
  if child_descendants is not None:
    for descandant in child_descendants:
      yield descendant
```

fpagnoux: Just switched to polymorphism

PR #769, opened 2018-11-15
Fix inconsistent access to sub-parameters

### LINKED ISSUE #762: Can't access extension parameter for some syntaxes
Hi there!

### Here is what I did:

I tried to using `openfisca-extension-template`.
I checked that everything was loaded by counting extended model parameters and variables
and accessing a parameter:
```py
    assert tbs.parameters('2016-01').local_town.child_allowance.amount == 100.0
    assert tbs.parameters.local_town.child_allowance.amount('2016-01') == 100.0
    assert tbs.parameters.local_town.child_allowance.amount is not None
```

No error. All assertions ok.

One assertion OK:
```
    assert tbs.parameters('2016-01').local_town.child_allowance.amount == 100.0
```

Two assertions KO:
```
    assert tbs.parameters.local_town.child_allowance.amount('2016-01') == 100.0
    assert tbs.parameters.local_town.child_allowance.amount is not None
```
For same reason:
```
AttributeError: 'ParameterNode' object has no attribute 'local_town'
```

As in this [CircleCI build](https://circleci.com/gh/openfisca/openfisca-core/2898).

See test on openfisca-core [extension-parameters](https://github.com/openfisca/openfisca-core/blob/e059feced8c9d7248d2bf5380379ed1e18f9436d/tests/core/test_extensions.py#L25) branch.

## Context

I identify more as a:

Morendil:
Fixes #762 

#### Bug fixes

- In `parameters.py`, there is rampant duplication of the following two lines:

```Python
self.children[child_name] = child
setattr(self, child_name, child)
```

However, in a few cases - including merging parameters from extensions, as in the issue referenced - the array is written to, without the call to setattr taking place. This will result in an inconsistent state and a ParameterNotFound exception even though the parameter seems to have been correctly loaded. This PR aims to channel all additions of a descendant to a parameter node via add_child.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,15 @@
 # Changelog
 
-### 24.9.5 [#771](https://github.com/openfisca/openfisca-core/pull/771)
+### 24.9.7 [#769](https://github.com/openfisca/openfisca-core/pull/769)
+
+- Unify the protocol for appending sub-parameters
+
+### 24.9.6 [#771](https://github.com/openfisca/openfisca-core/pull/771)
 
 - Improve serve command by documenting the bind option
 - Avoid crashing when no arguments are supplied
 
-### 24.9.4 [#774](https://github.com/openfisca/openfisca-core/pull/774)
+### 24.9.5 [#774](https://github.com/openfisca/openfisca-core/pull/774)
 
 - Clarify the error message when assigning a value larger than MaxInt32 to an 'int' variable
 
--- a/openfisca_core/parameters.py
+++ b/openfisca_core/parameters.py
@@ -380,15 +380,13 @@ class ParameterNode(object):
                     else:
                         child_name_expanded = _compose_name(name, child_name)
                         child = load_parameter_file(child_path, child_name_expanded)
-                        self.children[child_name] = child
-                        setattr(self, child_name, child)
+                        self.add_child(child_name, child)
 
                 elif os.path.isdir(child_path):
                     child_name = os.path.basename(child_path)
                     child_name_expanded = _compose_name(name, child_name)
                     child = ParameterNode(child_name_expanded, directory_path = child_path)
-                    self.children[child_name] = child
-                    setattr(self, child_name, child)
+                    self.add_child(child_name, child)
 
         else:
             self.file_path = file_path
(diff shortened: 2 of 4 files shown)


fpagnoux:
On a meta note, we usually pretty rarely dismiss reviews unless:
- somethings urgent (major bug, CI is broken) and the blocking person is not available at the moment
- the reviewers in on vacation

CHANGELOG.md
```diff
@@ -1,11 +1,15 @@
 # Changelog
 
-### 24.9.5 [#771](https://github.com/openfisca/openfisca-core/pull/771)
+### 24.9.7 [#769](https://github.com/openfisca/openfisca-core/pull/769)
+
+- Unify the protocol for appending sub-parameters
```
fpagnoux: Not sure this is limpid for a user of the OpenFisca-Core library. 
What does mean "appending sub-parameters" from them? Is there a _protocol_ for that? 

From the user point of view, I think what has been done here is:

```
- Fix an inconsistency in parameter navigation
   - In some cases, for instance while using extensions, the `parameters.path.to.subparams` notation was not working
```

On a meta note, I think this is another reason why the Changelog should ideally written before the review, so that it can be improve, as it's a pretty important part of each PR.

Morendil: Yes, too implementation-centric, thanks for the suggestion; I can pick it up on a later PR. I didn't much like "fix" as a verb though, it's too generic, and that's part of why I looked for more specific terms. How about

- Ensure `path.to.parameter` syntax works consistently across country and extension

On the meta level: fixing unnecessary merge conflicts is one of the least appealing parts of the merge process at the moment, so much I'd make the diametrically opposed policy proposal, i.e. delay the CHANGELOG.md/setup.py until just before merge.

As a "best of both worlds" proposal, I suggest improving the description collaboratively by editing the PR's description at review time. Then creating the CHANGELOG entry is a matter of copy/paste (until we automate it away).

Morendil: (Never mind, I hadn't seen the fixup PR.)

fpagnoux: We can include the fixup PR content in another PR to come (especially as right now editing the changelog triggers the version bump requirement), and feel free to edit the text it if you like.

PR #771, opened 2018-11-16
Improve serve command

### LINKED ISSUE #634: Add a "host" option to openfisca-serve
_Imported from openfisca/openfisca-web-api#132 by @Morendil._

When using the `openfisca serve` script, I can specify a port but not a host to bind to; the script defaults to `localhost` which is appropriate only in a narrow range of conditions. It will for instance prevent usage from a Docker container, a throwaway virtual machine on a cloud provider, or that home NAS that you use as a server for tinkering with stuff.

Possibly a better default is `0.0.0.0`, a pseudo-host which will result in the server socket binding on every available interface. Alternately, a config option `--host` to override `localhost` is perfectly suitable.

Morendil:
Fixes #634 

We need to be able to specify host as well as port in `openfisca serve`.

Absent a specific `host` option, `openfisca serve --bind 0.0.0.0:7777` will do, as the [documentation](https://openfisca.org/doc/openfisca-python-api/openfisca_serve.html) hints at.

Discoverability of this was limited in two ways:
- the bare command `openfisca` crashed with `AttributeError: 'Namespace' object has no attribute 'vars'`
- the help option `openfisca serve -h` does not mention gunicorn's `--bind`

This PR fixes both shortcomings.

### DIFF
--- a/openfisca_web_api/scripts/serve.py
+++ b/openfisca_web_api/scripts/serve.py
@@ -34,7 +34,7 @@ def define_command_line_options(parser):
     parser = add_tax_benefit_system_arguments(parser)
 
     # Define server configuration
-    parser.add_argument('-p', '--port', action = 'store', help = "port to serve on", type = int)
+    parser.add_argument('-p', '--port', action = 'store', help = "port to serve on (use --bind to specify host and port)", type = int)
     parser.add_argument('--tracker-url', action = 'store', help = "tracking service url", type = str)
     parser.add_argument('--tracker-idsite', action = 'store', help = "tracking service id site", type = int)
     parser.add_argument('--tracker-token', action = 'store', help = "tracking service authentication token", type = str)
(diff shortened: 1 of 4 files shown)


fpagnoux:
> the bare command openfisca crashed with AttributeError: 'Namespace' object has no attribute 'vars'

Hum, the bare `openfisca` command was not supposed to trigger a `serve`. The initial behavior was to crash and indicate that the command was missing. This still works in Python 2.7, but not in Python 3, as the behavior or ArgParse changed.

fpagnoux:
> Hum, the bare openfisca command was not supposed to trigger a serve. The initial behavior was to crash and indicate that the command was missing. This still works in Python 2.7, but not in Python 3, as the behavior or ArgParse changed.

Just pushed a small commit to adapt to Python 3:
image

openfisca_web_api/scripts/serve.py
```diff
@@ -48,7 +48,7 @@ def read_user_configuration(default_configuration, command_line_parser):
     configuration = default_configuration
     args, unknown_args = command_line_parser.parse_known_args()
 
-    if args.configuration_file:
+    if 'configuration_file' in vars(args) and args.configuration_file:
```
fpagnoux: Now that we throw a clear error message in case of a bare `openfisca`, I think we don't need this anymore.

```suggestion
    if args.configuration_file:
```

PR #773, opened 2018-11-16
Cap all dependency versions

maukoquiroga:
Applies learning from #772 

Make sure to cap all dependency versions, in order to avoid unwanted functional and integration breaks caused by external code updates.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,13 +5,15 @@
 from __future__ import unicode_literals, print_function, division, absolute_import
 from setuptools import setup, find_packages
 
+# Please make sure to cap all dependency versions, in order to avoid unwanted
+# functional and integration breaks caused by external code updates.
 
 general_requirements = [
     'Biryani[datetimeconv] >= 0.10.8',
     'dpath == 1.4.0',
     'enum34 >= 1.1.6',
-    'future',
-    'nose',  # For openfisca-run-test
+    'future < 1.0.0',
+    'nose < 2.0.0',  # For openfisca-run-test
     'numpy >= 1.11, < 1.16',
     'psutil == 5.4.6',
     'PyYAML >= 3.10',
@@ -28,7 +30,7 @@ dev_requirements = [
     'autopep8 == 1.4.0',
     'flake8 >= 3.5.0, < 3.6.0',
     'pycodestyle >= 2.3.0, < 2.4.0',  # To avoid incompatibility with flake8
-    'pytest < 4.0',
+    'pytest < 4.0.0',
     'openfisca-country-template >= 3.4.0, < 4.0.0',
     'openfisca-extension-template >= 1.1.3, < 2.0.0',
     ] + api_requirements
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ dev_requirements = [
 
 setup(
     name = 'OpenFisca-Core',
-    version = '24.8.1',
+    version = '24.8.2',
     author = 'OpenFisca Team',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -28,14 +30,14 @@
     'autopep8 == 1.4.0',
     'flake8 >= 3.5.0, < 3.6.0',
     'pycodestyle >= 2.3.0, < 2.4.0',  # To avoid incompatibility with flake8
-    'pytest < 4.0',
+    'pytest < 4.0.0',
```
fpagnoux: This is strictly the same 😄

setup.py
```diff
@@ -5,13 +5,15 @@
...
+# Please make sure to cap all dependency versions, in order to avoid unwanted
+# functional and integration breaks caused by external code updates.
 
 general_requirements = [
     'Biryani[datetimeconv] >= 0.10.8',
     'dpath == 1.4.0',
     'enum34 >= 1.1.6',
-    'future',
-    'nose',  # For openfisca-run-test
+    'future < 1.0.0',
```
fpagnoux: I doubt this protect us from anything. Pre-`1.0`, there can be breaking changes without bumping the first digit...

fpagnoux:
I must say I'm not a fan of merging PR in a "faster-than-our-shadow" style (unlike there is a good reason to do so, like [fixing the CI](https://github.com/openfisca/openfisca-core/pull/772/commits/d987557c3cc43502187369e237053ec635fdd8b9) . 

I think this leads to bugs, and/or half-done job. 

How would @openfisca/core-contrib feel about a default 24 hours delay between the opening of a PR and the merge, unless there is an actual good reason to rush?

setup.py
Morendil: Appreciated for consistency though.

Morendil:
@fpagnoux Not necessarily against, but not entirely sure that it makes sense as a heuristic; less so than our notion yesterday that "Approve entails taking as much (if not more) responsibility for the merge as if you'd authored the PR in the first place". (Which formulation feels like it puts more pressure on me to solve the problems linked to hasty approvals, than "I followed the 24-hour rule so I'm OK".)

I'd prefer to continue receiving coaching in "here's what to pay attention to in review", so that it doesn't matter if I happen to review in the same hour that the PR is created, and the PR is then be merged in a similar timeframe. That includes developing a sense of when it makes sense to wait for more eyes.

In this particular case it seems to me, in conscience, that the original PR was fine and the post-merge remarks above would not have blocked the merge anyway.

fpagnoux:
> In this particular case it seems to me, in conscience, that the original PR was fine and the post-merge remarks above would not have blocked the merge anyway.

I understand that the initial PR was doing some good, as it was setting a relevant version cap for at least one dependency, but I'd be annoyed if our Changelog started to look like this:

```
### 24.8.4 
  - Fixup one dependency cap
       - Just realized we forgot one 

### 24.8.3
  - Actually make sure to cap all dependency versions
       - One of the version cap was not doing anything 

### 24.8.2 
  - Make sure to cap all dependency versions...
```

which I see as a risk of the "any PR doing with a net positive amount of good should be merged".

fpagnoux:
> (Which formulation feels like it puts more pressure on me to solve the problems linked to hasty approvals, than "I followed the 24-hour rule so I'm OK".)

I don't see the 24 hours heuristics as a way to incentivize thorough reviews, but more as the way to give the community a chance to look at upcoming changes before they are merged. If OpenFisca is to become more of an international digital common, that would become even more important.

PR #774, opened 2018-11-16
Deal with OverflowError

Morendil:
Fixes https://github.com/openfisca/openfisca-france/issues/1197

#### Technical changes

- In `Entity.init_from_json/init_variable_values`, attempting to assign a value larger than MaxInt32 resulted in a somewhat cryptic OverflowError.
- This PR fixes that, and also nudges the language away from "Invalid X", which blames the user for a deficiency of the software, and towards "Can't handle X", which acknowledges the deficiency.

This PR does NOT as yet include a test. The presumed fix for the issue consists of catching OverflowError in `init_variable_values`. However:
- `init_variable_values` has no existing unit tests
- its caller, `init_from_json` has no existing unit tests
- it's not possible to instantiate the class Entity in isolation - it requires a Simulation, which in turns require a TaxBenefitSystem, which in turn requires one or more entities
- entities in general seem to be tested only indirectly through the country template

In summary, the design of Core in this area is (in my expert opinion) untestable at the unit test level. This is a serious but fixable flaw. It is better fixed as a pair or mob activity, so as to amplify the relevant design and TDD skills in the team. I'm happy to offer my help there; meanwhile my recommendation is to merge this PR as a strict improvement over the current state of the codebase.

### DIFF
--- a/openfisca_core/entities.py
+++ b/openfisca_core/entities.py
@@ -104,7 +104,7 @@ class Entity(object):
                     try:
                         array[entity_index] = value
                     except (OverflowError):
-                        error_message = "Can't deal with value: '{}', it's too large for type {}.".format(value, holder.variable.value_type)
+                        error_message = "Can't deal with value: '{}', it's too large for type '{}'.".format(value, holder.variable.json_type)
                         raise SituationParsingError(path_in_json, error_message)
                     except (ValueError, TypeError):
                         if holder.variable.value_type == date:
--- a/openfisca_core/entities.py
+++ b/openfisca_core/entities.py
@@ -79,7 +79,7 @@ class Entity(object):
 
             if not isinstance(variable_values, dict):
                 raise SituationParsingError(path_in_json,
-                    "Invalid type: must be of type object. Input variables must be set for specific periods. For instance: {'salary': {'2017-01': 2000, '2017-02': 2500}}, or {'birth_date': {'ETERNITY': '1980-01-01'}}.")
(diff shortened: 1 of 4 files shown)


openfisca_core/entities.py
```diff
@@ -98,16 +98,19 @@ def init_variable_values(self, entity_object, entity_id):
...
                             possible_values = [item.name for item in holder.variable.possible_values]
                             raise SituationParsingError(path_in_json,
-                                "'{}' is not a valid value for '{}'. Possible values are ['{}'].".format(
+                                "'{}' is not a known value for '{}'. Possible values are ['{}'].".format(
                                     value, variable_name, "', '".join(possible_values))
                                 )
                     try:
                         array[entity_index] = value
+                    except (OverflowError):
+                        error_message = "Can't deal with value: '{}', it's too large for type {}.".format(value, holder.variable.value_type)
```
fpagnoux: This gives `{"2017":"Can't deal with value: '10000000000000000000', it's too large for type <class 'int'>."}` which doesn't read super well.

fpagnoux:
```suggestion
                        error_message = "Can't deal with value: '{}', it's too large for type '{}'.".format(value, holder.variable.json_type)
```

fpagnoux:
> In summary, the design of Core in this area is (in my expert opinion) untestable at the unit test level. This is a serious but fixable flaw. It is better fixed as a pair or mob activity, so as to amplify the relevant design and TDD skills in the team. I'm happy to offer my help there; meanwhile my recommendation is to merge this PR as a strict improvement over the current state of the codebase.

I would be interested in joining such a mob programming session, if it's remote friendly 🙂 . There is indeed a lot of coupling between entities and simulations.

PR #775, opened 2018-11-19
Fix amount scale parameters' access

### LINKED ISSUE #770: Unable to detect error on amount scales
Hi there!

### Here is what I did:

On `openfisca-france`, we added two amount scales in same yaml file and called them.

The first time, we forgot to add `value` attribute in brackets:
```yaml
    - amount:
         2014-01: 0.23
       threshold:
         2014-01: 6
```

Then, we added the `value` attributes. See [this commit](https://github.com/openfisca/openfisca-france/commit/e57fa6de128494de069a230db7e3567ee9f8a500#diff-e79521b2f36d9e2e8fc12f9789d0479aR4541) (call first, new parameters under).

First time, we expected to see an error explaining that we missed the `value` attribute.

Then, we expected to see no error at amount scale call (at [this line](https://github.com/openfisca/openfisca-france/commit/e57fa6de128494de069a230db7e3567ee9f8a500#diff-e79521b2f36d9e2e8fc12f9789d0479aR4541)).

To call it, we used `make test`.

The same error is raised without and with `value` attributes:
```
openfisca_core.parameters.ParameterNotFound: b"The parameter 'impot_revenu.reductions_impots.rpinel.taux.metropole' was not found in the 2017-01-01 tax and benefit system."
```

It breaks on `openfisca-france` with `pytest tests/test_api.py`.
See this [CircleCI build](https://circleci.com/gh/openfisca/openfisca-france/5957)

See `openfisca-france`, [refactor-rpinel-bottom-up](https://github.com/openfisca/openfisca-france/tree/e57fa6de128494de069a230db7e3567ee9f8a500) branch.

## Context

I identify more as a:

pblayo:
Fixes #770

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 24.9.1 [#770](https://github.com/openfisca/openfisca-core/pull/770)
+
+- Fix a bug that prevented amount scale parameters from being accessed in a formula
+
 ## 24.9.0 [#767](https://github.com/openfisca/openfisca-core/pull/767)
 
 - Introduce `ParameterNode.get_descendants`
--- a/openfisca_core/parameters.py
+++ b/openfisca_core/parameters.py
@@ -724,6 +724,7 @@ class Scale(object):
                     amount = bracket.amount
                     threshold = bracket.threshold
                     scale.add_bracket(threshold, amount)
+            return scale
         elif any('average_rate' in bracket._children for bracket in brackets):
             scale = taxscales.LinearAverageRateTaxScale()
 
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ dev_requirements = [
 
 setup(
     name = 'OpenFisca-Core',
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 24.9.1 [#770](https://github.com/openfisca/openfisca-core/pull/770)
+
+- Fix instant (period) for AmoutTaxScale.
+  - Unblock the factorization of the French's tax system https://github.com/openfisca/openfisca-france/pull/1213 (Pinel)
```
Morendil: This is somewhat contingent, I'm not sure it's worth mentioning in a Changelog.

In the line above I would rather take the user perspective and highlight that we are solving a rather serious problem, e.g. "Fix a bug that prevented amount scale parameters from being accessed in a formula".

pblayo: Thanks @Morendil for your suggestion, I've taken it into account.

PR #777, opened 2018-11-21
Don't impose log config

fpagnoux:
- Allow OpenFisca-Core users to define their own level of log)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 24.9.4 [#777](https://github.com/openfisca/openfisca-core/pull/777)
+
+- Allow OpenFisca-Core users to define their own level of log
+
 ### 24.9.3 [#778](https://github.com/openfisca/openfisca-core/pull/778)
 
 - Remove the reference doc source, as it's been moved to https://github.com/openfisca/openfisca-doc.
--- a/openfisca_core/__init__.py
+++ b/openfisca_core/__init__.py
@@ -1,4 +1,0 @@
-from __future__ import unicode_literals, print_function, division, absolute_import
-import logging
-
-logging.basicConfig()
--- a/openfisca_core/scripts/__init__.py
+++ b/openfisca_core/scripts/__init__.py
@@ -8,7 +8,6 @@ import pkgutil
 from os import linesep
 
 log = logging.getLogger(__name__)
-logging.basicConfig(format='%(levelname)s: %(message)s')
 
 
 def add_tax_benefit_system_arguments(parser):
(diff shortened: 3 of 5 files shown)


Morendil:
This doesn't seem to solve the problem I've been struggling with.

Steps to repro:
- instrument code in e.g. `simulations.py`, for instance insert after line 147 in `calculate`: `log.debug("Calculate {}".format(variable_name))`
- run a test from the France model with `openfisca-run-tests`

The debug logs are emitted in the absence of a -v option to `openfisca-run-tests`, on `master` and on this PR's branch both.

Running the Core tests with PyTest doesn't seem to emit unexpected logs with this PR, but neither did it do so in `master`.

pblayo:
> "Request changes" is a bit of a misnomer here as I don't know what changes would make this PR good to merge, but I'd prefer not to merge if it doesn't improve the behaviour.

Then, @Morendil and @fpagnoux I'd like to close this PR. It can always be reopen later if need be.

fpagnoux:
I confirm I reproduce the issue. 

However, logging within the `run_test.py` file didn't work as expected before this PR, and works well after.

fpagnoux:
It seems to come from `nose`:
- logging in the `run_tests` function in `test_runner.py` works as expected after this PR
- logging in the `generate_tests` function in the same file always print debug logs. This function passed as a lambda to `nose.run()`, so I think something happens there.

fpagnoux:
Ok, digging more into nose, the issue seems to be related to the nose plugin that captures logs. I deactivated, and it seems to solve the issue.

@Morendil can you confirm? 

The investigation was not really fun, this is IMO another reason to move away from nose.

Morendil:
@fpagnoux Makes sense! Yes, no unexpected debug logs now, and `-v` does its thing.

Yes, it's really annoying that the capture output plugin works that way - I'd have expected it to be logging-aware, so that `print` calls are captured but `log` calls handled normally.

On the other hand, I do rely on captured output quite often so being able to have both would be valuable; how about making `nologcapture` a default but still have a way to turn it back on? For example on a `--logcapture` argument?

fpagnoux:
> Yes, it's really annoying that the capture output plugin works that way - I'd have expected it to be logging-aware, so that print calls are captured but log calls handled normally.

>On the other hand, I do rely on captured output quite often so being able to have both would be valuable; how about making nologcapture a default but still have a way to turn it back on? For example on a --logcapture argument?

From what I've seen, it seems that print capturing and log capturing are handled by two different nose plugins, and can be deactivated separately. 
Print capturing is right now activated by default, unless the `-v` option is provided.
Log capturing doesn't seem to work as expected, as activating the capture brings _more_ logs, which is conter-intuitive. It is deactivated in this PR. 

Unless you're seeing a case where, despite this weird behavior, log capturing still does something useful, I'd welcome any additional commits on this branch, but right now I'm not seeing any.

Morendil:
Ugh, that's even weirder.

Actually log capture _does_ do something useful - it outputs the logs close to the error that caused a test to fail, as opposed to right in the middle of the stream of pass/fail results (dots, E's and F's). But not so useful that I can't live without.

Let's talk soon about the POC I had to run YAML tests with PyTest, and how that can be merged with the new parsing of these tests. Meanwhile, this is GTM for me.

PR #778, opened 2018-11-21
Adapt to new doc

fpagnoux:
- Remove the reference doc source, as it's been moved to https://github.com/openfisca/openfisca-doc.
- Rebuild the doc when the Core codebase changes
- Update links


fpagnoux:
> did you consider using https links instead of http?

Good point, I think at the time HTTPS was not available for github pages with a custom domain name, but it seems to be working now.

PR #779, opened 2018-11-22
Fixup changelog

fpagnoux:


fpagnoux:
See https://github.com/openfisca/openfisca-core/pull/769/files#r235811409 for a better wording suggestion (cc @Morendil)
Also, it seems easier to just sneak the content of the PR in another PR, to avoid a version bump.

fpagnoux:
See previous comment

Issue #716, opened 2018-08-23
Align YAML tests and JSON API situation description


fpagnoux:
Right now, the description of a situation in a YAML test and in the Web API have a slightly different structure:

- In the JSON API, they follow [this structure](http://openfisca.org/doc/openfisca-web-api/input-output-data.html#describing-entities)
- In the YAML tests, the structure in inherited from the former Web API, which is slightly different. 

This is often [confusing for users](https://github.com/openfisca/openfisca-core/issues/599#issuecomment-349316330)

The test runner also currently depends on Scenarios to parse these YAML files, while we've been trying to uncouple scenarios from the core of OpenFisca.  

We should align the test structure to the API one, and process them the same way.

fpagnoux:
Solving this issue means defining a new format for YAML tests. Here are some thoughts:

- As mentioned in the opening post, we want to avoid differences between the YAML test structure and the JSON request one.
    - to avoid confusion
    - so that an user can easily just copy-paste a simulation they sent to the API in a YAML test
- However, strictly limiting YAML tests syntax to the current Web API syntax is really heavy for test writers:
    - Unlike API requests, which are coded "once and for all" and then automatically generated, tests are written manually. Therefore, every piece of boilerplate can become a hassle, and discourage testing. We certainly don't want that!

Trying to balance these two principles, here is a proposition of specification for a new test format:

fpagnoux:
#### RFC 1 - General structure of a YAML test:

```YAML
- name: "Income tax - basic case"
  description: ...
  ...  # other test-level data, such as error margin, reforms to use, etc.
  input:
    ... # Simulation input
  output:
    ...  # Expected output of the simulation
```

The main change relatively to today's format is to:
  - clearly separate input description from test-level data

fpagnoux:
#### RFC 2 - Entityless mode

The following format is valid:

```YAML
- name: "Income tax - basic case"
  input:
    salary:
      2018-10: 2000
  output:
    income_tax:
      2018-10: 300
```

In this case, we make the assumption that:
  - There is only one person in the simulation
  - There is only one instance of each group entity in the simulation (family, household, etc.)
  - This person has, in each group entity, the first role (e.g. `first_parent`, `principal_caregiver`)

fpagnoux:
### RFC 3 - Simplified period mode

The following format is valid:

```YAML
- name: "Income tax - basic case"
  period: 2018-10
  input:
    salary: 2000
  output:
    income_tax: 300
```

[...]

Issue #691, opened 2018-07-20
Refactor code using Biryani


maukoquiroga:
As it is no longer maintained, we should either :

- replace it for other specific libraries doing currying, type checking and conversions separately
- refactor code to do python-style composition
- avoid composition

Note that most code using it is untested.

Connected to openfisca/openfisca-core#716
Connected to #540

fpagnoux:
FI, we tried to remove dependencies to Biryani each time we worked on a module since late 2016. 

While the Byriani is intellectually seducing, practically it was really hard to debug and find what's going on. We replaced it with basic python code, which can be a little tedious, but allows us to finely set our error message.

I think the last Byriani stronghold is the scenarios, which are used to parse tests.

Morendil:
> the last Byriani stronghold is the scenarios

Not quite, `periods.py` is also heavily dependent on `conv`.

fpagnoux:
True, but there is not much complex logics there, unlike scenarios

PR #780, opened 2018-11-24
Allow non-integral inputs to int variables

### LINKED ISSUE #463: Support decimal values for integer-typed values
As an _analyst_ or _producer_ depending for a computation on a value defined as an integer in the legislation,
I can provide that value as a decimal,
So that I don't have to systematically clean my input data

As a _contributor_ writing a formula defined as an integer in the legislation,
I can define as formula metadata how the input data should be normalised into an integer,
So that I don't have to write my own rounding code on each formula

Morendil:
Fixes #463

#### Technical changes

- Allow non-integral inputs to int variables

-----

The unit test included in this PR shows that a Holder will happily accept an array of float as an input to an `int` variable.

The underlying cause here is that parsing test cases still relies on `scenarios` which we would like to drop support for, per #716. Scenarios in turn rely on `columns`, of which [the code says they are "considered deprecated"](https://github.com/openfisca/openfisca-core/blob/56e31c6186e1e354d074ee1e6f4f760442b44981/openfisca_core/columns.py#L16). Columns and scenarios rely on `conv`, which I understand is our main remaining link with [Biryani](https://github.com/etalab/biryani). We have been saying for a while that we intend to drop Biryani, e.g. #691.

This looks like a textbook case for applying the [Mikado Method](http://mikadomethod.info/). ;)

### DIFF
--- a/tests/core/test_holders.py
+++ b/tests/core/test_holders.py
@@ -178,3 +178,11 @@ def test_set_not_chaged_variable():
     array = np.asarray([2000])
     holder.set_input('2015-01', array)
     assert_equal(simulation.calculate('salary', '2015-01'), array)
+
+
+def test_set_input_float_to_int():
+    simulation = get_simulation(couple)
+    age = np.asarray([50.6])
+    simulation.person.get_holder('age').set_input(period, age)
+    result = simulation.calculate('age', period)
+    assert_equal(result, np.asarray([50]))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 24.9.8 [#780](https://github.com/openfisca/openfisca-core/pull/780)
+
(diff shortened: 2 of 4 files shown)


tests/core/test_holders.py
```diff
@@ -178,3 +178,11 @@ def test_set_not_chaged_variable():
     array = np.asarray([2000])
     holder.set_input('2015-01', array)
     assert_equal(simulation.calculate('salary', '2015-01'), array)
+
+
+def test_set_input_float_to_int():
```
fpagnoux: This tests pass without the change on `columns.py` 😅 

(tests in this file don't use `scenarios`)

Morendil: Yep, that's on purpose - documenting at what level the desired behaviour is being blocked.

Morendil:
It's not very urgent, but it allows us to close #463 right away with a clean conscience.

Issue #540, opened 2017-06-29
Remove scenarios


MattiSG:
They are poorly named, as they do three things:

1. Input validation through entities consistency checks.
2. Relationships completion between entities.
3. Parse the situation Python instance from an input.

## Consensus

- Refactor by renaming & splitting roles.

## Open questions

1. Should we keep input validation? This is in the law, but is not consistently handled by OpenFisca.
2. Should we replace relationships completion by situation templates?
3. How to support transition for users who rely on scenarios?

Morendil:
One additional things that scenarios did (still do, pending PRs currently in review):
- manage "axes" - these are a feature undocumented (except in [notebooks](https://github.com/openfisca/tutorial/blob/ee8d873889eec7e7d887f9610e6225443e0c4ca0/notebooks/how_to_handle_axes.ipynb)) whereby an input variable can give rise to a set of values increasing by a fixed step

sandcha:
Linked to #691

PR #781, opened 2018-11-26
Uniformize tests with the new API

pblayo:
Connected to #716


CHANGELOG.md
```diff
@@ -1,5 +1,121 @@
...
+  > Note for country package using Scenarios (e.g. France, Tunisia):
+  > Tests are not using scenarios anymore. Therefore, tests cannot partially specify entities anymore. Tests using entities inference may need manual adaptation in addition to the script.
+
+  - The `Entity` constructor (usually not directly called by users) does not accept an `entities_json` parameter anymore.
+
+
+#### Deprecation
+
+  - Deprecate `openfisca-run-test`
+    - `openfisca test` should be used instead.
```
fpagnoux: Unifying `openfisca serve` and `openfisca-run-test` in a single `openfisca` command was the plan when `openfisca serve` was introduced.

This seem like a good time to do it. It could be `openfisca run-test` if you think it's better, no strong opinion (but `test` is shorter and also a verb.)

Morendil:
@fpagnoux Currently reviewing and refactoring, both to enhance testability and to gain a better understanding of the code. First pass this afternoon with @pblayo, I'll continue tomorrow with William (possible contributor in the future).

I may have a few other comments as well.

Morendil:
To leave a trace of the thinking behind the session with @pblayo, we reviewed with the following heuristics in mind:
- a unit test will generally target one method of one class
- a unit test should be able to instantiate/fixture the class under test with a no-args constructor
- a unit test should not have to instantiate/fixture anything beyond the inputs to the method under test
- all methods of the class under test should have one or more unit tests
- the code under tests should be more [demeter](https://en.wikipedia.org/wiki/Law_of_Demeter)-ish and [functional](https://en.wikipedia.org/wiki/Referential_transparency)

In this first step we decoupled SimulationBuilder from TaxBenefitSystem, removing domain logic from the constructor and moving it to TaxBenefitSystem, and injecting TaxBenefitSystem as an explicit parameter rather than as SimulationBuilder state.

Some afterthoughts:
- each exception raised should have a unit test where it's expected with pytest.raises
- TaxBenefitSystem seems to want to have an EntityContainer piece carved off; the Builder doesn't really need *all* of it, except to pass it as a dependency to Simulation

As usual with refactoring, there's a question of when to stop (i.e. when to merge this PR and leave further refactoring for future improvement work). We're aware that this PR is already a marked improvement in maintainability over Scenarios, and not all of the code we're reviewing is original (and so should be held to the most current quality standards afforded by the team's skill base), some of it was pieced together from older code and can be "grandfathered in".

One useful stopping criterion is "when we understand how it all works". Both @pblayo and I looked over the builder code and agreed that we had only the foggiest of clues how it worked. Based on that, I'd like to continue for a bit.

`hydrate_entity` violates Demeter, has unobvious temporal coupling (you have to call it for persons entities first, then group entities), works by side-effects and raises exceptions, it feels like a good candidate for some additional work.

Morendil:
We inlined `split_variables_and_roles_json` and `init_entity_members` neither of which had a unit test, and which looked like they had some overlap of responsibilities (in particular via `roles_json`).

We then looked for something worth extracting and unit testing and decided on `check_persons_to_allocate`. This is a consistency check and works purely by side-effects, with inputs that could be cleanly isolated. There are too many of them but we can improve that, in particular `entity_plural` and `persons_plural` are very stable, entities-related properties of the TaxBenefitSystem that we could for instance store in SimulationBuilder (reinforcing the earlier intuition that the part of TaxBenefitSystem we're interested in is the part that manages Entities and their names).

It looks like a good next step would be to explicitly split `hydrate_entity` into `hydrate_person_entity` and `hydrate_group_entity`, so that the latter consumes the **person ids** produced by the former. This would make the temporal coupling explicit rather than implicit and ease unit testing of each variant.

Morendil:
@fpagnoux Please warn before adding commits, I keep odd hours ;)

openfisca_core/simulation_builder.py
```diff
@@ -0,0 +1,360 @@
...
+from openfisca_core.errors import VariableNotFound, SituationParsingError, PeriodMismatchError
+from openfisca_core.indexed_enums import Enum
+from openfisca_core.periods import key_period_size, period as make_period
+from openfisca_core.simulations import Simulation
+from openfisca_core.tools import eval_expression
+
+
+class SimulationBuilder(object):
+
+    def build_from_dict(self, tax_benefit_system, input_dict, default_period = None, **kwargs):
```
fpagnoux: @Morendil @pblayo if `SimulationBuilder ` doesn't have a state anymore, is there any point to keep the `self` here? Shouldn't we just make these methods static with `@staticmethod`?

Morendil: It will have a state, eventually. Builders generally do; their state tends to be the elements of partially constructed objects prior to assembly, and ancillary things. It already looks like the "ancillary things" here will include a mapping between singular and plural names of entities (to be named, perhaps EntityDictionary).

openfisca_web_api/handlers.py
```diff
@@ -6,13 +6,13 @@
...
 
-from openfisca_core.simulations import Simulation
+from openfisca_core.simulation_builder import SimulationBuilder
 from openfisca_core.indexed_enums import Enum
 from openfisca_core.commons import to_unicode
 
 
 def calculate(tax_benefit_system, input_data):
-    simulation = Simulation(tax_benefit_system = tax_benefit_system, simulation_json = input_data)
+    simulation = SimulationBuilder().build_from_entities(tax_benefit_system, input_data)
```
Morendil: Good catch, thanks! I'm puzzled, why wasn't that caught by the web API tests?

Morendil: Oh, never mind. We went from using the old API to using the SimulationBuilder, not from one form of SimulationBuilder to another. The tests will break in due time when we continue refactoring ;)

Morendil:
This latest commit
- introduces a transitional adapter so that tests relying on Scenarios now use SimulationBuilder behind the scenes
- adapts some remaining tests to format changes (dicts vs arrays mostly)
- removes the tests for `axes` (a feature of scenarios), `columns` and `scenarios`
- removes `scenarios` and `columns` altogether, since that was the point of this PR in the first place.

I've attempted to remove `conv` as well but it turns out `periods.py` needs it, alas. It looks like it shouldn't be too hard to remove.

However I might have been too ambitious, cf. the test failure on Python 2. I'll try to figure it out, and rebase to remove that commit if I can't.

Morendil:
>  It looks like it shouldn't be too hard to remove.

Not hard at all, the two functions that used it were dead code.

fpagnoux:
The point was not make sure the last line was also included :)

Morendil:
See also #540

openfisca_core/entities.py
```diff
@@ -240,12 +134,15 @@ def filled_array(self, value, dtype = None):
             warnings.simplefilter("ignore")
             return np.full(self.count, value, dtype)
 
+    def get_variable(self, variable_name, check_existence = False):
```
fpagnoux: > What are the benefits of extracting this to a method? 
> This saves 30 characters on two calls to `self.simulation.tax_benefit_system.get_variable`.

openfisca_core/taxbenefitsystems.py
```diff
@@ -401,3 +402,23 @@ def get_variables(self, entity = None):
...
+            if key not in ('parameters', '_parameters_at_instant_cache', 'variables', 'open_api_config'):
+                new_dict[key] = value
+
+        new_dict['parameters'] = self.parameters.clone()
+        new_dict['_parameters_at_instant_cache'] = {}
+        new_dict['variables'] = self.variables.copy()
+        new_dict['open_api_config'] = self.open_api_config.copy()
+        return new
+
+    def entities_plural(self):
```
fpagnoux: > I don't really know what to think of the multiplications of tiny methods like this. 
> If you have some references, I'd like to read more about the benefits of this practice

openfisca_core/entities.py
fpagnoux: Ok got it, that's to be able to mock this method in tests

fpagnoux:
Test coverage of the new simulation builder:

`py.test --cov=openfisca_core tests/` gives:

```
/Users/florianpagnoux/Dev/openfisca/openfisca-core/openfisca_core/simulation_builder.py                                                 199      6    97%
```

openfisca_core/taxbenefitsystems.py
pblayo: @fpagnoux : It's a consequence of the application of the Demeter's law, to access only the objects that are directly accessible from each class.

maukoquiroga:
Sorry arriving a bit late to the party.

> As usual with refactoring, there's a question of when to stop. One useful stopping criterion is "when we understand how it all works".

I think of two additional questions :

- What's is the value perceived by our users?
- What's the smaller amount improvement that maximises that value that we can deliver?

This allows us to deploy an incremental strategy for improvement, so users can benefit from the value earlier, often and along the whole life-time of the improvement initiative.

"When we understand how it all works" may be too severe as a criterion when working with legacy systems, as it'll delay perfectly shippable improvements and can have a demoralising effect for contributors.

CHANGELOG.md
```diff
@@ -1,5 +1,121 @@
...
+          birth: 2002-01-15
+    output:
+      housing_tax:
+        2017: 1000
+  ```
+
+  A **migration script** is available to automatically convert tests:
+
+  ```sh
+    python openfisca_core/scripts/migrations/v24_to_25.py /path/to/tests/
```
maukoquiroga: No need for tab there.

CHANGELOG.md
```diff
@@ -1,5 +1,121 @@
...
+  ```
+
+  > Note for country package using Scenarios (e.g. France, Tunisia):
```
maukoquiroga: > package

packages

CHANGELOG.md
maukoquiroga: `openfisca test`!

openfisca_core/holders.py
```diff
@@ -183,22 +182,31 @@ def set_input(self, period, array):
                 warning_message,
                 Warning
                 )
+        if self.variable.value_type in (float, int) and isinstance(array, str):
+            array = eval_expression(array)
```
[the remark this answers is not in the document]
fpagnoux: `tools` is actually part of `openfisca_core` (bad naming for sure).

Do you want it to be moved to `commons.py` instead?

CHANGELOG.md
```diff
@@ -1,5 +1,121 @@
 # Changelog
 
+# 25.0 [#781](https://github.com/openfisca/openfisca-core/pull/781)
+
+#### Breaking changes
+
+  - Change the syntax of OpenFisca YAML tests
+
+  For instance, a test that was using the `input_variable` keyword like:
```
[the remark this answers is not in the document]
fpagnoux: Some tests using the `input_variable` were on more than one person, so I'd prefer not give the false impression that this is the only case.

openfisca_core/entities.py
```diff
@@ -32,111 +25,12 @@ class Entity(object):
...
-                        None)
-                    if culprit_path:
-                        path = [self.plural] + culprit_path[0].split('/')
-                    else:
-                        path = [self.plural]  # Fallback: if we can't find the culprit, just set the error at the entities level
-
-                    raise SituationParsingError(path, e.message)
+        self.count = 0
+        self.ids = []
+        self.step_size = 0
```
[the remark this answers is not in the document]
fpagnoux: These attributes are not introduced by this PR, and this PR has IMHO been delayed too long for opportunistic out-of-scope improvements.

openfisca_core/entities.py
Morendil: It makes sense on its own too. `self.simulation.tax_benefit_system.get_variable ` is in technical parlance a [train wreck](http://wiki.c2.com/?TrainWreck). At some later point we'll want to have a `get_variable()` on `Simulation`, which itself delegates to `TaxBenefitSystem`. Having only one call to fix will make this easier.

openfisca_core/taxbenefitsystems.py
Morendil: Yes.

There's a further benefit too: as we prevent SimulationBuilder from directly accessing the internal implementation of TaxBenefitSystem, we make clear what the contract is for the dependency between SimulationBuilder and TaxBenefitSystem.

Eventually we'll notice that whenever we call `tax_benefit_system.something()` the something() has to do with the singular or plural of entities. The sum of these `something()`s defines an interface. This interface is all that we have to pass to SimulationBuilder rather than the entire TaxBenefitSystem. (The TaxBenefitSystem is needed by Simulation, but it could be passed to it by a different route.)

Suppose we call this interface EntityDictionary or something like that. Then to test SimulationBuilder we don't have to instantiate a full TaxBenefitSystem. We can instantiate the much smaller EntityDictionary. And we can instantiate it as a small fixture internal to the test, rather than having to take the entire country template. We get smaller, faster, more decoupled tests.

But the benefit also extends to the objects - we now have a way of assembling a TaxBenefitSystem in smaller steps: first we build an EntityDictionary, then we do the rest. Of course "the rest" can probably be split up as well. TaxBenefitSystem could be thought of as the combination of a VariableRepository, a ParameterRepository and possibly other things. 

All of these smaller objects are easier to unit test. We don't need to load the variables when we want to test parameters, and vice versa.

References: Martin Fowler's _Refactoring_ (2nd edition just came out), Bob Martin's _Clean Code_, Michael Feather's _Working Effectively With Legacy Code_, Sandi Metz's _Practical Object Oriented Design_.

Morendil:
We still need to revert the commits marked "to revert", yes ?

openfisca_core/taxbenefitsystems.py
fpagnoux: Ok, thanks a lot for taking the time to explain and for the references!

fpagnoux:
> We still need to revert the commits marked "to revert", yes ?

Done!

CHANGELOG.md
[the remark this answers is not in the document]
maukoquiroga: Is it just a typo? `input_variable` => `input_variables`

setup.py
```diff
@@ -31,13 +33,13 @@
     'flake8 >= 3.5.0, < 3.6.0',
     'pycodestyle >= 2.3.0, < 2.4.0',  # To avoid incompatibility with flake8
     'pytest < 4.0.0',
-    'openfisca-country-template >= 3.4.0, < 4.0.0',
-    'openfisca-extension-template >= 1.1.3, < 2.0.0',
+    'openfisca-country-template >= 3.6.0rc0, < 4.0.0',
```
maukoquiroga: Do not forget to remove the release candidate.

PR #784, opened 2018-11-30
Simulation delete arrays

### LINKED ISSUE #740: Improve input deleting arrays in Python
What I have to do:

````
simulation = ...
holder = simulation.persons.get_holder('salary')
holder.delete_arrays('2018-08')
````

What I'd prefer:

````
simulation = ...
simulation.delete_arrays('salary', '2018-08')
````
What do you folks think?

benjello:
Fix #740 

@maukoquiroga @fpagnoux : i also tried to make the doctest work for simulation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Changelog
 
+## 24.10.0 [#784](https://github.com/openfisca/openfisca-core/pull/784)
+
+- In Python, simplify simulation array deletion:
+
+Before:
+
+```py
+simulation = ...
+holder = simulation.persons.get_holder('salary')
+holder.delete_arrays('2018-08')
+```
+
+After:
+
+```py
+simulation = ...
+simulation.delete_arrays('salary', '2018-08')
+```
+
 ### 24.9.9 [#786](https://github.com/openfisca/openfisca-core/pull/786)
 
 - Set a 120s default timeout in `openfisca serve`
--- a/openfisca_core/simulations.py
+++ b/openfisca_core/simulations.py
@@ -428,6 +428,22 @@ class Simulation(object):
 
     # ----- Misc ----- #
 
+    def delete_arrays(self, variable, period):
+        """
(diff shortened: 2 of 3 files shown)


benjello:
@maukoquiroga @fpagnoux @Morendil : I can merge this PR but do you know if doctests are executed in tests. It seems that it is not the case because the test passed but the doctest was failing locally.

Morendil:
It's [supported](https://docs.pytest.org/en/latest/doctest.html) and seems enabled via `setup.cfg`. What do you get when running `make test` locally ?

benjello:
Ok maybe the tests where not ran at all. Not expecting results

Morendil:
OK, checked locally by introducing a deliberate error in doctests for `periods.py` and they are in fact not run in CI, thanks for catching that!

benjello:
@Morendil : so should I rebase, bump and merge this Pr or should I wait for a fix ?

Morendil:
The right command to run doctests would be `pytest --doctest-modules openfisca_core`, except that fails locally with a bunch of problems:
- `openfisca_core/scripts/find_placeholders.py` needs `bs4`
- `openfisca_core/scripts/measure_performances.py` fails with `Missing attribute 'definition_period' in definition of variable 'age_en_mois'` 
- `openfisca_core/scripts/measure_performances_fancy_indexing.py` depends on `openfisca_france` (!)
- `openfisca_core/scripts/remove_fuzzy.py` has an assertion failure
- `openfisca_core/scripts/migrations/v16_2_to_v17/xml_to_yaml.py` fails due to lacking `lxml`
- `openfisca_core/scripts/migrations/v16_2_to_v17/xml_to_yaml_country_template.py` and `openfisca_core/scripts/migrations/v16_2_to_v17/xml_to_yaml_extension_template.py` ditto
- `openfisca_core/conv.py` because PyTest mistakes `test_in_pop` for a test function
- …and probably more (I gave up here)

Issue #667, opened 2018-05-19
Don't force persons to be assigned to all defined entities


jacob88:
Hi there!

I really enjoy OpenFisca, but I recently encountered an issue.

### Here is what I did:
Declared "Bob" in Persons without assigning "Bob" to a titled_property,
through the JSON web API.

```JSON
{
	"persons": {
		"Bob": {
			"sum_of_earnings_in_last_52_weeks": {
				"2018": 1200
			},
			"earnings_period_in_weeks": {
				"2018": 52
			},
			"weekly_compensation_before_tax": {
				"2018": null
			}
		}
	}
}
```

### Here is what I expected to happen:
I expected the API to return "200 OK" with the null field filled with the result

```JSON
{
	"persons": {
		"Bob": {
			"earnings_period_in_weeks": {
				"2018": 52
			},
			"sum_of_earnings_in_last_52_weeks": {
				"2018": 1200
			},
			"weekly_compensation_before_tax": {
				"2018": 23.076923370361328
			}
		}
	}
}
```

### Here is what actually happened:
```JSON
{
  "titled_properties": "Bob has been declared in persons, but is not a member of any titled_property. All persons must be allocated to a titled_property."
}
```

### Here is data (or links to it) that can help you reproduce this issue:

Required an additional titled property.

Below works fine.
```JSON
{
	"persons": {
		"Bob": {
			"sum_of_earnings_in_last_52_weeks": {
				"2018": 1200
			},
			"earnings_period_in_weeks": {
				"2018": 52
			},
			"weekly_compensation_before_tax": {
				"2018": null
			}
		}
	},
	"titled_properties": {
		"donors": {
			"owners": [
			"Bob"
			]
		}
	}
}
```

## Context

I identify more as a:

- Developer _(I create tools that use the existing OpenFisca code)_.

(remove this line as well as all items in the list that don't fit your situation)

MattiSG:
As an additional bit of context for other readers: this is in the context of https://github.com/ServiceInnovationLab/openfisca-aotearoa, with [these entities](https://github.com/ServiceInnovationLab/openfisca-aotearoa/blob/d280357660fe8cbf9cc04300cb1b494010939239/openfisca_aotearoa/entities.py).

MattiSG:
Thank you so much @jacob88 for this detailed issue!

[...]

PR #785, opened 2018-11-30
 Don't force persons to belong to all entities

fpagnoux:
Connected to #667 
Depends on #781 

Following our discussions during the webinar about entities, I made a quick POC to check how we could stop enforcing persons to belong to all entities, as it's a common frustration among users.

The chosen behavior A in this PR here is:

- Allow a person to not be defined in any household
- If `person.household` is used and _at least one_ person doesn't have a household, an error is raised.

As a reminder, the alternative behavior B would be:

- Allow a person to not be defined in any household
- Still allow to call `person.household` in that case
- Make `person.household('variable', period)` return the `'variable'` default value for person who don't belong to any household.

I think there was a preference for behavior A during the webinar, let me know if you don't agree.

I tried to see how much we could use this to avoid boilerplate in France tests, and in some cases, [we can](https://github.com/openfisca/openfisca-france/compare/unif-tests-api...flexible-entities). In others though, we get an error, even if the test doesn't seem to be related to that specific entity (typically, we need the zipcode of a "famille", which is defined for a "menage").

I'd say that when trying to remove entity specifications that looked like boilerplate, I was successful in maybe 1/3 of the cases.  I'd expect this to work better in countries were the entities are not as coupled as the French ones.

Note:
- I'm opening against the `uniformize-tests-api` branch to avoid conflicts
- This is not ready to merge as is, see todo further down
- It could be interesting to merge this before https://github.com/openfisca/openfisca-france/pull/1224, as it would allow us to reduce the amount of boilerplate noise that we add to France tests (cc @Morendil)

I'm planning to keep working on this if contributors who expressed frustrations feel that this approach is the right one cc @jvalduvieco @jacob88.

Inputs from @guillett @benjello @sandcha @maukoquiroga @pblayo welcome as well :).

Todo if we keep going toward that direction:

- [ ] Make sure aggregators like `sum` still work when some people don't belong to an entity
- [ ] Intercept `UndefinedEntityError` when they happen during a Web API call to return a 400 instead of a 500, with information about which formula tried to access the undefined entity.

### DIFF
--- a/openfisca_core/entities.py
+++ b/openfisca_core/entities.py
@@ -10,6 +10,7 @@ import numpy as np
 
 from openfisca_core.indexed_enums import EnumArray
 from openfisca_core.holders import Holder
+from openfisca_core.errors import UndefinedEntityError
 
 ADD = 'add'
 DIVIDE = 'divide'
@@ -279,6 +280,7 @@ class GroupEntity(Entity):
         self.members = self.simulation.persons
         self._roles_count = None
         self._ordered_members_map = None
+        self._is_defined_for_all_persons = None
 
     @property
     def members_role(self):
@@ -343,6 +345,12 @@ class GroupEntity(Entity):
         if self._ordered_members_map is None:
             return np.argsort(self.members_entity_id)
         return self._ordered_members_map
+
+    @property
+    def is_defined_for_all_persons(self):
+        if self._is_defined_for_all_persons is None:
+            self._is_defined_for_all_persons = (self.members_entity_id >= 0).all()
+        return self._is_defined_for_all_persons
 
     #  Aggregation persons -> entity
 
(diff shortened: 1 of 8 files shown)


maukoquiroga:
Rebased.

Br3nda:
Thank you for this! When it's ready i'd like to test against Aotearoa NZ. We have both families and properties that people own/live in.

Morendil:
Having had time to think about this, I'd be much happier with an option C - if a person doesn't belong to any group entity, create a group entity just for them.

My rationale is that option A is only pushing the problem further back - in a densely connected model (France being the obvious example) you will be able to create a simulation, but the first attempt to calculate anything in it will fail with an error. I don't think this is desirable. I'm not entirely sure I understand Option B above, but I suspect it would also yield non-intuitive results.

Morendil:
Rebased and removed the commit modifying Changelog; I find it easier to deal with these just before merging. As a reminder:
- [ ] correct version numbering of 25.0 to 25.0.0 before merging

sandcha:
Given [country-template entities](https://github.com/openfisca/country-template/blob/ca782a924295e5278fa9236e46fc447ff2011a8c/openfisca_country_template/entities.py), what I understand from C is: if we define a simulation with Person 🍋  and Person 🍅  and no group entity affiliation, openfisca will create a Household 🍋 containing Person 🍋  only and an other Household 🍅 containing Person 🍅 only.
I prefer this C option to A or B.

Morendil:
Closing in favor of #835

PR #786, opened 2018-11-30
Add default timeout

fpagnoux:
As suggested by @alexsegura

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+### 24.9.9 [#780](https://github.com/openfisca/openfisca-core/pull/780)
+
+- Set a 120s default timeout in OpenFisca Serve
+  - Otherwise, they regularly restart with an anxiety-provoking "[CRITICAL] WORKER TIMEOUT (pid:2137)" message in log
+
 ### 24.9.8 [#780](https://github.com/openfisca/openfisca-core/pull/780)
 
 - Allow non-integral inputs to int variables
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+### 24.9.9 [#786](https://github.com/openfisca/openfisca-core/pull/786)
+
+- Set a 120s default timeout in `openfisca serve`
+  - Otherwise, they regularly restart with an anxiety-provoking "[CRITICAL] WORKER TIMEOUT (pid:2137)" message in log
```
Morendil: What will happen instead now?

alexsegura:
Just as an additional note: previously, on Mes Aides, the timeout was set to `60`. 
But now, it is not sufficient anymore. 
It could be great to try to understand what causes the increase in boot time. 
Maybe a bunch of YAML files added recently?

CHANGELOG.md
fpagnoux: They should now stay awake and no reboot for no reason

fpagnoux:
> It could be great to try to understand what causes the increase in boot time.
Maybe a bunch of YAML files added recently?

Is the timeout issue really related to the booting time 🤔 ? I had it on the country template, which is really fast to start. 

But if you noticed a slower API boot, I think https://github.com/openfisca/openfisca-core/pull/744 actually had some impact, as the API runs a first calculation when it starts.

guillett:
capture d ecran de 2018-12-03 21-34-27

Here is OpenFisca-Frnace ong booting time explanation.

fpagnoux:
If booting time is important, there are a lot of things that could be loader only on demand ;)

guillett:
To me, it is really hard to _sell_ OpenFisca if the booting time is 20s.

PR #787, opened 2018-12-03
Don't sort JSON keys in the Web API

fpagnoux:
By default, FLASK sorts the JSON keys of the API responses. This PR opts out of this behavior.

~~In some use-cases (merge with IPP parameters in France), having the same orders in API responses that in the parameter YAML files avoid extra work and boilerplate.~~

---

Edit by @maukoquiroga 2018-12-09:

I'm ok with the fact that we don't order an object defined as unordered by the JSON specification. However, as discussed in the comments, I omitted any mention of use-case from the motivation of this PR's introduced changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 
-# 25.0 [#781](https://github.com/openfisca/openfisca-core/pull/781)
+## 25.1.0 [#787](https://github.com/openfisca/openfisca-core/pull/787)
+
+- Don't sort JSON keys in the Web API
+  - By default, `flask` sorts JSON _object_ keys alphabetically
+  - This is useful for reusing http caches for example, at the cost of some performance overhead
+  - But the [JSON specification](https://www.json.org/) doesn’t require to keep an order, as _objects_ are defined as "an unordered set of name/value pairs"
+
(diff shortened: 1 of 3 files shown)


fpagnoux:
@Morendil this should allow to avoid the `order` keys in YAML parameters file.

fpagnoux:
> Or, as I understand it, the use case described here implies that a client will be relying on a specific order of the served JSON keys.
> That means that, from now on, any change to this behaviour will have to be considered de facto as a breaking-change.

Well, not really. If a client of your library decide to use an internal method of a class, it doesn't mean that this method becomes part of your interface. It means that your clients are taking a risk, and that's on them ;).

I think NodeJS de facto preserves dictionary insertion order. That's however not part of their API contract. Some people rely on that, even if it's not advised. That's a risk they are taking. Node warns against doing that, but Node doesn't shuffle the dictionary order just for the sake of it. This is what we are doing right now, and this PR just suggests to stop doing so, without any promise to the API user.

I haven't written the changelog yet, but I'd phrase it like "Do not alphabetically sort JSON keys in Web API response". Once again, no promise, and no contract.

Does this address your concern ?

maukoquiroga:
@fpagnoux I edited the description

Issue #560, opened 2017-08-22
Simplify entity definition


fpagnoux:
Suggestions:

- Remove and deprecate subroles
- Assume an entity describes persons if no role is defined.
- Add a `min` property
- Assume a role is unique if it has no plural
- Remove label properties if they are not used

MattiSG:
@benjello We think subroles should be split into flat roles to avoid nesting. Opinion?

benjello:
To be sure that I understand you well, you want to switch from:
 - family:
    - parents:
      - first_parent
      - second_parent
   - children
      - child
      - child 

to 

- family:
  - first_parent
  - second_parent
  - child
  - child

?

MattiSG:
Not really: if Family is an entity and `parent` and `child` are roles, then we'd have the hierarchy you described in the first example. Subroles apparently refine roles, as in “mother” being a possible subrole for “parent”. That adds a lot of complexity, and the suggested change would be to replace these by `parent_mother` in this example.

fpagnoux:
Some context about why subroles even came to exist in the first place:

OpenFisca France has 3 group entities: `famille`, `foyer_fiscal` and `menage`. These entities have  a common structure: one or two people that are "in charge" (typically the parents), and other persons that gravitate around them (typically the children).

One important modelling choice is: do these two people "in charge"  have symmetrical roles? Or is there a clear distinction between them? 

For `menage`, the answer has always been clear for france: the two roles are distinct. I believe this is because INSEE historically loved this idea of "Chef de famille", or "Personne de référence" 👨‍💼 .

For `famille` and `foyer_fiscal`, I'm talking under the control of @benjello but I believe OpenFisca France leant more towards symmetric roles. The way to declare a `famille` has always been something like `{parents: [Alice, Bob], children: [Kevin]}`. Because, _in the model_, there is no reason to distinguish the two parents.

However, practically, indistinguishable parents were challenging in two situations. Both related to projecting values from one group entity to another. This is where subroles came to help.

1. I'm writing a formula for a variable `benefit`, defined for a `famille`. I need to access a variable 
 `code_postal`, defined for a `menage`. But which one of the `famille`'s member `menage` should I use? Having a subrole pointing to _one person_ helped. I could just use: `famille.demandeur.menage('code_postal', period)`, with the convention that the first `parent` of a `famille` has the subrole `demandeur`.

[...]

Issue #782, opened 2018-11-26
Improve error messages for entity role issues


guillett:
Hi there!
I really enjoy OpenFisca, but I recently encountered an issue.

### Here is what I did:

I generated a relatively big payload with 3000 families and got an error because I associate more than two people to the parent role in a family.

The error is also visible with that payload:
```json
{
    "individus": {
        "parent1": {
            "age": {
                "2018-10": null
            }
        },
        "parent2": {},
        "parent3": {}
    },
    "familles": {
        "_": {
            "parents": [
                "parent1",
                "parent2",
                "parent3"
            ]
        }
    },
    "foyers_fiscaux": {
        "_": {
            "declarants": [
                "parent1",
                "parent2",
                "parent3"
            ]
        }
    },
    "menages": {
        "_": {
            "personne_de_reference": [
                "parent1"
            ],
            "conjoint": [
                "parent2",
                "parent3"
            ]
        }
    }
}
```
### Here is what I expected to happen:
I expected to have a error message that explicit the bogus family of group entity.

### Here is what actually happened:
{"error":"Internal server error: list index out of range"} 

## Context

I identify more as a:

- Analyst _(I make macroscopic computations on real populations)_.
- Developer _(I create tools that use the existing OpenFisca code)_.

(remove this line as well as all items in the list that don't fit your situation)

Morendil:
@fpagnoux As I look into this I notice that `iter_over_entity_members` is in `scenarios.py`, it doesn't use `conv` so we might want to move it somewhere else to isolate further the dependency on scenarios.

There is indeed no check for an out of bounds index over the `subroles` array in [Famille](https://github.com/openfisca/openfisca-france/blob/02439030d7024af1a229f38780f8f30ec082e1d2/openfisca_france/entities.py#L5): https://github.com/openfisca/openfisca-core/blob/56e31c6186e1e354d074ee1e6f4f760442b44981/openfisca_core/scenarios.py#L726

A quick and dirty fix is to change the line above to silently ignore the implied extra subroles and (as a consequence) only assign the individuals to the "parent" role:
```
if role.subroles and legacy_role_j < len(role.subroles)
```

I'm not sure how much effort is required for a "proper" test-driven fix (possibly some untangling of dependencies between classes).

[...]

PR #789, opened 2018-12-03
Deprecate subroles

fpagnoux:
Connected to #560 

- Deprecate subroles, and introduce a new syntax to replace them:
  - For instance, if the role `parent` had two subroles `first_parent` and `second_parent`:
    - `household.first_parent('salary', period)` can be replaced by `household.parents[0]('salary', period)`
    - `person.has_role(Household.FIRST_PARENT)` can be replaced by `person.has_role(Household.PARENT, 0)`
    - `family.sum(salary, role = Household.FIRST_PARENT)` can be replaced by `family.sum(salary, role = (Household.PARENT, 0))`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 24.11.0 [#789](https://github.com/openfisca/openfisca-core/pull/789)
+
+- Deprecate subroles, and introduce a new syntax to replace them:
+  - For instance, if the role `parent` had two subroles `first_parent` and `second_parent`:
+    - `household.first_parent('salary', period)` can be replaced by `household.parents[0]('salary', period)`
+    - `person.has_role(Household.FIRST_PARENT)` can be replaced by `person.has_role(Household.PARENT, 0)`
+    - `family.sum(salary, role = Household.FIRST_PARENT)` can be replaced by `family.sum(salary, role = (Household.PARENT, 0))`
+
 ## 24.10.0 [#784](https://github.com/openfisca/openfisca-core/pull/784)
 
 - In Python, simplify simulation array deletion:
--- a/openfisca_core/entities.py
+++ b/openfisca_core/entities.py
@@ -804,7 +804,7 @@ class Projector(object):
             return self.parent.transform_and_bubble_up(transformed_result)
 
     def transform(self, result):
-        return NotImplementedError()
+        raise NotImplementedError
 
 
 # For instance person.family
@@ -856,7 +856,12 @@ class MultipleRoleToEntityProjector(Projector):
         return RoleAndPositionProjector(self.target_entity, self.role, key, self.parent)
 
     def transform(self, result):
-        return result * self.reference_entity.has_role(self.role)
+        error_message = "You cannot use '{entity}.{role_plural}' to calculate a variable, as '{role_key}' is not a unique role. However, you can for instance calculate the variable for the 1st '{role_key}' with '{entity}.{role_plural}[0]'".format(
+            entity = self.target_entity.key,
+            role_key = self.role.key,
+            role_plural = self.role.plural
+            )
+        raise NotImplementedError(error_message)  # TODO: Add a nice error message
 
 
 # For instance household.parents[0]
--- a/openfisca_core/entities.py
+++ b/openfisca_core/entities.py
@@ -923,6 +923,10 @@ def build_entity(key, plural, label, doc = "", roles = None, is_person = False):
             entity_class.roles.append(role)
             setattr(entity_class, role.key.upper(), role)
             if role_description.get('subroles'):
+                warnings.warn(
+                    "Entities subroles are deprecated since OpenFisca Core 24.11.0, and will be removed in the future. Check that version's Changelog for help migrating.",
+                    Warning
+                    )
                 role.subroles = []
                 for subrole_key in role_description['subroles']:
                     subrole = Role({'key': subrole_key, 'max': 1}, entity_class, parent_role = role)
--- a/openfisca_core/entities.py
+++ b/openfisca_core/entities.py
@@ -191,6 +191,8 @@ class Entity(object):
                 array, self.key, array.size, self.count))
 
     def check_role_validity(self, role):
+        if isinstance(role, tuple):
+            role, _ = role
         if role is not None and not type(role) == Role:
             raise ValueError("{} is not a valid role".format(role))
 
@@ -287,7 +289,7 @@ class PersonEntity(Entity):
     # Projection person -> person
 
     @projectable
-    def has_role(self, role):
+    def has_role(self, role, position = None):
         """
             Check if a person has a given role within its :any:`GroupEntity`
 
@@ -297,11 +299,18 @@ class PersonEntity(Entity):
             >>> array([False])
         """
         self.check_role_validity(role)
+        if isinstance(role, tuple):
+            role, position = role
         entity = self.simulation.get_entity(role.entity_class)
-        if role.subroles:
-            return np.logical_or.reduce([entity.members_role == subrole for subrole in role.subroles])
+
+        if position is None:
+            position_filter = True
         else:
-            return entity.members_role == role
+            position_filter = entity.members_position_by_role == position
+        if role.subroles:
+            return np.logical_or.reduce([entity.members_role == subrole for subrole in role.subroles]) * position_filter
+        else:
+            return (entity.members_role == role) * position_filter
 
     @projectable
     def value_from_partner(self, array, entity, role):
--- a/openfisca_core/entities.py
+++ b/openfisca_core/entities.py
@@ -380,6 +380,7 @@ class GroupEntity(Entity):
             self._members_role = None
             self._members_position = None
             self.members_legacy_role = None
+            self.members_position_by_role = None
         self.members = self.simulation.persons
         self._roles_count = None
         self._ordered_members_map = None
@@ -406,8 +407,12 @@ class GroupEntity(Entity):
             self.simulation.persons.count,
             dtype = np.int32
             )
-        self._members_position = None
+        self.members_position_by_role = np.empty(
+            self.simulation.persons.count,
+            dtype = np.int32
+            )
 
+        self._members_position = None
         self.persons_to_allocate = set(self.simulation.persons.ids)
 
         Entity.init_from_json(self, entities_json)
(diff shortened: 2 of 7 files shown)


fpagnoux:
I had this on an experimental branch, and #782 seems to indicated once more time that we should move away from subroles.

Morendil:
I'm rather 👎 on the `household.parents[0]('salary', period)` syntax. The previous syntax was entity-like, this one violates the principle of least surprise, the `[0]` strikes me as a serious stumbling block. The other two syntax additions do not strike me as very fortunate either.

I'm also not super happy with a net +200LOC PR, where the path to removing complexity later is not entirely clear.

Finally, I'm not sure I understand what problem, precisely, is being solved.

A few stats from France:
- for the "demandeur" subrole, out of 91 uses
  - 65 are part of famille.demandeur.menage
  - 8 are part of famille.demandeur.foyer_fiscal
  - 16 are direct uses
  - 2 are `has_role` checks
- for the "conjoint" subrole, out of 14 uses,
  - 9 are direct
  - 2 are `has_role` checks for DECLARANT_PRINCIPAL
  - 2 are famille.conjoint.foyer_fiscal
- there are 5 uses of `foyer_fiscal.declarant_principal`, all in reforms
- there are 12 direct uses of `foyer_fiscal.conjoint`, all but one in `ir.py`
- `famille.sum` is *never* used with a subrole (out of 33 sums with roles)
- `foyer_fiscal.sum` is *never* used with a subrole (out of 5 sums with a role)
- `menage.sum` is used with declarant_principal in all but 1 of 20 uses

One of the `foyer_fiscal.sum` with a role filter uses a Menage role (`isf.py`:780), which can't be correct; I see this as a weak signal that role filters confuse model authors.

I don't necessarily have a better solution having said all this, but it seems to me that subroles have been used in France in heterogenous ways, a lot of which are *not* attempting to pick out a distinguished invididual but merely to navigate to a different group entity.

Just for the exercise of coming up with an alternative, we can go the other way: the roles for which we use subroles are usually couples, so we could instead support "super-roles". In the Famille entity have "demandeur" and "conjoint" as main roles, with "parent" having the meaning "demandeur OR conjoint".  Having the sub-roles as "main" roles seems to work well enough for Menage.

But again, it's hard to know if this alternative is better, because I don't yet understand how this solves a  problem, and it seems to me that the syntax changes potentially introduce new problems.

fpagnoux:
> Finally, I'm not sure I understand what problem, precisely, is being solved.

More details on the discussion about subroles: https://github.com/openfisca/openfisca-core/issues/560

The main issue is that the concept of subrole is not very straightforward: subroles are a way to explicitly name the n (practically, 2) persons with a given role like "parent". In my API request, I only declare "parents", but in OpenFisca  individuals will have roles like "demandeur" or "conjoint". That can be surprising. 

> One of the foyer_fiscal.sum with a role filter uses a Menage role (isf.py:780), which can't be correct; I see this as a weak signal that role filters confuse model authors.

This is actually correct, and done quite often (for instance [here](https://github.com/openfisca/openfisca-france/blob/master/openfisca_france/model/mesures.py#L316-L319)) when "ventilating" a variable defined on foyer fiscal to menages. The idea is to make sure that for one foyer fiscal, the amount of a variable `revenu_fiscal` is allocated for _one and only one_ menage, and that involves a sum over the menage members who are  head of their foyer fiscal. 

> Just for the exercise of coming up with an alternative, we can go the other way: the roles for which we use subroles are usually couples, so we could instead support "super-roles". In the Famille entity have "demandeur" and "conjoint" as main roles, with "parent" having the meaning "demandeur OR conjoint". Having the sub-roles as "main" roles seems to work well enough for Menage.

The more realistic alternative that would actually reduce complexity would be to just have a `demandeur` and a `conjoint` role, with no `parent` role. The underlying philosophy would be that _if you ever need to differentiate two persons, then these two persons must have a different role_. Or rephrase, _you can never treat someone individually unless they have a unique role_.

> a lot of which are not attempting to pick out a distinguished invididual but merely to navigate to a different group entity.

But, with our very flexible model of a menage, famille and foyer fiscal with no assumption of relationship between them, what would you propose other than distinguish individuals to navigate to a different group entity?

Morendil:
> In a API request, I only declare "parents", but in OpenFisca individuals will have roles like "demandeur" or "conjoint". That can be surprising.

Ouch. Yes, that's a problem.

> > One of the foyer_fiscal.sum with a role filter uses a Menage role (isf.py:780), which can't be correct; I see this as a weak signal that role filters confuse model authors.

> This is actually correct

OK, then it's a strong signal that it confuses _me_. ;)

> The idea is to make sure that for one foyer fiscal, the amount of a variable revenu_fiscal is allocated for one and only one menage

I wouldn't have guessed that from looking at the code _or_ the docstring. (This is one area where we could use some UX methodology - notice ambiguities and observe how people interpret things.)

> what would you propose other than distinguish individuals to navigate to a different group entity

I'm not quite sure. How about we set up a list of test cases (not necessarily code, at first a text description will do) where we seem to need it, and see what we come up with?

Morendil:
In cases 1 and 2, would the `first_person` projector is an adequate solution? I don't see why we care at all about the distinguished role "demandeur".

I find it very hard to build a clear mental model of what happens in constructions like case 2, even after poring over the Core source code for a while (it's painful to put myself in the shoes of someone who only wants to use the API). One reason is:
- `famille('benefits', period)` will return an array of count = count(**famille**)
- `menage.members.famille('benefits', period)` an array of count = count(**individus**)

If we didn't have this discrepancy, and famille returned the right size array, we wouldn't have this concern of double counting, right? Would it be worth solving the discrepancy instead?

For case 3, what if some `famille`s have only one `parent` ?

fpagnoux:
> In cases 1 and 2, would the `first_person` projector is an adequate solution? I don't see why we care at all about the distinguished role "demandeur".

@benjello, do you think that would work? What I find not ideal is that:
- we would have a legislation formula explicitly relying on a _technical_ concept that has no _functional_ meaning. The "first person" of a household doesn't make sense from the law point of view, it's just the person that comes arbitrarily first in the person vector.
- That could some some strange behaviors: two datasets with the same data, but in a different order, could result in different simulation outcome
- I'd also see a potential issue if, let's say a 20 year old student belongs to a "foyer fiscal" as a child, but lives in a different "menage" that her parents. If she comes as the "first person", then we would keep her address for the whole foyer fiscal, which is probably not the expected behavior.

I guess that would work better if we made sure `first_person` was always the first person with the first role in the entity 🤔 . But such a mechanism would not be super explicit.

fpagnoux:
> If we didn't have this discrepancy, and famille returned the right size array, we wouldn't have this concern of double counting, right? Would it be worth solving the discrepancy instead?

I don't understand. Which of the two function calls you mentioned would you say return an outcome of an unexpected length? 
I don't see how it is related to double counting. If we want to ventilate a benefit defined to `famille` to `menage`s, we need to find one person in each family to whom we attribute this benefit. We thus need to single out this person in a deterministic way.

fpagnoux:
> For case 3, what if some familles have only one parent ?

`person.famille.parents[1]('salary', period)` would be 0, as `person.famille.conjoint('salary', period)` is also 0 when there is no conjoint.

Morendil:
> I expect it to return a famille-sized result.

Afterthought: in the context of a variable defined on Menage, we need a Menage-sized array, so if there are a different number of Menage and Famille we still need to do a projection in some way.

It still doesn't make sense to me that as a model author I'm expected to make this projection myself, explicitly, with persons as a mediating step and through a role that I do not really care about. If "relying on a technical concept that has no functional meaning" is a valid complaint, then, it seems to me, so is "relying on an unrelated functional concept for purely technical reasons".

From a syntax standpoint, one of the following would make more sense to me:
- `menage('benefits', period)` (and implement a standard projection which makes sense)
- `menage.famille('benefits', period)` to be more explicit about the "casting"
- `menage.famille.select('benefits', period)` to make it even clearer we're narrowing the vector

What really matters though isn't so much the syntax as having a clear concept of what's going on in these constructions. What exactly is a "projection"? The docs are cryptic about that, and the syntax doesn't distinguish between cases where there are more `Famille`s than `Menage`s, versus the other way around - whereas this distinction seems to me like an important one to understand this idea of "projection" and mentally link it up to the vectorial view.

fpagnoux:
> The docs are cryptic about that, and the syntax doesn't distinguish between cases where there are more Familles than Menages, versus the other way around - whereas this distinction seems to me like an important one to understand this idea of "projection" and mentally link it up to the vectorial view.

But this is not _just_ about the size of the vectors. There is _no general relationship_ between "famille" and "menage" in France. A "famille" doesn't necessarily have a "menage", or vice-versa. Even if there are the same numbers of them, it by no means implies that the first menage matches the first family. 

This lack of relationship between entities is often practically inconvenient, but [there seem to be good reasons from economists to maintain that](https://github.com/openfisca/openfisca-france/issues/593) (cc @benjello).

So as much as I would have love to write `famille.menage('rent', period)` or better, `famille('rent', period)` when working on Mes-Aides,  I don't see how we allow that _in Core_ without making strong assumptions about entities.

Morendil:
@fpagnoux Can you turn this into a PR on the main repo, as opposed to a fork as it is now? It's way harder to rebase PRs on a fork. (I'd like to rebase as part of removing the Python 2 build, which has been disabled in the repo's settings.)

sandcha:
Connected to openfisca/openfisca-doc#168

fpagnoux:
I'm going to close the PR, as it seems that the new syntax suggested is not very popular, and we have other things to deal with. 

I still do not think the status quo is ideal either. To summarize the main issues with subroles:

### For users

- Subroles are not documented, and hard to explain
- In a API request, I only declare "parents", but in OpenFisca individuals will have roles like "demandeur" or "conjoint". That can be surprising.
  - Declaring `demandeur` or `conjoint` in the Web API would not work
- When creating a situation from CSV-type datas, it is not clear for users if they should provide the 1-st level role (like `parent`) or the subroles (like `first_parent`)

### For Core maintainers

- Any work involving roles always bring confusion: should I use `household.roles` (`[parent, child]`) or `entity.legacy_role` (`[first_parent, second_parent, child]`)?
- They are a complex mechanism that was tailored-made for France, but is unused by new models (e.g. NZ), and feels like a burden on a contributive common.  

(All, please feel fee to comment or edit this post if you want to add to the list)

Morendil:
Thanks! It looks as if @sandcha and I will be looking into CSVs/entities/SimulationBuilder this week and this should be an opportunity for both of us to get up to speed on roles and subroles.

PR #790, opened 2018-12-04
Improve design and test coverage of complex simulation construction

### LINKED ISSUE #540: Remove scenarios

### LINKED ISSUE #691: Refactor code using Biryani

Morendil:
#### What this PR brings

An exciting but under-documented feature, "axes", now has much better test coverage and thus long-term maintainability (the documentation is still lacking, but see https://github.com/openfisca/tutorial for demos)

#### Breaking changes

This PR deprecates the `new_scenario` approach to constructing Simulation objects. This will impact you if:
- your notebooks or scripts or other Python code rely on the France model and use the old form of creating a Simulation object (see below)
- **or** your country package defines a Scenario class and injects it [the way France does](https://github.com/openfisca/openfisca-france/blob/11b18985ce4decc31b5666114b2525dddf42652b/openfisca_france/france_taxbenefitsystem.py#L29)

**To migrate to this version**, if you are in the first case, the minimum required change is this:

*The old way:*
```
simulation = tax_benefit_system.new_scenario().init_single_entity(...some data...).new_simulation()
```
*The new way:*
```
# At the top of your file
from openfisca_france.scenarios import init_single_entity
# Below
simulation = init_single_entity(tax_benefit_system.new_scenario(), ...some data...).new_simulation()
```

If you are in the latter case, you must also transform your `init_single_entity` from a Scenario method to a regular function at global scope, and change your tests and reuses as described above.

#### More detail on what's changed

- introduce a new implementation and associated tests for axes (see [demo](https://github.com/openfisca/tutorial/blob/b2d0f0a2ddd0658e401fdad4657bdd8aa6c431ae/notebooks/how_to_handle_axes.ipynb))
- introduce a transitional adapter so that tests which relied on Scenarios now use SimulationBuilder behind the scenes
- adapt some remaining tests to format changes (dicts vs arrays mostly)
- remove `scenarios` and `columns` and their tests
- remove decompositions, which relied on `columns` and `conv`
- remove dead code from `periods.py`
- remove `conv.py`
- remove Biryani

-----
Historical note:

Scenarios were provided in Core as an abstract base class. Country packages were expected to extend this class and provide, via the Scenario instance variable of TaxBenefitSystem, the name of the concrete class to be instantiated in new_scenario: https://github.com/openfisca/openfisca-core/blob/08c6fab5c9559283a568f4dbc9e156f52c810f99/openfisca_core/taxbenefitsystems.py#L93-L96

Country packages were expected to override some methods of the base class (`suggest`, `attribute_groupless_persons_to_entities`, `post_process_test_case`) to extend it.

In practice, the main way that the base class was extended was via an undocumented method `init_single_entity` that wasn't part of the interface of the base class, but was still called for instances returned by `new_scenario` - a painful violation of OO design principles.

Contrary to what the name implies, `post_process_test_case` was almost exclusively used for input checking of entity validity constraints, something we should do elsewhere. As for `attribute_groupless_persons_to_entities` we are taking the opposite tack with this PR - fully specifying all entities in test cases.

Axes were tested in Core but otherwise unused in Core; not documented except via examples of use in notebooks; and used in France, mostly to test reforms, in a way that was backported in openfisca/openfisca-france#1224 to pass the existing tests there. The only way to use axes was via `init_single_entity`, this is preserved but will need to make way for a new and more stable API.

----

See also openfisca/openfisca-france#1227

Fixes #540 
Fixes #691


maukoquiroga:
Rebased.

benjello:
Merci de ne pas merger avant mon ok. Je vais utiliser les scénarios dans mes cours dans les 3 prochaines semaines. Je vais tester la nouvelle syntaxe mais je ne voudrais pas qu'une adoption trop rapide casse beaucoup de chose en aval pour un gain minime de réduction des délais.

tests/core/test_axes.py
```diff
@@ -1,182 +1,149 @@
...
+        persons:
+          Alicia: {salary: {2018-11: 0}}
+          Javier: {}
+          Tom: {}
+        households:
+          housea:
+            parents: [Alicia, Javier]
+          houseb:
+            parents: [Tom]
+        axes:
```
fpagnoux: I'm curious what brought us to mix axes and the `build_from_dict` method. I was more imagining that we would make a separate method, such as `build_from_axes`, as I think instantiating a simulation from axes and building one from JSON/YAML are pretty distinct use cases. Is it because there is a lot of common logics between both cases?

openfisca_core/entities.py
```diff
@@ -270,23 +270,17 @@ class GroupEntity(Entity):
...
         self._members_position = None
-        self.members_legacy_role = None
-        self.members = self.simulation.persons
+        self.members = persons
         self._roles_count = None
         self._ordered_members_map = None
 
     @property
     def members_role(self):
-        if self._members_role is None and self.members_legacy_role is not None:
```
fpagnoux: > 👍, it's probably time to remove this, as it's been marked as legacy for a while. 

> For information, they are still used in the survey manager (for instance [here](https://github.com/openfisca/openfisca-survey-manager/blob/master/openfisca_survey_manager/scenarios.py#L711) to set the roles of simulation individuals.

> One easy way to adapt the survey manager would be to move the logic removed here there, but at some point there should probably be a discussion about how roles should be/are represented in dataframes

benjello: Yes and I may need help for that.

openfisca_core/taxbenefitsystems.py
```diff
@@ -80,20 +75,51 @@ def base_tax_benefit_system(self):
...
-            )(value, state = state or conv.default_state)
-        if error is not None:
-            return attributes, error
-        return cls(**attributes), None
-
     def new_scenario(self):
-        scenario = self.Scenario()
-        scenario.tax_benefit_system = self
-        return scenario
+        class FakeScenario(object):
```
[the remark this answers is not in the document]
Morendil: There are about 30 calls in Core to `new_scenario`, mostly in tests, which is why we need this adapter (renamed in line with that role).

I agree with cleaning up and I think we can defer that to a later PR.

openfisca_core/simulation_builder.py
```diff
@@ -49,11 +59,19 @@ def build_from_entities(self, tax_benefit_system, input_dict, default_period = N
                 'households': {'household': {'parents': ['Javier']}}
                 })
         """
+        input_dict = deepcopy(input_dict)
+
         simulation = kwargs.pop('simulation', None)  # Only for backward compatibility with previous Simulation constructor
```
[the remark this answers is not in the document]
Morendil: We have about 20 places in Core still using the `simulation_json` parameter, best deferred to a further PR.

tests/core/test_axes.py
Morendil: You can't really build a simulation only from axes, rather axes are a way to replicate an already constructed situation (which is convenient to do using the dictionary format) a certain number of times, varying one or more input variables along the way. You have to have a situation in place already.

openfisca_core/simulation_builder.py
```diff
@@ -228,86 +281,198 @@ def check_persons_to_allocate(self, persons_plural, entity_plural,
...
+            dpath.search(json, "*/{}/{}".format(e.variable_name, str(e.period)), yielded = True),
+            None)
+        if culprit_path:
+            path = [entity.plural] + culprit_path[0].split('/')
+        else:
+            path = [entity.plural]  # Fallback: if we can't find the culprit, just set the error at the entities level
+
+        raise SituationParsingError(path, e.message)
+
+    def get_count(self, entity_name):
```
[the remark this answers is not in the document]
Morendil: It is indeed generic. It returns the count of the particular entity, for instance if you have defined 3 households `get_count("households")` returns 3. However, replication of the base situation along an axis also needs to be accounted for; if you have 3 households and replicate that 100 times along an axis involving households, then `get_count("households")` returns 300.

Added a comment to reflect this.

openfisca_core/taxbenefitsystems.py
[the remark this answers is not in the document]
sandcha: Tried to move `ScenarioAdapter.init_from_dict` to `openfisca_france` but this leads to more than 6 failing tests. So, a new issue is created in #832 to fix this in a new PR.

tests/core/test_axes.py
fpagnoux: Ok, thanks for the explanation!

tests/core/test_axes.py
```diff
@@ -1,182 +1,149 @@
...
+                - count: 2
+                  name: rent
+                  min: 0
+                  max: 3000
+                  period: 2018-11
+    """
+    data = yaml.load(input_yaml)
+    simulation = simulation_builder.build_from_dict(tax_benefit_system, data)
+    assert simulation.get_array('salary', '2018-11') == approx([0, 0, 0, 0, 0, 0])
+    assert simulation.get_array('rent', '2018-11') == approx([0, 0, 3000, 0])
```
guillett: Hey @Morendil , I don't understand why we expect a 0 as the last element of this array.

tests/core/test_axes.py
```diff
@@ -1,182 +1,149 @@
...
+    simulation_builder.add_group_entity('persons', ['Alicia', 'Javier', 'Tom'], group_entity, {
+        'housea': {'parents': ['Alicia', 'Javier']},
+        'houseb': {'parents': ['Tom']},
+        })
+    simulation_builder.register_variable('rent', group_entity)
+    simulation_builder.add_parallel_axis({'count': 2, 'name': 'rent', 'min': 0, 'max': 3000, 'period': '2018-11'})
+    simulation_builder.expand_axes()
+    assert simulation_builder.get_count('households') == 4
+    assert simulation_builder.get_ids('households') == ['housea0', 'houseb1', 'housea2', 'houseb3']
+    assert simulation_builder.get_input('rent', '2018-11') == approx([0, 0, 3000, 0])
```
guillett: Hey @Morendil , I don't understand why we expect a 0 as the last element of this array.

PR #794, opened 2018-12-10
Explicit test runner dependencies in Python 2.7

fpagnoux:
- Explicit test runner dependencies in Python 2.7
  - Using an older version of `ruamel` caused a `'CommentedSeq' object has no attribute 'get'` error.


guillett:
Merci !!

PR #797, opened 2018-12-12
Add test for tracer log

guillett:
### 25.2.2

- Add tracer format log test to ensure no regression occur

As an intense user of the tracer log I would like to make sure that this feature is kept as is or at least to be notified when changed.

Sublime text allows OK navigation:

sans titre

### DIFF
--- a/openfisca_core/tracers.py
+++ b/openfisca_core/tracers.py
@@ -198,7 +198,7 @@ class Tracer(object):
     def _print_node(self, key, depth, aggregate):
 
         def print_line(depth, node, value):
-            print("{}{} >> {}".format('  ' * depth, node, value))
+            return "{}{} >> {}".format('  ' * depth, node, value)
 
         if not self.trace.get(key):
             return print_line(depth, key, "Calculation aborted due to a circular dependency")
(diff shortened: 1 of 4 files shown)


tests/core/test_tracer.py
```diff
@@ -29,6 +29,25 @@ def test_variable_stats():
...
+    tracer.record_calculation_start("B", 2017)
+    tracer.record_calculation_end("B", 2017, 1)
+    tracer.record_calculation_end("A", 2017, 2)
+
+    from io import StringIO
+    import contextlib
+
+    log = StringIO()
+    with contextlib.redirect_stdout(log):
+        tracer.print_computation_log()
```
Morendil: Just to avoid this awkwardness, it might be worth reworking the implementation to a more functional style; have a `generate_computation_log` return an array of appropriately indented lines, and unit test that more easily; and rewrite `print_computation_log` as the trivially correct

```
def print_computation_log():
  print(generate_computation_log.join("\n")
```

Not a very hard change, and it's a small but appreciable bit of improvement in code quality.

Relatedly I notice that the `print_trace` function in `tracers.py` seems unused anywhere in the entire https://github.com/openfisca/ org. Maybe remove it as dead code, if that doesn't break any test ?

Morendil:
Rebased.

Morendil:
Split `print_computation_log` into functional core and imperative shell as suggested above.

openfisca_core/tracers.py
```diff
@@ -224,14 +224,14 @@ def print_trace(self, variable_name, period, extra_params = None, max_depth = 1,
...
                 for dependency in self.trace[key]['dependencies']:
-                    _print_details(dependency, depth + 1)
+                    return _print_details(dependency, depth + 1)
 
-        _print_details(key, 0)
+        return _print_details(key, 0)
 
-    def print_computation_log(self, aggregate = False):
+    def computation_log(self, aggregate = False):
         """
```
fpagnoux: @Morendil  this doc string should be moved to the `print_computation_log` method (this causes a warning in the doc build)

Issue #749, opened 2018-10-23
Cycle checking results in inconsistent calculations


Morendil:
Minimal repro:

```py
# -*- coding: utf-8 -*-
import openfisca_france
from openfisca_core.simulations import Simulation

tax_benefit_system = openfisca_france.CountryTaxBenefitSystem()
situation = {
    "individus": {
        "demandeur": {
            "date_naissance": {
                "2018-10": "1952-01-05",
                "2018-09": "1952-01-05",
                "2018-08": "1952-01-05",
                "2018-07": "1952-01-05"
            },
            "salaire_imposable": {
                "2018": 1000,
                "2017": 1000,
                "2016": 1000
            }
        }
    },
    "familles": {
        "_": {
            "parents": [
                "demandeur"
            ],
            "enfants": [],
        }
    },
    "foyers_fiscaux": {
        "_": {
            "declarants": [
                "demandeur"
            ],
            "personnes_a_charge": []
        }
    },
    "menages": {
        "_": {
            "personne_de_reference": [
                "demandeur"
            ],
            "enfants": []
        }
    }
}

allMonths = ['2018-09']

calculs_ok = [
{'variable': 'rfr', 'periods': ['2016']},
{'variable': 'ass', 'periods': allMonths}
]

calculs_ko = [
{'variable': 'ass', 'periods': allMonths},
{'variable': 'rfr', 'periods': ['2016']}
]

simulation_actuelle = Simulation(
tax_benefit_system=tax_benefit_system,
simulation_json=situation)

for computation in calculs_ok:
    for period in computation['periods']:
        data = simulation_actuelle.calculate(computation['variable'], period)
        print (computation['variable'])
        print (data)

simulation_actuelle = Simulation(
tax_benefit_system=tax_benefit_system,
simulation_json=situation)

for computation in calculs_ko:
    for period in computation['periods']:
        data = simulation_actuelle.calculate(computation['variable'], period)
        print (computation['variable'])
        print (data)
```

The results of computing `ass` and `rfr` should not depend on the order of evaluation, but they do:

```
rfr
[573.99994]
ass
[0.]
ass
[0.]
rfr
[0.]
```

This is impacting France and indirectly Mes-Aides (preventing an upgrade to the latest version of France in production). The France model is very ramified and cycles abound; on inspection I think the current solution (`max_nb_cycles`) is misguided.

Connected to openfisca/openfisca-france#1211

Morendil:
[...]

PR #805, opened 2018-12-28
Failing test for N/A management

guillett:
This PR contains a failing test that exhibit an unexpected behavior.

Heterogeneous inputs are acceptable in JSON input but they are not managed as I would expect.

In the following test, Javier basic income in known to be null, so it is specified but Alicia's isn't. Only her salary is known and set to 4000.
With this salary she is eligible to a basic income of 600.

Those 600 are set to 0 because Javier's basic_income is known.
When creating the input dataset, this 0 should be replaced by a NA value and that variable should not be considered as fully known.

### DIFF
--- a/tests/core/yaml_tests/test_success_2.yaml
+++ b/tests/core/yaml_tests/test_success_2.yaml
@@ -1,0 +1,14 @@
+- name: "Basic test - N/A management"
+  description:  Alicia basic_income should not be null
+  period: 2017-01
+  input:
+    persons:
+      Alicia:
+        salary: 4000
+      Javier:
+        basic_income: 0
+    household:
+      parents: [Alicia, Javier]
+  output:
+    household:
+      total_benefits: 600


Morendil:
@guillett Are you sure this test is failing for the reasons you think it is? I can see two reasons why it fails, a trivial one and a deeper one. I'll start with the latter because that's the more important point.

It's hard to make *any* version of this test pass, IMO, because of an architectural issue: as far as I'm aware, OpenFisca does not have a concept of "NA" or missing values. Its underlying computation engine, Numpy, does not have them as a first-class citizen, they were [proposed](https://docs.scipy.org/doc/numpy-1.10.4/neps/missing-data.html) but [not accepted](https://docs.scipy.org/doc/numpy-1.10.4/neps/).

As discussed [elsewhere](https://github.com/openfisca/openfisca-core/issues/749#issuecomment-432355031), OpenFisca's computation strategy is "eager evaluation". Every branch (select, where, etc.) of every formula is evaluated for every cell, for every variable you request. Every variable has either a default value or a default default value. There is no notion of "missing data" or NA.

Now for the more trivial issues. First: the country template says "basic income is provided to any adult, without considering their income" so as I read it the inputs (whether 4000, 0 or NA) shouldn't make a difference here, you're always going to get 1200.

Also, in Core's tests, `test_yaml` isn't a general facility to add YAML tests, as you would in France or other country packages; it's a test designed to check that certain YAML tests fail and others succeed. Think of it as an integration test for Core's test runner.

For a test like this we would need more of a unit test, perhaps in `test_simulations.py`. Set up a simulation with two individuals, one output variable which depends on an input variable, set a heterogenous input vector with missing values, and check that the output variable is also undefined (or NA or whatever) wherever the input has missing values. (This is one thing [syso](https://github.com/betagouv/syso) puts a lot of effort in. But it does not have the notion of entities, and I have an intuition that entities make the propagation of missing values more complex still.)

Morendil:
Closing per our discussion IRL. There are various issues that arise with non-homogenous inputs, which can in some cases be worked around. We're likely to run into them again, so there's value in thinking about where we want to go with Core's computation logic, e.g. partitioning input vectors to short-circuit computations or to distinguish missing values from zeros. That work is very much "discovery" still rather than easily and quickly implemented, so it doesn't really make sense to keep around a PR consisting of a failing test, it would be around a long time.

Issue #733, opened 2018-10-02
Run testdocs


maukoquiroga:
Hi there!

I really enjoy OpenFisca, but I recently encountered an issue.

### Here is what I did:

I modified `setup.cfg` to run doctests:

```
[nosetests]
where        = .
exe          = true
with-doctest = true
```

### Here is what I expected to happen:

```
OK
```

### Here is what actually happened:

```
FAILED (errors=10, failures=38)
make: *** [test] Error 1
```

## Context

I identify more as a:

- Developer

Morendil:
Per discussions on #806, this is not actionable as most of our doctests assume some context (fixtures, and so on) which is in fact not present.

benjello:
Does that means that we forgot about having working doctests ?

Morendil:
@benjello Yes, I think so :(

PR #806, opened 2018-12-28
Fix method delete_arrays

claireleroy:
The `Simulation` method `"delete_arrays"` does not work if the argument `period` is not provided.

But the `holders` method `"delete_arrays"` does work if the argument `period` is not provided : it then deletes all values from all periods. This is a nice feature that should also be in the `Simulation` method.

#### New features

- Add a default value equal to None for argument `period` in `delete_arrays`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ dev_requirements = [
 
 setup(
     name = 'OpenFisca-Core',
-    version = '25.2.3',
+    version = '25.2.4',
     author = 'OpenFisca Team',
     author_email = 'contact@openfisca.org',
     classifiers = [


benjello:
@Morendil : I completed the doctest but I end up locally with the following error
```bash
/openfisca/openfisca-core(fix-delete-arrays *$=)$ pytest  --doctest-modules openfisca_core/simulations.py 
=============================================================================================== test session starts ===============================================================================================
platform linux2 -- Python 2.7.15+, pytest-3.6.4, py-1.7.0, pluggy-0.8.0
rootdir: /home/benjello/openfisca/openfisca-core, inifile: setup.cfg
collected 3 items                                                                                                                                                                                                 

openfisca_core/simulations.py F

==================================================================================================== FAILURES =====================================================================================================
__________________________________________________________________________ [doctest] openfisca_core.simulations.Simulation.delete_arrays __________________________________________________________________________
413             Delete a variable's value for a given period
414 
415             :param variable: the variable to be set
416             :param period: the period for which the value should be deleted
417 
418             Example:
419 
420             >>> from openfisca_country_template import CountryTaxBenefitSystem
421             >>> simulation = Simulation(CountryTaxBenefitSystem())
422             >>> simulation.set_input('age', '2018-04', [12, 14])
UNEXPECTED EXCEPTION: ValueError(u'Unable to set value "[12 14]" for variable "age", as its length is 2 while there are 0 persons in the simulation.',)

  File "/usr/lib/python2.7/doctest.py", line 1315, in __run
    compileflags, 1) in test.globs

  File "<doctest openfisca_core.simulations.Simulation.delete_arrays[2]>", line 1, in <module>

  File "/home/benjello/openfisca/openfisca-core/openfisca_core/simulations.py", line 475, in set_input
    self.get_holder(variable).set_input(period, value)

  File "/home/benjello/openfisca/openfisca-core/openfisca_core/holders.py", line 189, in set_input
    return self._set(period, array)

  File "/home/benjello/openfisca/openfisca-core/openfisca_core/holders.py", line 200, in _set
    .format(value, self.variable.name, len(value), self.entity.count, self.entity.plural))

ValueError: Unable to set value "[12 14]" for variable "age", as its length is 2 while there are 0 persons in the simulation.

/home/benjello/openfisca/openfisca-core/openfisca_core/simulations.py:422: UnexpectedException
============================================================================================ 1 failed in 0.49 seconds =============================================================================================
```

I am surprised to get this error since I ran those doctest locally for a previous PR IIRC.

benjello:
@Morendil : should I merge or should we fix the doctest issue before ?

Morendil:
Can you try running the doctests in Python 3 ? If they pass in Python 3 I'd just merge.

Why are we not running doctests in CI ?

benjello:
They do not pass in python3 too.
I suspect a recent change in initialisation simulation which initialise the entity count to zero when running  
the first two lines of the doctest above reproduced here:
```python
simulation = Simulation(CountryTaxBenefitSystem()) 
```
which forbids initialisation
```python
ValueError: Unable to set value "[12 14]" for variable "age", as its length is 2 while there are 0 persons in the simulation.
```

As for the doctest see https://github.com/openfisca/openfisca-core/pull/784 and https://github.com/openfisca/openfisca-core/issues/733

I am ok to start tackling #733 but the set_input problem should be investigated and solved ASAP as a an example of a working doctest.

@Morendil:
 - what about merging this PR, and fixing all the doctest on another one ?
 - what about you helping me solve the set_input problem and I can propose comprehensive doctest fixes ;-)

benjello:
@openfisca/core-contrib @Morendil ; should I merge ?

Morendil:
We had a look at this with @sandcha this Monday, I'm only now getting around to writing down some notes. Some of our conclusions were:
- we don't think we are going to get the doctests to pass without a significant effort; there are many errors to fix, and many doctests which use constructs like `person('variable')` which will not pass because we do not have doctest fixtures like "person"
- we went back to a commit before the introduction of SimulationBuilder; at that point you could indeed call `set_input` with a vector of any number of entries, but then the `persons.count` in that Simulation instance would still be 0, which isn't a consistent state; though we didn't check what happened if you tried to calculate something in that situation, how could you get a correct result without a count, person ids and so on ?

So, for the moment, I think it's best to treat the doctests as just documentation and go ahead with merging.

PR #811, opened 2019-01-17
Introduce LookupTaxScale

Morendil:
#### Technical changes

- Allow France to model "Chèque Energie" in a cleaner way:
  - Introduce SingleAmountTaxScale (SATS), a simpler form of MarginalAmountTaxScale (MATS)
    - whereas MATS sums the values in brackets, up to the amount subject to the scale, SATS only "looks up" the appropriate value for the amount subject to the scale, and is thus the simpler mechanism
    - use `numpy.digitize`, allowing callers to specify right or left intervals
    - introduce a `type` tag in `brackets` object, thus far with only `single_amount` allowed
  - Rename AmountTaxScale to MarginalAmountTaxScale and make it inherit from SATS

This is non-breaking, as there are no direct clients of these classes outside of Core.

### DIFF
--- a/openfisca_core/taxscales.py
+++ b/openfisca_core/taxscales.py
@@ -146,9 +146,19 @@ class AmountTaxScale(AbstractTaxScale):
             self.thresholds.insert(i, threshold)
             self.amounts.insert(i, amount)
 
-    def calc(self, base, right=False):
-        bracket_indices = np.digitize(base, self.thresholds, right=right)
-        return np.array(self.amounts)[bracket_indices - 1]
+    def calc(self, base):
+        base1 = np.tile(base, (len(self.thresholds), 1)).T
+        thresholds1 = np.tile(np.hstack((self.thresholds, np.inf)), (len(base), 1))
+        a = max_(min_(base1, thresholds1[:, 1:]) - thresholds1[:, :-1], 0)
+        return np.dot(self.amounts, a.T > 0)
+
+
+class LookupTaxScale(AmountTaxScale):
+    def calc(self, base, right=True):
+        guarded_thresholds = np.array([-np.inf] + self.thresholds + [np.inf])
+        bracket_indices = np.digitize(base, guarded_thresholds, right=right)
+        guarded_amounts = np.array([0] + self.amounts + [0])
+        return guarded_amounts[bracket_indices - 1]
 
 
 class LinearAverageRateTaxScale(AbstractRateTaxScale):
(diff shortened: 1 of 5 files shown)


Morendil:
Oops! We've just realized that we have divergent ideas of what AmountTaxScale is supposed to compute. My interpretation was wrong, the previous implementation *summed* the amount up to the interval containing base, rather than a simple lookup as I thought. It's nice that the unit test written previously by @sandcha and @pblayo caught that!

I think we really need two concepts:
- AmountTaxScale sums its amounts up to the interval containing base
- LookupTaxScale looks up the amount in the interval containing base

However this runs into a second problem, how to tell one kind of scale apart from the other.

openfisca_core/taxscales.py
```diff
@@ -153,6 +153,14 @@ def calc(self, base):
         return np.dot(self.amounts, a.T > 0)
 
 
+class LookupTaxScale(AmountTaxScale):
```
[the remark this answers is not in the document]
sandcha: > Reverse amount scales classes hierarchy and contents (except `calc` function) in order to define : 
> * what we name today as `LookupTaxScale` as basic case for amount tax scale
> * what we name today as `AmountTaxScale` as kind of marginal amount tax scale
> ?

sandcha: Add docstring to explicit this scale behaviour?

openfisca_core/taxscales.py
```diff
@@ -153,6 +153,14 @@ def calc(self, base):
...
+    def calc(self, base, right=False):
+        guarded_thresholds = np.array([-np.inf] + self.thresholds + [np.inf])
+        bracket_indices = np.digitize(base, guarded_thresholds, right=right)
+        guarded_amounts = np.array([0] + self.amounts + [0])
```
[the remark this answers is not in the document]
Morendil: Agreed, that's an issue to watch out for if the use of these scales is generalized. The notion of "unit" for parameters (and of values that make sense for a given "type" of parameter) isn't as well-defined as it is for variables. In any case, it looks like we will be forced to revisit parameters/scales pretty soon, due to the need to merge IPP parameters.

openfisca_core/taxscales.py
Morendil: Done. However I have no idea how to describe the *RateTaxScale classes and leave that to future improvement :)

benjello: A rate tax scale compute the tax as a fraction of the taxable amount (or taxable base) according to a marginal or average tax scale.

PR #813, opened 2019-01-20
Remove base functions

### LINKED ISSUE #460: Make explicit input variables temporal inferences
## This issue handles a part of #458.

For `base_function`, the issue rose with the formula `rsa`, where macrosimulation (where missing values should be the default ones (or raise warnings, @benjello?) and production (where missing values should be inferred) expected behaviours differ.

Default values in input variables are the source of the issue: if a formula needs values that are undefined (e.g. because they are in previous months from the calculation), it won't crash nor log anything because the input variables return their default values. This inference behaviour seems to be fine for macrosimulation, but production would expect a time-based inference, where the value is copied from its closest defined value rather than a global default one. Changing the inference mode means changing the `base_function`, but that raises conflicts between uses.

There is a consensus between @cbenz @michelbl @fpagnoux that `base_function` is incomprehensible and counterintuitive. It is at least misnamed and it should be more explicit that it is currently used purely as an extrapolation mechanism.

There is also a consensus that defaulting to the simulation period when requesting to compute a formula is not a proper behaviour: this piece of information should be mandatory to avoid incomprehensible behaviour on omission.

### Candidates

- Remove the default period of the input variables: it is currently the period of the simulation, while it could be totally explicit. _**This solution seems to be the preferred one.** For example, it is “well-known” that the revenue tax is defined on the year 2 years ago, not the current one. Documenting the requested definition periods in each formula and forcing the inputs to explicitly state their period could solve the problem._
- Change the default period of the input variables: it is currently the period of the simulation, while it could be “at all times”. _This needs defining “at all times” in Core._

### Eliminated

- Configuration switch to alternate between the two behaviours in the input variables. _This solution does not scale and increases complexity._
- Implement two different formulas, each one implementing one specific behaviour. _Considering that callers would need to define which version to call, that would cascade upwards with each caller formula also having an alternate version, which would end up being the same as having a configuration switch._
- Improve debugging tools to report used inputs. _The produced data would be irrelevant because of the size of the dependency graph, which would make it too hard to identify the pertinent parts._
- Stop providing default values for inputs, which would make computations report errors if they don't define necessary values. _This solution is probably not viable since the dependency graph is quickly huge, and that would mean defining very large amounts of disconnected values in order to achieve the simplest computations. Since there is no point in providing different “default values” sets, there is no point either in a solution where default values would be an opt-in provided by a file, for example._

Suggested action plan
----------------------

- [ ] Remove `base_function`, check how many tests fail and evaluate the amount of work to refactor the inputs.
- [ ] Make it mandatory to specify input variable periods in test cases.
- [ ] Recommend to document input periods in the formula documentation whenever those periods do not match the formula computation period.

Morendil:
Fixes #460
Depends on #841

#### Breaking changes

- Require clients to make explicit when input variables cover a range of dates, rather than allowing inputs to be derived from past or future values; also removes the `missing_value` base function

#### Migration instructions

You might need to change your code if any of the following applies:
- you are a model author, and there are variables in your model that use a `base_function` attribute
- your model **or application** uses any non-numeric variables (boolean, date, string or Enum) used in, for which the period on which you define **inputs** does not match the period for which you are requesting **outputs**

An example is the test below, where `unemployed` is assumed to be a boolean variable, and `unemployment_benefits` an aid equal to 80% of the last known salary **only if** the beneficiary has the status "unemployed" contemporary with evaluating the formula.

This test would be **passing** prior to this version, because the default "base functions" behavior was to "prolong" the definition of any boolean, enumerated, date or string variables indefinitely in the future after the last known period.

It will be **failing** after this version, because this inference was removed. OpenFisca only evaluates Alicia's status to "unemployed" if we are specifically telling it to do so.

The test could be made to pass again as follows:

#### Before

```yaml
name: Base function removal example
  period: 2017-01
  input:
    persons:
      Alicia:
        unemployed:
          2010-01: true
        salary:
          2009-12: 1000
  output:
    persons:
      Alicia:
        unemployment_benefits:
          2017-01: 800
```

#### After

```yaml
name: Base function removal example
  period: 2017-01
  input:
    persons:
      Alicia:
        unemployed:
          "month:2010-01:85": true
        salary:
          2009-12: 1000
  output:
    persons:
      Alicia:
        unemployment_benefits:
          2017-01: 800
```

For this to work, the variable `unemployed` must have its attribute `set_input` set either to `set_input_dispatch_by_period` or to `set_input_divide_by_period`.

Note that migration might be needed not only for YAML tests, but for any code using OpenFisca relying (on purpose or by accident) on the temporal inferences that are now removed.

For a real-world example covering many cases see [the corresponding PR migrating the France model](https://github.com/openfisca/openfisca-france/pull/1261).


sandcha:
This will impact this [section](https://github.com/openfisca/openfisca-doc/blob/602c7204dfa79a4eb6371436886e3d469cbac57b/source/coding-the-legislation/inferences.md#with-base_function) of the official documentation.

fpagnoux:
> * Requires clients to make explicit when input variables cover a range of dates, rather than allowing inputs to be derived from past or future values; also removes the `missing_value` base function

I think we should make sure that this would be a positive change for users.

 I entirely agree that `base_functions` are shaky, and neither well named nor properly documented. But I think we should get user input on the underlying inference behavior before remove it all together.

The question I'd like to ask users is, if they declare the following input:

```yaml
Tom:
  handicap:
    2019-02: True
```

What is the behavior they find the least surprising ?
- A: Tom is handicapped only in February, but neither in March, nor in January
- B: Tom is handicapped from February (in March, April, etc.), but not in January
- C: Tom has always been handicaped

This PR suggests to switch to behavior A. The current behavior is B for boolean and date variables, with the possibility to force A or C by specifying an explicit base_function.

fpagnoux:
If we were to switch to A, I think we should also make sure that there is an easy way for users to specify that, for instance, Tom has always been handicapped, or has been handicapped since February, without end date. 

>  And on a more technical note, we should make sure that relying heavily on `set_input_dispatch` rather than `base_function` doesn't degrade memory performances by creating multiple identical vectors in the cache.

Br3nda:
Im worried about this change. We use this feature to mark when someone's state changes, for our citizenship act (the days someone arrived or departed the country) instead of requiring an api call have a value for every single day over 15 years. Sending all values would be a massive json string and require rebuilding parts of our citizenship eligibility app.

Morendil:
@Br3nda Syntax-wise, you can [set a value for a period range](https://openfisca.org/doc/coding-the-legislation/35_periods.html) like so:

```YAML
persons:
    - id: "Tahi"
      present_in_new_zealand:
        "day:2015-01-01:365": true
```

This would set the variable to true for the entire year. So in general, you won't ever need more entries in your input than there are state changes. Likewise, the only period that would require paying special attention to is the last one you are expecting to hold "up until present day"; you can extend this by however many days have elapsed since this last change, relative to the period for which you are computing the formula.

Morendil:
> The current behavior is B for boolean and date variables

Also enums and strings.

Br3nda:
@Morendil that sounds like an easily implemented work around -- We'll be working on that system again in 2 weeks and can give your PR a test to check we're still good.

fpagnoux:
If users (ping @benjello @guillett) understand the behavior change and don't see any major risk with it, let's keep moving 👌.

If we are to rely more on `set_input_dispatch_by_period`, and use periods like `day:2015-01-01:365`, we need a small optimization to avoid duplicating vectors in memory: #827

Morendil:
@Br3nda Did you have time to take a look at this in terms of impact on citizenship act evaluation?

sandcha:
@Morendil With alternative solution, shouldn't those two tests pass?
```yaml
- name: "Basic income, default syntax for age input value"
  input:
    age:
      2015-01: 17
      2016-01: 17
      2017-01: 19
  output:
    basic_income:
      2015-01: 0
      2016-01: 0
      2017-01: 600

- name: "Basic income, shortened syntax for age input value"
  input:
    age:
      month:2015-01:2: 17
      2017-01: 19
  output:
    basic_income:
      2015-01: 0
      2016-01: 0
      2017-01: 600
```
The second one fails with `openfisca test`.
> And setting a wrong period size (like `year:2015-01:2`) fails with this non helping message `(...) File "/.../openfisca-core/openfisca_core/tools/test_runner.py", line 235, in _parse_test
    raise ValueError("Unexpected error while parsing {}".format(test['file_path']))
ValueError: Unexpected error while parsing ***.yaml`

Morendil:
@sandcha The key needs to be enclosed in quotes whenever it contains characters that could be ambiguous, in this case ":". The following should work:

`"month:2015-01:24": 17`

Morendil:
@Br3nda @fpagnoux At least one assumption I was making turns out to be wrong:

```Python
from openfisca_france.france_taxbenefitsystem import FranceTaxBenefitSystem
tbs = FranceTaxBenefitSystem()
sb = SimulationBuilder()
s = sb.build_default_simulation(tbs, count = 1)
s.set_input('contribution_developpement_apprentissage','month:2019-01:7',100000)
s.calculate('contribution_developpement_apprentissage','2019-01')
```

This variable `contribution_developpement_apprentissage` has neither `set_input_dispatch_by_period` nor `set_input_divide_by_period`. The above input yields:

```Python
array([0.], dtype=float32)
```

I find that quite counter-intuitive, to the point that if I were new to OpenFisca I would probably consider it a bug. In terms of the current implementation, my implicit assumption was that `set_input_dispatch_by_period` would be the default.

In particular this probably means the example [described in the documentation](https://openfisca.org/doc/coding-the-legislation/35_periods.html) ("Income tax over time") doesn't actually work. (I'm going to check now.)

@fpagnoux Is there any reason not to make "dispatch" the default behaviour here, as the documentation seems to suggest it is?

fpagnoux:
That is indeed a bug ☹️ .

The current behavior is supposed to be: raise an error if the input period is different than the definition period of the variable, except if a `set_input` has been defined. A 7-month period input for a month-defined variable is supposed to raise an error. 

Here, instead, no error is raised and the input is lost in the limbo of the memory storage. That's not good. 

Making "dispatch" the default behavior when a variable is set for a period bigger than its definition period seems like a safe assumption to me, and would be useful now that this PR removes the previous inference mechanism.

fpagnoux:
A fixed to the #841 blocking bug is suggested in #859

mengwong:
> Im worried about this change.

to bring up `openfisca-aotearoa`, I ended checking out an old commit `fa134946b @ 30.0.3`

```
20190502-18:33:56 mengwong@venice4:~/src/ServiceInnovationLab/openfisca-aotearoa% grep missing_value ALL
Binary file openfisca_aotearoa/variables/acts/income_tax/family_scheme/__pycache__/family_scheme.cpython-37.pyc matches
openfisca_aotearoa/variables/acts/income_tax/family_scheme/family_scheme.py:    base_function = missing_value
openfisca_aotearoa/variables/acts/income_tax/family_scheme/family_scheme.py:    base_function = missing_value
Binary file openfisca_aotearoa/variables/demographics/__pycache__/ages.cpython-37.pyc matches
openfisca_aotearoa/variables/demographics/ages.py:    base_function = missing_value
```

fpagnoux:
Hi @mengwong,

Please let us know if there is any way we can help you migrate to the latest versions of Core. 

The `missing_value` behavior was to raise an error when trying to calculate a variable that hasn't been set as an input (instead of using a default value, which is the usual behavior). Is that a behavior that you need in your model?

PR #814, opened 2019-01-21
Ensure correct period is given in failing test help messages

guillett:
Given the test:

```Markdown
- name: "A test"
  period: 2015-01
  input:
  output:
    income_tax:
      2015-02: 10
```

The failure is currently reported as:

> AssertionError: In test 'A test', in file '/app/tests/core/yaml_tests/test_failure_help_message.yaml', income_tax@**2015-01**: [0.] differs from 10.0 with an absolute margin [10.] > 0

The problem is in the bolded part: the period reported is the simulation period, it should properly be the period for which the output variable was requested, since that's what was compared to the expected value and found to differ. The current behaviour is confusing.


Morendil:
Can you describe in words what this fixes or improves? The code strikes me as roundabout and complex, and I don't know enough about it to know if that's appropriate.

guillett:
Thanks @Morendil for your message. I updated the PR title to be more explicit.

guillett:
Thanks for your review. I did not find any existing test of `_check_variable` nor `_run_test` so I decided to use the current testing strategy.

Most people with test failures while developping with OpenFisca are suffering from that bug. I am OK with that fix without tests merged in master.

PR #817, opened 2019-01-28
Ensure consistent results in the presence of "spirals"

sandcha:
* Breaking change
* Improve cycle and spiral detection, giving consistent results more systematically
* Migration instructions:
  - remove all optional parameters `max_nb_cycles`
  - avoid relying on cached values of a computation (especially wrong values)

--------
## Detailed explanation of the changes

Terminology:
- a "cycle" describes the situation where computing an OpenFisca variable for a given period eventually triggers the computation of the same variable for the same period; this cannot be resolved, so an exception is raised and the computation is aborted;
- a "spiral" is a related situation where computing a variable for a given period requires computing the same variable for a different period.

This PR makes that distinction more explicit by the type of exception raised.

Previously it was the responsibility of callers (country package authors and other reusers) to control the behaviour in the presence of spirals, by setting a limit on the number of "loops" a computation can go through. In such cases, the computation is not aborted with an error but allowed to proceed after returning a default value (usually 0). **This resulted in inconsistent computations in interaction with caching of computed values.**

This PR removes this responsibility from the caller, and Core now "goes the extra mile" to detect this condition, and to behave more reliably - returning consistent results that do not depend on the order in which computations are performed.

Spirals are more complex to analyse in the absence of static determination of the dependency between variables, so a heuristic is applied in this PR: limit the number of loops to 1. In addition, values computed in a "spiral" configuration **should not be cached**. We have verified empirically that with these changes, the result of computations no longer depends on their ordering.

This PR effects the following changes:
- maintain a separate stack structure that mimics the actual Python stack
- upon exceeding the loop limit, propagate the error up that parallel stack
- but propagate it only up to the variable for which a spiral exists, not above
- raise an exception to exit the formula in which the error was detected
- return the default value from this computation
- on completion of the original computation, invalidate the cached values of all variables marked as involved in the spiral

Connected to openfisca/openfisca-france#1211

### DIFF
--- a/openfisca_core/simulations.py
+++ b/openfisca_core/simulations.py
@@ -320,12 +320,11 @@ class Simulation(object):
         a heuristic, against "quasicircles", where the evaluation of a variable at a period involves
         the same variable at a different period.
         """
-        previous = [(_name, _period) for (_name, _period) in self.computation_stack if _name == variable.name]
+        previous_periods = [_period for (_name, _period) in self.computation_stack if _name == variable.name]
         self.computation_stack.append((variable.name, str(period)))
-        for pair in previous:
-            if pair == (variable.name, str(period)):
-                raise CycleError("Circular definition detected on formula {}@{}".format(variable.name, period))
-        spiral = len(previous) >= self.max_spiral_loops
+        if str(period) in previous_periods:
+            raise CycleError("Circular definition detected on formula {}@{}".format(variable.name, period))
+        spiral = len(previous_periods) >= self.max_spiral_loops
         if spiral:
             self.invalidate_spiral_variables(variable)
             message = "Quasicircular definition detected on formula {}@{} involving {}".format(variable.name, period, self.computation_stack)
--- a/openfisca_core/simulations.py
+++ b/openfisca_core/simulations.py
@@ -131,23 +131,30 @@ class Simulation(object):
         array = None
 
         # First, try to run a formula
-        proceed = self._check_for_cycle(variable, period)
-        if proceed:
+        try:
+            self._check_for_cycle(variable, period)
             array = self._run_formula(variable, entity, period)
 
             # If no result, try a base function
             if array is None and variable.base_function:
                 array = variable.base_function(holder, period)
 
-        # If no result, or aborted computation, use the default value
-        if array is None:
-            array = holder.default_array()
+                # If no result, use the default value and cache it
+            if array is None:
+                array = holder.default_array()
+            if self.trace:
+                self.tracer.record_calculation_end(variable.name, period, array, **parameters)
 
-        holder.put_in_cache(array, period)
-
-        if self.trace:
-            self.tracer.record_calculation_end(variable.name, period, array, **parameters)
-        self._clean_cycle_detection_data(variable.name)
+            holder.put_in_cache(array, period)
+        except SpiralError:
+            _name, _period, _status = self.computation_stack[-1]
+            if not _status:
+                raise
+            # We're done unwinding the stack, return default
+            if array is None:
+                array = holder.default_array()
+        finally:
+            self._clean_cycle_detection_data(variable.name)
 
         self._check_end_of_calculate()
 
@@ -318,25 +325,28 @@ class Simulation(object):
         a heuristic, against "quasicircles", where the evaluation of a variable at a period involves
         the same variable at a different period.
         """
-        previous = [(_name, _period) for (_name, _period) in self.computation_stack if _name == variable.name]
-        self.computation_stack.append((variable.name, str(period)))
+        previous = [(_name, _period) for (_name, _period, _status) in self.computation_stack if _name == variable.name]
+        self.computation_stack.append([variable.name, str(period), True])
         for pair in previous:
             if pair == (variable.name, str(period)):
                 raise CycleError("Circular definition detected on formula {}@{}".format(variable.name, period))
         spiral = len(previous) >= self.max_spiral_loops
         if spiral:
             self.invalidate_spiral_variables(variable)
-            return False
-        return True
+            message = "Quasicircular definition detected on formula {}@{} involving {}".format(variable.name, period, self.computation_stack)
+            raise SpiralError(message, variable.name)
 
     def invalidate_spiral_variables(self, variable):
         count = 0
         for frame in self.computation_stack[::-1]:
-            self.invalidated_caches.add(frame)
+            (name, period, status) = frame
+            self.invalidated_caches.add((name, period))
             if frame[0] == variable.name:
                 count += 1
                 if count > self.max_spiral_loops:
                     break
+            # The status of the variable that started the loop is true, so that it returns a value
+            frame[2] = False
 
     def _clean_cycle_detection_data(self, variable_name):
         """
--- a/openfisca_core/simulations.py
+++ b/openfisca_core/simulations.py
@@ -139,7 +139,7 @@ class Simulation(object):
             if array is None and variable.base_function:
                 array = variable.base_function(holder, period)
 
-                # If no result, use the default value and cache it
+            # If no result, use the default value and cache it
             if array is None:
                 array = holder.default_array()
 
--- a/openfisca_core/simulations.py
+++ b/openfisca_core/simulations.py
@@ -332,7 +332,7 @@ class Simulation(object):
 
     def invalidate_spiral_variables(self, variable):
         count = 0
-        for frame in self.computation_stack[::-1]:
+        for frame in reversed(self.computation_stack):
             self.invalidated_caches.add(frame)
             if frame[0] == variable.name:
                 count += 1
--- a/openfisca_core/simulations.py
+++ b/openfisca_core/simulations.py
@@ -129,20 +129,32 @@ class Simulation(object):
 
         array = None
 
-        # First, try to run a formula
-        array = self._run_formula(variable, entity, period)
+        try:
+            # First, try to run a formula
+            self._check_for_cycle(variable, period)
+            array = self._run_formula(variable, entity, period)
 
-        # If no result, try a base function
-        if array is None and variable.base_function:
-            array = variable.base_function(holder, period)
+            # If no result, try a base function
+            if array is None and variable.base_function:
+                array = variable.base_function(holder, period)
+
+            if array is not None:
+                holder.put_in_cache(array, period)
+        except SpiralError as spiral:
+            spiral_marker = spiral.args[1]
+            occurrences = [pair[0] for pair in self.computation_stack if pair[0] == spiral_marker]
+            if len(occurrences) > 1:
+                raise spiral
+            if self.tracer:
+                self.tracer._computation_log.append(["spiral_" + variable.name + ":" + str(period), 1])
+        finally:
+            if self.trace:
+                self.tracer.record_calculation_end(variable.name, period, array, **parameters)
+            self._clean_cycle_detection_data(variable.name)
 
         # If no result, use the default value
         if array is None:
             array = holder.default_array()
-
-        holder.put_in_cache(array, period)
-        if self.trace:
-            self.tracer.record_calculation_end(variable.name, period, array, **parameters)
 
         return array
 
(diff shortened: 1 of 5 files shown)


Morendil:
Rebased to get CI configuration with Python 3 only.

tests/core/test_cycles.py
```diff
@@ -116,36 +116,37 @@ def formula(person, period):
...
         ).new_simulation(debug = True)
-    simulation.calculate('variable3', period = reference_period)
+    variable3 = simulation.calculate('variable3', period = reference_period)
+    assert_near(variable3, [0])
 
 
-def test_allowed_cycle():
+def test_spiral_heuristic():
     """
     Calculate variable5 then variable6 then in the order order, to verify that the first calculated variable
```
benjello: other order

benjello:
@Morendil thanks for the details. I would like to be 100 % sure. Are every spiral always stopped at the first step or would they follow max_nb_cycles ? As far as I understand, when we hit a spiral we use the default value as a heuristic but we do not set the variable at this value for that period.

Morendil:
@benjello I should probably have posted an update after the above comment (with `test_spiral_heuristic`), the real test in the PR is a bit different. But yes, the idea is that when we hit a spiral we return the default but we don't cache the result. It's the interaction of cycle detection and caching that is at the root of the problem.

Morendil:
Rebased.

benjello:
Ok @Morendil this is neat. How can we help to have this merged ASAP ?
Does it causes troubles to openfisca-france ?

Morendil:
I have bad news and good news. The bad news is that we are trying too hard to **not** cache values involved in spirals, but these variables are recomputed a lot. So this kills performance - not just a little degradation but to the point of feeling like an infinite loop. Unusable in production, anyway.

I'm reminded of the saying "There are only two hard things in Computer Science: cache invalidation and naming things." Also, much gratitude to @sandcha and @fpagnoux for the performance measurement script which identified the problem last night.

The good news is that thanks to our earlier work, and in particular representing the computation stack explicitly, we should still be able to a) solve the original problem, b) simplify the code quite a bit, c) expect some modest performance gains.

Our problem is that successive calls to calculate yield inconsistent results because of cached values. We can see this clearly on the first spiral detected in `revenu_disponible`:

`[['revenu_disponible', '2018', True], ['pensions_nettes', '2018', True], ['chomage_net', '2018-01', True], ['chomage_imposable', '2018-01', False], ['csg_deductible_chomage', '2018-01', False], ['taux_csg_remplacement', '2018-01', False], ['rfr', '2016', False], ['rni', '2016', False], ['rng', '2016', False], ['rbg', '2016', False], ['revenu_categoriel', '2016', False], ['revenu_categoriel_tspr', '2016', False], ['traitements_salaires_pensions_rentes', '2016', False], ['revenu_assimile_salaire_apres_abattements', '2016', False], ['revenu_assimile_salaire', '2016', False], ['chomage_imposable', '2016-01', False]]`

Variable `rfr` is collateral damage in the spiral that involves `chomage_imposable` and is set to 0. But this does not matter *during* a computation. It's only a problem because we reuse the cache *across* computations (successive calls to calculate).

Instead of trying **not to cache** values involved in a spiral, we can use the same approach but **clean the cache** of the values afterwards. We check that the stack is empty, to detect returning from the initial call to `calculate`. I'm hopeful this will have the same result. (Pushing the commit that uses this approach, but I haven't yet ran the tests for inconsistent values again.)

Also, there's really no reason now to keep raising and catching exceptions so the code should eventually look simpler, and in addition these operations are expensive so we should see a performance gain.

### Performance on master:
Premier test revenu disponible
2.927500 s
Second test revenu disponible
2.691405 s
3e test revenu disponible
2.688748 s
Premier test ciblé spirale
1.298005 s
Second test ciblé spirale
1.320369 s
3e test ciblé spirale
1.310184 s

### Performance on this PR:
Premier test revenu disponible
2.604535 s
Second test revenu disponible
2.420090 s
3e test revenu disponible
2.381879 s
Premier test ciblé spirale
0.774254 s
Second test ciblé spirale
0.809069 s
3e test ciblé spirale
0.809284 s

Morendil:
(Unfortunately Core tests are passing but a lot of France tests are broken with this… still investigating.)

Morendil:
The strategy "partially invalidate the cache, removing variables caught in a spiral" runs into some issues.

Here is a simplified version of `tests/formulas/rsa/rsa_2017.yaml` from France:
```
- name: Les primes sur un mois ne sont pas moyennées
  input:
    Alicia:
      salaire_net:
        2017-01: 200
  output:
    rsa:
      2017-01: 223
    rsa_fictif:
      2016-12: 0
      2016-11: 535.17 - 200
      2016-10: 535.17 - 200
```

This fails with the cache invalidation changes, because the `rsa_fictif` output check only works if `rsa_fictif` is already in the cache; and because `rsa_fictif` is caught in a spiral of `rsa`, we remove it from the cache in this PR.

If you try running this test on master, but interchange `rsa` and `rsa_fictif`, the test will fail with:
`TypeError: formula_2016_10() missing 1 required positional argument: 'mois_demande'`

This is because of `extra_params`, which prevents `rsa_fictif` from being computed directly. But of course this breaks the implicit rules of YAML tests: the results should not depend on the order of computation, and any variable should be computable directly without supplying extra parameters.

This confirms the urgency of https://github.com/openfisca/openfisca-france/pull/1284 - and that this `extra_params` business was really not a good idea.

Morendil:
The situation is a bit worse than I thought. This branch causes 6 errors and 12 failures in France tests. The 6 errors are due to the extra_params situation. One of the failures could be solved by max_spiral_loops=2. However, the rest (11) appear to be related to RSA.

Here is a test adapted from one of the cases in `tests/formulas/revenu_disponible.yaml`:

```
- name: f1aw_f4ba_2017
  description: Revenus fonciers (4BA) et rentes viagères à titre onéreux (1AW), 2017 (vérifie notamment la formule de revenus_nets_du_capital qui est un peu complexe pour ces deux types de revenu)
  period: 2017
  absolute_error_margin: 1
  input:
    f1aw: 30000
    f4ba: 20000
  output:
    prelevements_sociaux_revenus_capital: -7052 # -(0.7*30000 + 20000) * (0.045 + 0.02 + 0.003 + 0.005 + 0.082)
    revenus_nets_du_capital: (20000 - 7052)
    pensions_nettes: 30000
    irpp: -6593 # Montant calculé sur le simulateur de la DGFiP
    impots_directs: -6593 # Montant calculé sur le simulateur de la DGFiP
    revenu_disponible: (20000 - 7052) + 30000 - 6593
- name: f1aw_f4ba_2017
  description: Revenus fonciers (4BA) et rentes viagères à titre onéreux (1AW), 2017 (vérifie notamment la formule de revenus_nets_du_capital qui est un peu complexe pour ces deux types de revenu)
  period: 2017
  absolute_error_margin: 1
  input:
    f1aw: 30000
    f4ba: 20000
  output:
    prelevements_sociaux_revenus_capital: -7052 # -(0.7*30000 + 20000) * (0.045 + 0.02 + 0.003 + 0.005 + 0.082)
    revenus_nets_du_capital: (20000 - 7052)
    pensions_nettes: 30000
    irpp: -6593 # Montant calculé sur le simulateur de la DGFiP
    impots_directs: -6593 # Montant calculé sur le simulateur de la DGFiP
    revenu_disponible: (20000 - 7052) + 30000 - 6593
    rsa:
      2017-01: 0
```

Note that we are testing `rsa` at the end. This test is passing. However, consider the following:

```
- name: f1aw_f4ba_2017
  description: Revenus fonciers (4BA) et rentes viagères à titre onéreux (1AW), 2017 (vérifie notamment la formule de revenus_nets_du_capital qui est un peu complexe pour ces deux types de revenu)
  period: 2017
  absolute_error_margin: 1
  input:
    f1aw: 30000
    f4ba: 20000
  output:
    prelevements_sociaux_revenus_capital: -7052 # -(0.7*30000 + 20000) * (0.045 + 0.02 + 0.003 + 0.005 + 0.082)
    revenus_nets_du_capital: (20000 - 7052)
    pensions_nettes: 30000
    irpp: -6593 # Montant calculé sur le simulateur de la DGFiP
    impots_directs: -6593 # Montant calculé sur le simulateur de la DGFiP
    revenu_disponible: (20000 - 7052) + 30000 - 6593
- name: f1aw_f4ba_2017
  description: Revenus fonciers (4BA) et rentes viagères à titre onéreux (1AW), 2017 (vérifie notamment la formule de revenus_nets_du_capital qui est un peu complexe pour ces deux types de revenu)
  period: 2017
  absolute_error_margin: 1
  input:
    f1aw: 30000
    f4ba: 20000
  output:
    rsa:
      2017-01: 0
    prelevements_sociaux_revenus_capital: -7052 # -(0.7*30000 + 20000) * (0.045 + 0.02 + 0.003 + 0.005 + 0.082)
    revenus_nets_du_capital: (20000 - 7052)
    pensions_nettes: 30000
    irpp: -6593 # Montant calculé sur le simulateur de la DGFiP
    impots_directs: -6593 # Montant calculé sur le simulateur de la DGFiP
    revenu_disponible: (20000 - 7052) + 30000 - 6593
```

This test **does not pass**, because OpenFisca calculates an amount of 535.17 for `rsa` instead of 0. 

This is the first computation OpenFisca performs, so it must be "the right result" for these inputs. The computed RSA of 0 must be the result of a side-effect. And in fact it turns out that `rsa` is set to 0 as the result of a computation aborted due to a "spiral".

fpagnoux:
Now that extra parameters have been removed, does this unblock this PR?

Morendil:
@fpagnoux It did help quite a bit. There may be other cache-related or depth-related issues (those I'm aware of are detailed in https://github.com/openfisca/openfisca-france/pull/1268). I've also updated the performance test results in the PR description.

openfisca_core/simulations.py
```diff
@@ -148,31 +153,39 @@ def calculate(self, variable_name, period, **parameters):
...
 
-        holder.put_in_cache(array, period, extra_params)
-        if self.trace:
-            self.tracer.record_calculation_end(variable.name, period, array, **parameters)
+        self._check_end_of_calculate()
 
         return array
 
+    def _check_end_of_calculate(self):
+        if len(self.computation_stack) == 0:
```
fpagnoux: > An early return could reduce nesting:

```suggestion
        if self.computation_stack:   # Some comment about why we don't delete from the cache in that case
          return
        for (_name, _period) in self.invalidated_caches:
          holder = self.get_holder(_name)
          holder.delete_arrays(_period)
        
```

tests/core/test_cycles.py
```diff
@@ -59,7 +58,7 @@ class variable5(Variable):
     definition_period = MONTH
 
     def formula(person, period):
-        variable6 = person('variable6', period.last_month, max_nb_cycles = 0)
```
fpagnoux: Some comments above feel a little outdated. For instance does "but without explicit cycle allowed" still make sense now that it's not the formula writer responsibility to allow a cycle or not?

openfisca_core/simulations.py
```diff
@@ -71,10 +76,10 @@ def __init__(
         self.period = period
 
         # To keep track of the values (formulas and periods) being calculated to detect circular definitions.
-        # See use in formulas.py.
-        # The data structure of requested_periods_by_variable_name is: {variable_name: [period1, period2]}
-        self.requested_periods_by_variable_name = {}
-        self.max_nb_cycles = None
+        # The data structure of computation_stack is:
+        # [('variable_name', 'period1'), ('variable_name', 'period2')]
+        self.computation_stack = []
```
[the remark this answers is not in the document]
Morendil: Yes… it's not clear how we can exploit the similarities though, as tracers are an optional feature at the moment, whereas cycle detection is mandatory.

openfisca_core/variables.py
```diff
@@ -277,6 +277,8 @@ def set_base_function(self, base_function):
             raise ValueError('Unexpected base_function {}'.format(base_function).encode('utf-8'))
 
         if self.is_period_size_independent and base_function is None:
+            if self.end is not None:
+                return requested_period_default_value
```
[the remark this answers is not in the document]
Morendil: Good catch, that was an attempt at effecting the same as #826, will remove.

openfisca_core/simulations.py
```diff
@@ -50,15 +55,17 @@ def __init__(
...
+        self.invalidated_caches = set()
 
         self.debug = False
         self.trace = False
         self.opt_out_cache = False
 
+        # controls the spirals detection; check for performance impact if > 1
+        self.max_spiral_loops = 1
```
fpagnoux: Should this really be a Simulation instance attribute? Are we considering allowing users to tune their `max_spiral_loops` ?

sandcha: Yes but it's not a hard decision yet. What would you recommend?
Note: It's not documented in the official documentation.

fpagnoux: Mmm not sure. But we can keep it as an instance attribute to leave the option later.

PR #819, opened 2019-01-31
Add optional arguments to add_variables_from_directory

### LINKED ISSUE #807: Allow loading variables from a tax-benefit-system with overlapping variables
Hi there!

### Here is what I did:

I tried to load the variables from `openfisca-ceq` to `openfisca-mali` that do have some overlapping variables with the openfisca-mali tax benefit system. 

I encountered as expected an error.

I think it would be nice  to be able to choose from which variables to choose/update in case of conflicting variables.

Comments and advice very welcome 

## Context

I identify more as a:

benjello:
#### New features

- Introduce `add_variables_from_directory` and `add_variables_from_file`
  - Allows for ignoring, replacing or updating variables from the original TaxBenefitSystem when adding variables from a file or a directory 

Fix #807

### DIFF
--- a/openfisca_core/taxbenefitsystems.py
+++ b/openfisca_core/taxbenefitsystems.py
@@ -151,7 +151,8 @@ class TaxBenefitSystem(object):
         """
         return self.load_variable(variable, update = True)
 
-    def add_variables_from_file(self, file_path):
+    def add_variables_from_file(self, file_path, ignored_variables = None, replaced_variables = None,
+            updated_variables = None):
         """
         Adds all OpenFisca variables contained in a given file to the tax and benefit system.
         """
@@ -173,21 +174,29 @@ class TaxBenefitSystem(object):
             for pot_variable in potential_variables:
                 # We only want to get the module classes defined in this module (not imported)
                 if isclass(pot_variable) and issubclass(pot_variable, Variable) and pot_variable.__module__ == module_name:
-                    self.add_variable(pot_variable)
+                    if ignored_variables is not None and pot_variable.__name__ in ignored_variables:
+                        continue
+                    elif replaced_variables is not None and pot_variable.__name__ in replaced_variables:
+                        self.replace_variable(pot_variable)
+                    elif updated_variables is not None and pot_variable.__name__ in updated_variables:
+                        self.update_variable(pot_variable)
+                    else:
+                        self.add_variable(pot_variable)
         except Exception:
             log.error('Unable to load OpenFisca variables from file "{}"'.format(file_path))
             raise
 
-    def add_variables_from_directory(self, directory):
+    def add_variables_from_directory(self, directory, ignored_variables = None, replaced_variables = None,
+            updated_variables = None):
         """
         Recursively explores a directory, and adds all OpenFisca variables found there to the tax and benefit system.
         """
         py_files = glob.glob(path.join(directory, "*.py"))
         for py_file in py_files:
-            self.add_variables_from_file(py_file)
+            self.add_variables_from_file(py_file, ignored_variables, replaced_variables, updated_variables)
         subdirectories = glob.glob(path.join(directory, "*/"))
         for subdirectory in subdirectories:
-            self.add_variables_from_directory(subdirectory)
+            self.add_variables_from_directory(subdirectory, ignored_variables, replaced_variables, updated_variables)
 
     def add_variables(self, *variables):
         """


benjello:
For a use cas see an example in this companion [PR](https://github.com/openfisca/openfisca-ceq/pull/4/files#diff-1e01928595f592506684e15d76980161)

Morendil:
This PR has several things missing, and may also have too much.

Starting with "too much", the CEQ use case only requires "ignored_variables" so I'm not sure what value we're getting from the additional two options. I would prefer only adding the behaviour that's needed to solve the problem at hand.

In any case, the additional behaviour needs two things:
- unit tests (see further remarks below), not just to ensure that "it works", but to mark the behaviour as officially supported in Core (and incidentally to support debugging)
- for the same reason, updating the docstrings so that the behaviour is officially documented

I know that we don't have an explicit policy about Core contributions with regard to testing and documentation. We've suffered [again](https://github.com/openfisca/openfisca-core/pull/790) and [again](https://github.com/openfisca/openfisca-tunisia/pull/82) from that, so I would prefer to make it explicit.

There are also some issues with code design; the new code in `add_variables_from_file` has high complexity from the if/elif/elif/else statements. High complexity means we will waste time later finding and fixing bugs in that area.

These issues are all related. The design complexity comes because we are working around deep-lying design flaws. We really need **unit** tests, that specify their own fixtures and ideally don't have to touch the file system or rely on external packages. And a lot of the code we have is hard to test at the unit level because variables, for instance, are tightly coupled with the internal details of Python's module system. And integration tests take longer to write, bring in a lot of code, and are brittle because they rely on details of how other packages (such as country-template) are implemented.

I know that because of this, it may seem that it its "too hard" or "too time-consuming" to submit contributions to Core that meet the minimal standards of documentation and tests. And I believe that if we merge contributions not meeting these minimal standards we will waste a lot more time later.

How do we want to handle this trade-off ?

benjello:
@Morendil :
  - I can try to reduce the options to only `ignored_variables`.
  - I can provide a docstring and some tests.
 
I just wanted to know if my need is legit and if it is the right way of fulfill it.

For the more general doctrine question ok for docstring and tests and even docsting tests !

Morendil:
> if my need is legit and if it is the right way of fulfill it.

The need is always legitimate, OF's job is to help users solve problems; and it's possible that there are different ways to solve the same problem, some better in the long term. I think @sandcha has some ideas about how Mali/CEQ fit together and will discuss them with you. I'm always available to discuss design and implementation!

benjello:
I am aware of @sandcha idea of using extensions but did not manage to use it. 
Extensions are ok to add new variables to an existing TaxBenefitSystem with no overlap.
My need is more like a specific kind of reform to load from files.

benjello:
@sandcha @Morendil : we can discuss this wherever you want.
I am for now using some monkey-patching in openfisca-ceq to provide the functionality to my users.

Morendil:
> I am for now using some monkey-patching in openfisca-ceq

That's 👍 as a temporary solution. OK for closing this PR then ?

My constructive suggestions for next steps would be:
 - [ ] change the issue #807 to be more general (it applies to all variable loading not just extensions)
 - [ ] put on our thinking caps and apply [OCP](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle) to instantiating a TaxBenefitSystem

In OO design terms, the problem here is that the TBS constructor is not open to extension, so when we want to change behaviour we have to modify it - by monkey-patching or by changing the actual code, the two amount to the same thing.

It's really the fact that we rely on everything happening in the constructor that makes it hard to apply OCP. As a rule, a constructor should do the least amount of work compatible with having an object in a consistent state. Is it consistent for a TaxBenefitSystem to have no variables at all? It certainly is - it's the same state as an empty container. So there's no reason for the constructor to go loading an entire directory. We want to let client code control more precisely the loading of variables into the model.

It's actually not too hard to make this open to extension, for instance we can:
- **not** load the entire TBS in the constructor (this is the most urgent design flaw)
- use instead optional "policy" objects that can be given to a TBS, for instance an object with a method `accept_variable(variable_definition)` which could inspect candidates variables and accept or reject them based on the need; these policy objects are given by client code (here that would be the init code of `openfisca_ceq`)

The above would be my "definition of done" for a first implementation that satisfies #807. (Plus of course unit tests and documentation!)

benjello:
Closing this PR following @Morendil suggestion and modifying #807.

PR #820, opened 2019-02-01
Fix message output by test runner when a variable is unknown

### LINKED ISSUE #801: Cryptic error message on running a test using an extension
Reported by @guillett after migrating openfisca/openfisca-bacASable to Core v25. (The openfisca-bacASable is a repo representative of various OF extensions.)

```
# openfisca-run-test tests/test_dispositif.yml -e openfisca_bacASable

ERROR:openfisca_core.tools.test_runner:test_dispositif.yml: Test - 2018-01
E
======================================================================
ERROR: unittest.case.FunctionTestCase (check)
----------------------------------------------------------------------
  File "/home/thomas/repos/venv-bas/local/lib/python2.7/site-packages/openfisca_core/tools/test_runner.py", line 143, in check
    _run_test(simulation, test)
  File "/home/thomas/repos/venv-bas/local/lib/python2.7/site-packages/openfisca_core/tools/test_runner.py", line 250, in _run_test
    entity_array = simulation.get_entity(plural = key)  # If key is an entity plural
  File "/home/thomas/repos/venv-bas/local/lib/python2.7/site-packages/openfisca_core/simulations.py", line 481, in get_entity
    return next((entity for entity in self.entities.values() if entity.plural == plural))
StopIteration
```

On investigation we found two distinct problems:
- the error message used to be more explicit (ValueError indicating a variable not found)
- loading the extension silently failed, which results in the variable not being found

**Definition of done**

When I'm verifying with a test verifies that a variable exists in the model and it doesn't
I want to have a clear and not cryptic message stating so,
So that I can stop hitting my head against the wall.

Morendil:
Fixes #801

- Outputs a more helpful message when a variable checked in a test doesn't exist
- Introduces unit tests for the test runners

### DIFF
--- a/openfisca_core/tools/test_runner.py
+++ b/openfisca_core/tools/test_runner.py
@@ -248,11 +248,15 @@ def _run_test(simulation, test):
             for variable_name, value in expected_value.items():
                 _check_variable(simulation, variable_name, value, test.get('period'), test)
         else:
-            entity_array = simulation.get_entity(plural = key)  # If key is an entity plural
-            for entity_id, value_by_entity in expected_value.items():
-                for variable_name, value in value_by_entity.items():
-                    entity_index = entity_array.ids.index(entity_id)
-                    _check_variable(simulation, variable_name, value, test.get('period'), test, entity_index)
+            entity_array = simulation.get_entity(plural = key)
+            if entity_array is not None:  # If key is an entity plural
+                for entity_id, value_by_entity in expected_value.items():
+                    for variable_name, value in value_by_entity.items():
+                        entity_index = entity_array.ids.index(entity_id)
+                        _check_variable(simulation, variable_name, value, test.get('period'), test, entity_index)
+            else:
+                message = "In test '{}', in file '{}', you wanted to check the value of {}, but we don't know a variable with this name."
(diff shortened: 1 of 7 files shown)


openfisca_core/tools/test_runner.py
```diff
@@ -248,11 +248,15 @@ def _run_test(simulation, test):
...
-                    entity_index = entity_array.ids.index(entity_id)
-                    _check_variable(simulation, variable_name, value, test.get('period'), test, entity_index)
+            entity_array = simulation.get_entity(plural = key)
+            if entity_array is not None:  # If key is an entity plural
+                for entity_id, value_by_entity in expected_value.items():
+                    for variable_name, value in value_by_entity.items():
+                        entity_index = entity_array.ids.index(entity_id)
+                        _check_variable(simulation, variable_name, value, test.get('period'), test, entity_index)
+            else:
+                message = "In test '{}', in file '{}', you wanted to check the value of {}, but we don't know a variable with this name."
```
sandcha: Add the name of the loaded tax benefit system as the user could be mistaken on the one he really uses? (country-template vs country specific model)?
> Use similar message for json and yaml errors on unknown variable?

Morendil: Taken care of by re-using VariableNotFound.

PR #826, opened 2019-02-11
Fix end date application

claireleroy:
Hi everyone,

I want to design a reform where all non-contributive taxes and benefits (e.g. income tax, housing benefits, family benefits, social benefits etc.) are neutralized. I want this reform to apply only from a given date.

What I tried to do : 

- The `neutralize_variable` function and the related attribute `is_neutralized` cannot help me with that. They apply for all periods (either a variable is always neutralized either it is never).
- So I tried a solution with the `end` attribute from `Variable`. But it turns out that it does not work for "input" variables. 
Example : *I provide a value for a variable for period `"2019"` and then I code a reform that create an end date for this variable which is equal to `"2018-12-31"`. If I ask OpenFisca to calculate this variable for period `"2019"`, it will return the value I provided. I would like it to return the default value (like if it were neutralized).*

What I propose as a solution : 

- Prevent the computation of a variable after its end date and prevent values to be set for variable after its  end date.

Note that this solution works only because I'm interested in neutralizing variables after a certain date. But if want to neutralize variables between date T and T+1 it won't work for example...

#### Breaking changes

- In `holders.py`: 
  - Treat the variable as if it was neutralized for periods after its end date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 #### Migration guide
 
-This change is not expected to actually break any reusing code, as it seems inconsistent to specify an end date for a variable and use it in computations for periods past that date. However, nothing has previously ruled out relying on that behaviour.
+This change is not expected to actually break much reusing code (if any), as it is unlikely that one would deliberately specify an end date for a variable and use it in computations for periods past that date. However, nothing has previously ruled out relying on that behaviour and it may have been introduced by accident. In particular, it can happen that a variable with a formula and an end date is also, in some applications, used as an input.
 
 ## 26.2.0 [#833](https://github.com/openfisca/openfisca-core/pull/833)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+# 27.0.0 [#826](https://github.com/openfisca/openfisca-core/pull/826)
+
+#### Breaking changes
+
+- No longer honor calls to set_input for variables which have an end date and for periods after that date.
+
+#### Migration guide
+
+This change is not expected to actually break any reusing code, as it seems inconsistent to specify an end date for a variable and use it in computations for periods past that date. However, nothing has previously ruled out relying on that behaviour.
+
 ## 26.2.0 [#833](https://github.com/openfisca/openfisca-core/pull/833)
 
 - Introduce a way to build a simulation for tabular inputs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 #### Breaking changes
 
-- No longer honor calls to set_input for variables which have an end date and for periods after that date.
+- No longer honor calls to `set_input` for variables which have an end date and for periods after that date.
 
 #### Migration guide
 
--- a/openfisca_core/simulation_builder.py
+++ b/openfisca_core/simulation_builder.py
@@ -404,7 +404,11 @@ class SimulationBuilder(object):
                 # Hack to replicate the values in the persons entity
                 # when we have an axis along a group entity but not persons
                 array = np.tile(values, entity.count // len(values))
-                holder.set_input(period_value, array)
+                variable = holder.variable
+                # TODO - this duplicates the check in Simulation.set_input, but
+                # fixing that requires improving Simulation's handling of entities
+                if (variable.end is None) or (period_value.start.date > variable.end):
+                    holder.set_input(period_value, array)
 
     def raise_period_mismatch(self, entity, json, e):
         # This error happens when we try to set a variable value for a period that doesn't match its definition period
--- a/openfisca_core/simulations.py
+++ b/openfisca_core/simulations.py
@@ -478,7 +478,7 @@ class Simulation(object):
 
             If a ``set_input`` property has been set for the variable, this method may accept inputs for periods not matching the ``definition_period`` of the variable. To read more about this, check the `documentation <https://openfisca.org/doc/coding-the-legislation/35_periods.html#automatically-process-variable-inputs-defined-for-periods-not-matching-the-definitionperiod>`_.
         """
-        variable = self.tax_benefit_system.variables[variable_name]
+        variable = self.tax_benefit_system.get_variable(variable_name)
         period = periods.period(period)
         if ((variable.end is not None) and (period.start.date > variable.end)):
             return
--- a/openfisca_core/simulations.py
+++ b/openfisca_core/simulations.py
@@ -461,7 +461,7 @@ class Simulation(object):
         """
         return self.get_holder(variable).get_known_periods()
 
-    def set_input(self, variable, period, value):
+    def set_input(self, variable_name, period, value):
         """
             Set a variable's value for a given period
 
@@ -478,7 +478,11 @@ class Simulation(object):
 
             If a ``set_input`` property has been set for the variable, this method may accept inputs for periods not matching the ``definition_period`` of the variable. To read more about this, check the `documentation <https://openfisca.org/doc/coding-the-legislation/35_periods.html#automatically-process-variable-inputs-defined-for-periods-not-matching-the-definitionperiod>`_.
         """
-        self.get_holder(variable).set_input(period, value)
+        variable = self.tax_benefit_system.variables[variable_name]
+        period = periods.period(period)
+        if ((variable.end is not None) and (period.start.date > variable.end)):
+            return
+        self.get_holder(variable_name).set_input(period, value)
 
     def get_variable_entity(self, variable_name):
 
--- a/openfisca_core/holders.py
+++ b/openfisca_core/holders.py
@@ -7,7 +7,6 @@ import warnings
 
 import numpy as np
 import psutil
-from datetime import date
 
 from openfisca_core import periods
 from openfisca_core.commons import empty_clone
--- a/openfisca_core/holders.py
+++ b/openfisca_core/holders.py
@@ -172,7 +172,7 @@ class Holder(object):
                 self.variable.definition_period,
                 error_message
                 )
-        if self.variable.is_neutralized or ((self.variable.end is not None) and (period.start.date > self.variable.end)):
+        if self.variable.is_neutralized:
             warning_message = "You cannot set a value for the variable {}, as it has been neutralized. The value you provided ({}) will be ignored.".format(self.variable.name, array)
             if sys.version_info < (3, 0):
                 warning_message = warning_message.encode('utf-8')
--- a/openfisca_core/holders.py
+++ b/openfisca_core/holders.py
@@ -7,6 +7,7 @@ import warnings
 
 import numpy as np
 import psutil
+from datetime import date
 
 from openfisca_core import periods
 from openfisca_core.commons import empty_clone
@@ -87,7 +88,7 @@ class Holder(object):
 
             If the value is not known, return ``None``.
         """
-        if self.variable.is_neutralized:
+        if self.variable.is_neutralized or ((self.variable.end is not None) and (period.startd.date > self.variable.end)):
             return self.default_array()
         value = self._memory_storage.get(period, extra_params)
         if value is not None:
--- a/openfisca_core/holders.py
+++ b/openfisca_core/holders.py
@@ -172,7 +172,7 @@ class Holder(object):
                 self.variable.definition_period,
                 error_message
                 )
-        if self.variable.is_neutralized or ((self.variable.end is not None) and (period.startd.date > self.variable.end)):
+        if self.variable.is_neutralized or ((self.variable.end is not None) and (period.start.date > self.variable.end)):
             warning_message = "You cannot set a value for the variable {}, as it has been neutralized. The value you provided ({}) will be ignored.".format(self.variable.name, array)
             if sys.version_info < (3, 0):
                 warning_message = warning_message.encode('utf-8')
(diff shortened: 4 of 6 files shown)


sandcha:
Thank you @claireleroy for this pull request!

Could you tell me more about your use case? 
> For example, did you have other shortcuts to do that previously or is it a fresh new use case?

And, in order to reproduce and analyse the effect of this update, would you share the/some code of your reform?

benjello:
@claireleroy says that a variable with no formula is not ended (means do not return its default value at a period > end and can be set to an arbitrary value) when an end attribute is set. For me it is a bug and not a feature.

claireleroy:
@sandcha Thank your for your reply!

It's the first time that I try to neutralize a variable in a Reform only and for a given period only, so I didn't have shortcuts before.

I'm working on the [TAXIPP](https://www.ipp.eu/outils/taxipp-outils/) microsimulation tool in its 1.0 version for which we did not publish the code yet. Besides the code is on a special environnement (called the [CASD](https://www.casd.eu/)) and it is very complicated for me to export my code out... 

But since my reform is short, I'll rewrite it here : 
```
dispositifs = ['rsa', 'ppa', 'aides_logement', 'irpp', ...and input variables...]

class reform_suppression_dispositifs(Reform):
  
  def apply(self):
    for variable_name in dispositifs:
       self.replace_variable(self.variable[variable_name].__class__)
       variable = self.variables[variable_name]
       variable.end = variable.set_end("2017-12-31")
```

Does it make sense ?

Morendil:
> PS : My tests failed on the CI and I'm surprised to see that the class `TestVariable` has no `end` attribute unless `Variable`... Is it normal ?

"Normal" would be too strong a word. But it needs a long explanation, given below. :)

The TLDR is "this is because there is a lot of latent technical debt in OpenFisca, and while we fix that our unit tests will occasionally slow us down, and **that's a good thing**".

When writing unit tests we often want to decouple the various parts of the "system under test" (standard jargon for "whatever code we are testing", often abbreviated SUT).

For instance if the unit test wants to use a TaxBenefitSystem, it can be painful to use a real one, because building a TaxBenefitSystem (currently) reads directories on the disk, calls the Python module loader, and a bunch of other things which cost a lot of CPU time. Because we (typically) create a new instance for every unit test, we end up with slow tests. (Also called "integration tests".)

The best technique I know to deal with this is to decouple the dependency itself: the SUT in this case is SimulationBuilder, which doesn't *need* an entire TaxBenefitSystem. For the particular test which fails, it needs only an Entity instance and some JSON. We should give it "[only the money and not our pants](http://www.arolla.fr/blog/2017/02/principes-solid-vie-de-jours/)" to borrow a metaphor for the Law of Demeter.

The problem is that we can't give it a "real" Entity instance, because the constructor of Entity takes a Simulation object, and the Simulation's constructor takes a TaxBenefitSystem object. This means we're at risk of having slow tests, again. So in that unit test we do the next best technique: we use so-called "[mock objects](https://www.agilealliance.org/glossary/mocks)" (a trick invented by my Extreme Programmer friends Tim, Steve and Phlip around 2000).

Sometimes your mock objects will return *other* mock objects, like the mock Entity in that test returning a mock Variable. I don't necessarily give a mock object all the attributes of the real one: one reason is that it's preferable to invoke methods on objects (which are part of their public interface contract) rather than directly query their attributes (which are an internal implementation detail). So the TestVariable mock object includes only the strict minimum attributes that were required to pass the unit tests for SimulationBuilder. (There's already too many of them for my taste.)

We have a test which fails because **it used to be possible to call set_input on a variable without accessing its end attribute**, and your change implies that it's no longer possible and that **set_input now relies on that attribute being part of the interface contract of Variable**. The test failing is a good thing; it reveals that these decisions are in conflict.

Now we make the choice consciously of whether we want to strongly couple one more Variable attribute to the internal implementation of `Holder`s. **Spoiler: I think that's not a desirable coupling.**    To me there's a strong case for making the change elsewhere (Variable seems like the best candidate), if this is the right change to make.

It's also possible that this isn't the right change to make, or that there are alternatives; but that's for tomorrow, @sandcha and I may have a few more questions. :)

claireleroy:
@Morendil @sandcha 

Did you have time to look at my PR ? If you have more questions or if I can help do not hesitate :)

benjello:
@Morendil @sandcha : we are about to move part of tech code base to CASD and we would like to have as much PR accepted as possible before doing the manual synchro ...

Could we converge on this fatser.

Morendil:
@benjello We're doing our best, and we have lots of stuff to do. We're not forgetting this PR, but even though it's a one-line change, that one line has implications that require careful thinking and we need time for that.

My feeling is that `holders.py` is not the right place for this change, and to find the right place for this change we'll need to do more refactoring of Holders first, which means adding unit tests first. And we need anyway to adjust the existing unit tests that this PR breaks.

My best idea would be to make this change manually in CASD for your immediate needs since it's a one-line change.

Morendil:
> we'll need to do more refactoring of Holders first

More specifically:
- [ ] move the following behaviours to Simulation instead:
  - [ ] is_neutralized check
  - [ ] end date check (per this PR)
  - [ ] "simple" dispatch and divide (see below)
  - [ ] use regular values for `set_input` variable attributes instead of function pointers
- [ ] make dispatch the default (see #813)
- [ ] move the following behaviours to SimulationBuilder instead:
  - [ ] expression evaluation (only makes sense for individual values, not vectors)
  - [ ] divide with "leftovers" (i.e. set 2015 to 10K, set two months of 2015 to 2K each, each remaining month should be 1K)
- [ ] make sure each behaviour has a proper unit test

claireleroy:
@Morendil I understand that there is a lot of things to do... what could I do to make this PR progress ? Or should I wait ?

Morendil:
@claireleroy The above list is what I learned after I made a first POC on a branch, so I have a somewhat clearer idea now. (At first I thought the set_input behaviours should be on SimulationBuilder, but I learned that doesn't work great.) We can scope it down to just implementing the end date check and a unit test for it if you want to tackle that, and implement the rest in a later PR.

The end date check should be in `set_input` of the `Simulation` class. For a unit test we can only use the variable `housing_allowance` in the country template. There is no need for a corresponding check in the getter, since `Simulation::calculate` already returns the default value for inputs that have not been defined.

I wouldn't bother to issue a warning like we do for neutralized variables. There are only three possibilities:
- what makes sense is to honor all `set_input` calls on periods after the end date of a variable; then this PR shouldn't be merged at all;
- what makes sense is to ignore all `set_input` calls on periods after the end date of a variable; then the change above is sufficient, and no warning is needed;
- what makes sense depends on the situation; then what we need is to specify when it makes sense and when it doesn't, and to make it configurable (for instance with an optional parameter to set_input).

If you prefer option 3, the default should initially be "whatever we did previously", i.e. to honor `set_input` calls after the end date.

claireleroy:
@Morendil Sorry for the delay in my response.. There are some things that I don't understand in your comments mainly I think because I'm reaching my technical limits as an economist and a non-developper. Maybe @benjello will take over with this PR !

What I do know is that the initial point of my PR was to find a way to "neutralize" a bunch of taxes and benefits from a given date. I couldn't do it with the `neutralize_variable` attribute so I just tried to find a way and I tried with the `end` attribute. So actually, if there is third way that you prefer, I'm open to it.

Let's say we stick with my way. From the options you've listed, I prefer **option 2**, because I don't see why the `end` date should apply differently to an input variable and to a formula variable.  Right now : 
- when a variable has a formula, it will always return default value if you try to calculate it after its end date
- if it's an input variable and you set a value via `set_input`, the end date can somehow be ignored. 

I find it weird and I propose to treat both variables the same way.

Morendil:
@claireleroy Option 2 works, go for it :)

My remaining concern is that we treat *start* dates differently, by encoding them in the formula's function name. That means an input can never have a start date. But we can solve this later.

claireleroy:
@Morendil Ok perfect ! What should be the next steps then ? Rebase ?

Morendil:
Next steps:
- [ ] Move the end date check to Simulation
- [ ] Add unit test (verify that set_input has no effect on periods past an input's end date)

There are no inputs with an end date in the country-template package so the unit test needs to add its own input variable, the best place is probably in `tests/core/test_variables.py`, taking inspiration from the tests in the section `# end, no formula` (a mix of `test_variable__end_attribute` and `test_call__end_attribute__one_simple_formula`)

I'll take care of rebasing and setup/CHANGELOG

benjello:
@Morendil : so if I understand correctly your request i should modify `simulation.get_array` and `simulation.set_input` to deal with the end.date.

Should I also do that for the neutralisation part ?

Morendil:
@benjello No need to modify `get_array`, I think, just `set_input`.

The goal of this PR was to add the end date check, so I think we can move the check for neutralized variables in a later PR.

benjello:
@Morendil : is everything ok now ?

openfisca_core/simulations.py
```diff
@@ -470,7 +470,11 @@ def set_input(self, variable, period, value):
 
             If a ``set_input`` property has been set for the variable, this method may accept inputs for periods not matching the ``definition_period`` of the variable. To read more about this, check the `documentation <https://openfisca.org/doc/coding-the-legislation/35_periods.html#automatically-process-variable-inputs-defined-for-periods-not-matching-the-definitionperiod>`_.
         """
-        self.get_holder(variable).set_input(period, value)
+        variable = self.tax_benefit_system.variables[variable_name]
```
Morendil:
```suggestion
        variable = self.tax_benefit_system.get_variable(variable_name)
```

Morendil:
Je prends le rebase et merge.

claireleroy:
@Morendil Thank you for your help with this PR ! And @benjello thank you for taking over ! 

I'm glad it's going to be merged :) it will really help us at IPP !

Morendil:
@claireleroy @benjello I've pinged @fpagnoux for reviewing the changelog text and version change, please hold off on merging until he's had a chance to look at it.

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
 
+# 27.0.0 [#826](https://github.com/openfisca/openfisca-core/pull/826)
+
+#### Breaking changes
+
+- No longer honor calls to set_input for variables which have an end date and for periods after that date.
+
+#### Migration guide
+
+This change is not expected to actually break any reusing code, as it seems inconsistent to specify an end date for a variable and use it in computations for periods past that date. However, nothing has previously ruled out relying on that behaviour.
```
fpagnoux: Meh about the "it seems inconsistent to...". This is quite judgmental, and I was a user who ended up needing to migrate code, this sentence would sound bitter to me.  
And, to be a little sassy, I could reply to this sentence by "It seems a little inconsistent to explicitly specify inputs for  a variable, and to want these inputs to be ignored", which is the goal of this PR 😛

benjello: @fpagnoux : when the end.date change comes from a reform and you want to use the same initialization data, this is clearly what you end doing !

Morendil: @fpagnoux I agree with you… even though I wrote those words. I'll rephrase to be more congruent with that inner conflict. "I am large, I contain multitudes."

Morendil:
@benjello No, sorry. To be quite safe I ran France tests on this branch, and one of them fails (`paje_clca` has an end date and is apparently used as an input past that date). I'll investigate later today, because that's going to be good information on "how to migrate" but I need to understand how the problem manifests.

Morendil:
`tests/formulas/cmuc_fiab.yaml` was failing with a strange-looking message:

> Inconsistent input: variable paje_clca has already been set for all months contained in period 2018-04, and value [469.75] provided for 2018-04 doesn't match the total ([0.]). This error may also be thrown if you try to call set_input twice for the same variable and period.

This was because SimulationBuilder was calling `Holder.set_input` directly, bypassing the Simulation. Fixing that makes the test fail in a more normal way.

Replacing `paje_clca` with `paje_prepare` fixed the test.

There is another failure in `tests/formulas/revenu_disponible.yaml`, investigating.

Morendil:
The test on `revenu_disponible` was failing because we were using `f3sa` (one of the tax returns variables) in a 2018 test `pfu_hors_av_epargne_non_solidaire_etats_non_cooperatifs_2018` with an end date of 2012 for that variable. Removing the end date gets the test to pass, and since that is the only test using that variable, that might be enough of a fix.

@bfabre01 you wrote that test apparently, @claireleroy you last touched the line with the end date, so if either of you have any domain insights on this they will be appreciated :)

claireleroy:
@Morendil I've juste checked this `f3sa` variable : apparently it was reintroduced since the 2016 income tax, so it's a mistake in the Openfisca-France code (that I'm surely responsible for). I will correct it in France

openfisca_core/simulation_builder.py
```diff
@@ -404,7 +404,11 @@ def finalize_variables_init(self, entity):
                 # Hack to replicate the values in the persons entity
                 # when we have an axis along a group entity but not persons
                 array = np.tile(values, entity.count // len(values))
-                holder.set_input(period_value, array)
+                variable = holder.variable
+                # TODO - this duplicates the check in Simulation.set_input, but
+                # fixing that requires improving Simulation's handling of entities
+                if (variable.end is None) or (period_value.start.date > variable.end):
```
Morendil: This:
`if (variable.end is None) or (period_value.start.date > variable.end):`
should of course have been
`if (variable.end is None) or (period_value.start.date < variable.end):`
…and the error breaks all input handling for variables with an end date.

PR #827, opened 2019-02-12
Optimize set_input_dispatch_by_period

fpagnoux:
Connected to #813

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-### 25.3.4 [???](https://github.com/openfisca/openfisca-core/pull/???)
+### 25.3.4 [827](https://github.com/openfisca/openfisca-core/pull/827)
 
 - Optimize `set_input_dispath_by_period` so that it doesn't create vecor duplicated in memory
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 25.3.4 [???](https://github.com/openfisca/openfisca-core/pull/???)
+
+- Optimize `set_input_dispath_by_period` so that it doesn't create vecor duplicated in memory
+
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+### 25.3.4 [827](https://github.com/openfisca/openfisca-core/pull/827)
+
+- Optimize `set_input_dispath_by_period` so that it doesn't create vecor duplicated in memory
```
Morendil: > vecor -> vector

tests/core/test_holders.py
```diff
@@ -109,6 +110,16 @@ def test_get_memory_usage_with_trace():
     assert_equal(memory_usage['nb_requests_by_array'], 1.25)  # 15 calculations / 12 arrays
 
 
+def test_set_input_dispath_by_period():
```
Morendil: test_set_input_dispath_by_period -> test_set_input_dispatch_by_period

tests/core/test_holders.py
```diff
@@ -109,6 +110,16 @@ def test_get_memory_usage_with_trace():
...
+    simulation = get_simulation(single)
+    variable = simulation.tax_benefit_system.get_variable('housing_occupancy_status')
+    entity = simulation.household
+    holder = Holder(variable, entity)
+    set_input_dispatch_by_period(holder, make_period(2019), 'owner')
```
Morendil:
```suggestion
    set_input_dispatch_by_period(holder, make_period(2019), np.asarray(['owner']))
```

It seems needlessly confusing to pass a scalar value to a method whose signature makes it clear it expects an array, and the test's pass/fail status does not as far as I can tell depend on that.

It could also make sense to use a non-Enum variable for this test, as they are the exception rather than the norm in OpenFisca, and using an exceptional case hints that the behaviour under test is restricted to that exceptional case.

PR #829, opened 2019-02-18
Update autopep8 and flake8

maukoquiroga:
#### Technical changes

- Update autopep8 and flake8, which in particular now enforce rules W504 and W605
  - W504 goes against house style, so we add it to ignored rules
  - W605 may seem like an [overreach](https://github.com/PyCQA/pycodestyle/issues/755) but does make sense (additional details in PR description), so we upgrade a few regexes to raw strings

#### Detailed explanation (W605)

in Python strings the character "\" normally introduces an escape sequence, and should be followed by only a few select characters (\n is newline, \t is tab, and there are a few more esoteric ones) - anything else, for instance "\d", is an "illegal escape sequence" and should normally require a double escape (i.e. "\\d").

However, in regular expressions (where "\d" is a common way to say "any digit from 0 to 9") this would be quite burdensome, and Python in fact "looks the other way" when you use an illegal escape sequence, it interprets the characters literally. So you *don't* have to use a double escape in regular expressions, but that's a bit of a hack.

So W605 is in fact the right thing, and preserves the "principle of least surprise", at the cost of an extra syntax: that of "[raw strings](https://stackoverflow.com/questions/2081640/what-exactly-do-u-and-r-string-flags-do-and-what-are-raw-string-literals)" denoted by the string prefix `r`. Like so: `r"\d"` - this is a string containing the characters "\" and "\d". It is exactly the same string as `"\d"`, but that is _explainable_, unlike `"\d"` which should generate an error if the Python parser applied its own specification strictly.

### DIFF
--- a/openfisca_core/periods.py
+++ b/openfisca_core/periods.py
@@ -21,7 +21,7 @@ MONTH = 'month'
 YEAR = 'year'
 ETERNITY = 'eternity'
 
-INSTANT_PATTERN = re.compile('^\d{4}(?:-\d{1,2}){0,2}$')  # matches '2015', '2015-01', '2015-01-01'
+INSTANT_PATTERN = re.compile(r'^\d{4}(?:-\d{1,2}){0,2}$')  # matches '2015', '2015-01', '2015-01-01'
 
 
 def N_(message):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Changelog
 
-## 26.0.4 [#825](https://github.com/openfisca/openfisca-core/pull/825)
+### 26.0.5 [#829](https://github.com/openfisca/openfisca-core/pull/829)
+
+- Update autopep8 and flake8, which in particular now enforce rules W504 and W605
+  - W504 goes against house style, so we add it to ignored rules
+  - W605 may seem like an [overreach](https://github.com/PyCQA/pycodestyle/issues/755) but does make sense (additional details in PR description), so we upgrade a few regexes to raw strings
+
+### 26.0.4 [#825](https://github.com/openfisca/openfisca-core/pull/825)
 
 - Fixes regression introduced by Core v25 when running YAML tests with reforms or extensions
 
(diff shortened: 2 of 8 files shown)


openfisca_core/holders.py
```diff
@@ -211,8 +211,7 @@ def _set(self, period, value, extra_params = None):
         if self.variable.definition_period != ETERNITY:
             if period is None:
                 raise ValueError('A period must be specified to set values, except for variables with ETERNITY as as period_definition.')
-            if ((self.variable.definition_period == MONTH and period.unit != periods.MONTH) or
-               (self.variable.definition_period == YEAR and period.unit != periods.YEAR)):
+            if self._is_not_month(period) or self._is_not_year(period):
```
Morendil: This PR is ostensibly for a minor upgrade of our linter, I'd prefer if the changes we make as a result of that were strictly the minimum necessary to pass the new rules: here for instance, moving the "or" from the end of one line to the beginning of the next.

Refactoring the code in `holders.py` is a worthwhile exercises and I'm very much in favor of allocating some time to that, to be done in pairs or mobs so we can get better aligned on our sense of good code design. Let's do that, with greater focus, as a task separate from upgrading linters.

openfisca_core/periods.py
```diff
@@ -27,7 +27,7 @@
 YEAR = 'year'
 ETERNITY = 'eternity'
 
-INSTANT_PATTERN = re.compile('^\d{4}(?:-\d{1,2}){0,2}$')  # matches '2015', '2015-01', '2015-01-01'
+INSTANT_PATTERN = re.compile(r'^\d{4}(?:-\d{1,2}){0,2}$')  # matches '2015', '2015-01', '2015-01-01'
```
Morendil: I have no idea why this "r" prefix is needed, why a minor linter update is requiring this change, or what I should do in the future to stay compliant; I'd rather not accept this PR without understanding what's going on. This does not necessarily require a code change, but definitely some explanation in the PR description.

Morendil:
Rebased.

Morendil:
Updated PR and description to address both of my requested changes, so I could approve and merge.

openfisca_core/holders.py
Morendil: On reflection, it seemed best to add W504 to ignored rules, as we already have W503 in there for the same reason: "house rule is to break after binary operators".

PR #833, opened 2019-02-27
Build simulation with tabular inputs

sandcha:
Connected to openfisca/openfisca-doc#168

#### New features

- Introduce tabular oriented simulation building
  - Allow to create a simulation step by step:
    1. Create entities with `SimulationBuilder.create_entities(...)`
    2. Declare your population 
        with `SimulationBuilder.declare_person_entity(...)` and `SimulationBuilder.declare_entity(...)`
    3. Link between persons in population definition with `SimulationBuilder.join_with_persons(...)`
    4. Build a simulation with `SimulationBuilder.build(...)`
    5. Set variable values with the already existing `Simulation.set_input(...)` method

---
#### Here is a an example with a CSV dataset:
 
For this `data_persons.csv`:
```csv
person_id,household_id,role_in_household,salary,age
1,a,first_parent,2694,40
2,a,second_parent,2720,43
3,b,first_parent,3865,45
4,b,child,1300,23
5,c,child,0,12
6,c,child,0,14
7,c,first_parent,2884,44
12,e,second_parent,1200,38
8,d,first_parent,3386,27
9,d,second_parent,2929,28
10,e,child,0,10
11,e,trolol,1600,35
```
And this `data_households.csv`:
```csv
household_id,rent,accommodation_size
b,1200,64
a,700,39
d,750,31
e,840,37
c,1100,68
```
You can calculate country_template `total_taxes` with this code:
```py
# -*- coding: utf-8 -*-

import pandas

data_persons = pandas.read_csv('./data_persons.csv')
data_households = pandas.read_csv('./data_households.csv')

persons_ids = data_persons.person_id
households_ids = data_households.household_id
persons_households = data_persons.household_id
persons_households_roles = data_persons.role_in_household

tbs = CountryTaxBenefitSystem()
sb = SimulationBuilder()

sb.create_entities(tbs)

sb.declare_person_entity('person', persons_ids)
household_instance = sb.declare_entity('household', households_ids)

# Join csv data on persons (roles are optional):
sb.join_with_persons(household_instance, persons_households, roles = persons_households_roles)

simulation = sb.build(tbs)
period = '2019-02'
simulation.set_input('salary', period, data_persons.salary)  
total_taxes = simulation.calculate('total_taxes', period)

print(data_persons)
print(data_households)
print("Household ids", household_instance.ids)
print("total_taxes", total_taxes)
```

Any comments on methods names/signatures are more than welcome! 🙂

### DIFF
--- a/openfisca_core/entities.py
+++ b/openfisca_core/entities.py
@@ -4,6 +4,7 @@ import traceback
 import warnings
 import textwrap
 from os import linesep
+from typing import Iterable
 
 import numpy as np
 
@@ -308,8 +309,14 @@ class GroupEntity(Entity):
         return self._members_role
 
     @members_role.setter
-    def members_role(self, members_role):
-        self._members_role = members_role
+    def members_role(self, members_role: Iterable):
+        if members_role is not None:
+            self._members_role = np.array(list([self.get_role(role_name) for role_name in members_role]))
+
+    def get_role(self, role_name):
+        for possible_role in self.roles:
+            if possible_role.key == role_name:
+                return possible_role
 
     @members_position.setter
     def members_position(self, members_position):
@@ -451,9 +458,7 @@ class GroupEntity(Entity):
             If ``role`` is provided, only the entity member with the given role are taken into account.
         """
         if role:
-            # role_condition = self.members.has_role(role)
             role_condition = self.members_role == role
-            print(role_condition)
             return self.sum(role_condition)
         else:
             return np.bincount(self.members_entity_id)
--- a/openfisca_core/entities.py
+++ b/openfisca_core/entities.py
@@ -15,6 +15,21 @@ ADD = 'add'
 DIVIDE = 'divide'
 
 
+class Role(object):
+
+    def __init__(self, description, entity):
+        self.entity_class = entity
+        self.key = description['key']
+        self.label = description.get('label')
+        self.plural = description.get('plural')
+        self.doc = textwrap.dedent(description.get('doc', ""))
+        self.max = description.get('max')
+        self.subroles = None
+
+    def __repr__(self):
+        return "Role({})".format(self.key)
+
+
 class Entity(object):
     """
         Represents an entity (e.g. a person, a household, etc.) on which calculations can be run.
@@ -305,16 +320,17 @@ class GroupEntity(Entity):
     @property
     def members_role(self):
         if self._members_role is None:
-            self._members_role = np.repeat(-1, len(self.members_entity_id))
+            default_role = self.flattened_roles[0]
+            self._members_role = np.repeat(default_role, len(self.members_entity_id))
         return self._members_role
 
     @members_role.setter
-    def members_role(self, members_role: Iterable):
+    def members_role(self, members_role: Iterable[Role]):
         if members_role is not None:
-            self._members_role = np.array(list([self.get_role(role_name) for role_name in members_role]))
+            self._members_role = np.array(list(members_role))
 
     def get_role(self, role_name):
-        for possible_role in self.roles:
+        for possible_role in self.flattened_roles:
             if possible_role.key == role_name:
                 return possible_role
 
@@ -458,7 +474,10 @@ class GroupEntity(Entity):
             If ``role`` is provided, only the entity member with the given role are taken into account.
         """
         if role:
-            role_condition = self.members_role == role
+            if role.subroles:
+                role_condition = np.logical_or.reduce([self.members_role == subrole for subrole in role.subroles])
+            else:
+                role_condition = self.members_role == role
             return self.sum(role_condition)
         else:
             return np.bincount(self.members_entity_id)
@@ -546,21 +565,6 @@ class GroupEntity(Entity):
         self.check_role_validity(role)
         nb_persons_per_entity = self.nb_persons(role)
         return self.project(array / nb_persons_per_entity, role = role)
-
-
-class Role(object):
-
-    def __init__(self, description, entity):
-        self.entity_class = entity
-        self.key = description['key']
-        self.label = description.get('label')
-        self.plural = description.get('plural')
-        self.doc = textwrap.dedent(description.get('doc', ""))
-        self.max = description.get('max')
-        self.subroles = None
-
-    def __repr__(self):
-        return "Role({})".format(self.key)
 
 
 class Projector(object):
--- a/openfisca_core/entities.py
+++ b/openfisca_core/entities.py
@@ -330,9 +330,7 @@ class GroupEntity(Entity):
             self._members_role = np.array(list(members_role))
 
     def get_role(self, role_name):
-        for possible_role in self.flattened_roles:
-            if possible_role.key == role_name:
-                return possible_role
+        return next((role for role in self.flattened_roles if role.key == role_name), None)
 
     @members_position.setter
     def members_position(self, members_position):
--- a/openfisca_core/entities.py
+++ b/openfisca_core/entities.py
@@ -193,11 +193,11 @@ class PersonEntity(Entity):
             >>> array([False])
         """
         self.check_role_validity(role)
-        entity = self.simulation.get_entity(role.entity_class)
+        group_entity = self.simulation.get_entity(role.entity_class)
         if role.subroles:
-            return np.logical_or.reduce([entity.members_role == subrole for subrole in role.subroles])
+            return np.logical_or.reduce([group_entity.members_role == subrole for subrole in role.subroles])
         else:
-            return entity.members_role == role
+            return group_entity.members_role == role
 
     @projectable
     def value_from_partner(self, array, entity, role):
@@ -271,16 +271,12 @@ class GroupEntity(Entity):
 
     def __init__(self, simulation, persons = None):
         Entity.__init__(self, simulation)
-        self.members_entity_id = None
+        self._members_entity_id = None
         self._members_role = None
         self._members_position = None
         self.members = persons
         self._roles_count = None
         self._ordered_members_map = None
-
-    @property
-    def members_role(self):
-        return self._members_role
 
     @property
     def members_position(self):
@@ -296,6 +292,20 @@ class GroupEntity(Entity):
                 counter_by_entity[entity_index] += 1
 
         return self._members_position
+
+    @property
+    def members_entity_id(self):
+        return self._members_entity_id
+
+    @members_entity_id.setter
+    def members_entity_id(self, members_entity_id):
+        self._members_entity_id = members_entity_id
+
+    @property
+    def members_role(self):
+        if self._members_role is None:
+            self._members_role = np.repeat(-1, len(self.members_entity_id))
+        return self._members_role
 
     @members_role.setter
     def members_role(self, members_role):
@@ -441,7 +451,9 @@ class GroupEntity(Entity):
             If ``role`` is provided, only the entity member with the given role are taken into account.
         """
         if role:
-            role_condition = self.members.has_role(role)
+            # role_condition = self.members.has_role(role)
+            role_condition = self.members_role == role
+            print(role_condition)
             return self.sum(role_condition)
         else:
             return np.bincount(self.members_entity_id)
--- a/openfisca_core/simulation_builder.py
+++ b/openfisca_core/simulation_builder.py
@@ -189,8 +189,8 @@ class SimulationBuilder(object):
     def nb_persons(self, entity_singular, role = None):
         return self.entities_instances[entity_singular].nb_persons(role = role)
 
-    def join_with_persons(self, group_instance, group_to_person_projection, roles: Iterable[str] = None):
-        group_sorted_indices = np.unique(group_to_person_projection, return_inverse = True)[1]
+    def join_with_persons(self, group_instance, persons_group_assignment, roles: Iterable[str] = None):
+        group_sorted_indices = np.unique(persons_group_assignment, return_inverse = True)[1]
         group_instance.members_entity_id = np.argsort(group_instance.ids)[group_sorted_indices]
 
         if roles is not None:
--- a/openfisca_core/simulation_builder.py
+++ b/openfisca_core/simulation_builder.py
@@ -196,6 +196,9 @@ class SimulationBuilder(object):
     def nb_persons(self, entity_singular):
         return self.entities_instances[entity_singular].nb_persons()
 
+    def build(self, tax_benefit_system):
+        return Simulation(tax_benefit_system, entities_instances = self.entities_instances)
+
     def bind(self, entity_plural, entity_ids, entity_persons_ids):
         if self.persons_plural is None:
             raise SituationParsingError(entity_plural, 'Unable to link {0} entity to undefined persons.')
--- a/openfisca_core/simulation_builder.py
+++ b/openfisca_core/simulation_builder.py
@@ -19,7 +19,6 @@ class SimulationBuilder(object):
     def __init__(self):
         self.default_period = None  # Simulation period used for variables when no period is defined
         self.persons_plural = None  # Plural name for person entity in current tax and benefits system
-        self.person_singular = None
 
         # JSON input - Memory of known input values. Indexed by variable or axis name.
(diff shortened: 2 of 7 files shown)


tests/core/test_simulation_builder.py
```diff
@@ -349,9 +351,85 @@ def test_unallocated_person(simulation_builder, group_entity):
     assert exception.value.error == {'households': "{'Sarah'} have been declared in persons, but are not members of any household. All persons must be allocated to a household."}
 
 
+def test_nb_persons_in_group_entity(simulation_builder):
+    persons_ids: Iterable = [2, 0, 1, 4, 3]
```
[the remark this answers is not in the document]
Morendil: @benjello Thanks for asking, we've started using this and they're a great tool which will be even more useful if we take time to communicate and explain :)

These are type hints, a new feature since Python 3, see here: https://docs.python.org/3/library/typing.html

We're using the type hints to document the code internally in places where we would otherwise wonder too often "what Python type is this supposed to be".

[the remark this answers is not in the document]
benjello: I do not see this exact syntax in the link you provided. So what are the effect of this line exactly ?
Is it different than 
```python
persons_ids = [2, 0, 1, 4, 3]
```
?

openfisca_core/taxbenefitsystems.py
```diff
@@ -74,6 +76,15 @@ def base_tax_benefit_system(self):
             self._base_tax_benefit_system = base_tax_benefit_system = baseline.base_tax_benefit_system
         return base_tax_benefit_system
 
+    def instantiate_entities(self):
+        person_instance = self.person_entity(None)
+        entities_instances: Dict[Entity.key, Entity] = {person_instance.key: person_instance}
```
fpagnoux: > What does `Entity.key` mean in a type annotation? I would have expected `Dict[String, Entity]`, but I'm not familiar with them. Any reference appreciated 🙂

openfisca_core/entities.py
```diff
@@ -441,7 +474,10 @@ def nb_persons(self, role = None):
             If ``role`` is provided, only the entity member with the given role are taken into account.
         """
         if role:
-            role_condition = self.members.has_role(role)
```
fpagnoux: Not sure I understand the reason of this change. Why can't we keep using the `has_role` method?

openfisca_core/entities.py
```diff
@@ -297,9 +309,30 @@ def members_position(self):
 
         return self._members_position
 
+    @property
+    def members_entity_id(self):
```
fpagnoux: > Not sure about the interest of using a property if we don't have any logic in the getter and setter. Is it just for consistency with the other attributes?

openfisca_core/entities.py
```diff
@@ -297,9 +309,30 @@ def members_position(self):
...
+        return self._members_entity_id
+
+    @members_entity_id.setter
+    def members_entity_id(self, members_entity_id):
+        self._members_entity_id = members_entity_id
+
+    @property
+    def members_role(self):
+        if self._members_role is None:
+            default_role = self.flattened_roles[0]
```
fpagnoux: Ok, now I understand better how we can make `role`initialization optional: we are just giving the first role to everyone by default.

I'm not sure whether that make sense: we could easily end up with situations with 5 parents in a family, despite having an entity definition that specifies a max amount to 2... 

Is there really a user demand for a way to initialize simulations with group entities, but without roles? If not, I'd rather just have the `members_role` be mandatory and not make any assumption.

openfisca_core/simulation_builder.py
```diff
@@ -159,9 +165,37 @@ def build_default_simulation(self, tax_benefit_system, count = 1, **kwargs):
             entity.ids = np.array(range(count))
             if not entity.is_person:
                 entity.members_entity_id = entity.ids  # Each person is its own group entity
-                entity.members_role = entity.filled_array(entity.flattened_roles[0])
         return simulation
 
+    def create_entities(self, tax_benefit_system):
+        self.entities_instances = tax_benefit_system.instantiate_entities()
+
+    def declare_person_entity(self, person_singular, persons_ids: Iterable):
```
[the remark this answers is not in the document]
Morendil: @fpagnoux We've used type hints to clarify variables and parameters experimentally, where we thought there was a substantial risk of getting confused. It's not clear how far we should go with type annotations, maybe this is something we could discuss on a "policy" level after this PR and debriefing what we've learned.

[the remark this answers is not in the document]
Morendil: @fpagnoux When naming methods that refer to types in an object system I like to make things consistent by pluralizing class names when referring to collections of instances of that class.

So, Entity in the singular refers to (a method that impacts) a single instance of the class Entity. I'd use the term "entities" when manipulating a list or a dict of instances of Entity. To me a single person is definitely not "an entity" (for the purposes of naming in the OpenFisca code).

openfisca_core/simulation_builder.py
```diff
@@ -159,9 +165,37 @@ def build_default_simulation(self, tax_benefit_system, count = 1, **kwargs):
...
+
+    def nb_persons(self, entity_singular, role = None):
+        return self.entities_instances[entity_singular].nb_persons(role = role)
+
+    def join_with_persons(self, group_instance, group_to_person_projection, roles: Iterable[str] = None):
+        group_sorted_indices = np.unique(group_to_person_projection, return_inverse = True)[1]
+        group_instance.members_entity_id = np.argsort(group_instance.ids)[group_sorted_indices]
+
+        if roles is not None:
+            group_instance.members_role = [group_instance.get_role(role_name) for role_name in roles]
```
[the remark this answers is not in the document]
Morendil: @fpagnoux I don't think we have a dire need to optimize this as it's a one-time setup cost. The `np.select` version strikes me as less readable (maybe because I'm used to seeing it in the context of formulas, doing a different job). I'll do a quick performance measurement to assess the trade-off.

openfisca_core/taxbenefitsystems.py
Morendil: @fpagnoux I feel a bit naughty about this, but it's for a good cause. We considered `Dict[str, Entity]` which is technically correct, but since these annotations are not checked, we can be more "poetic": so we use `Dict[Entity.key, Entity]` to distinguish from `Dict[Entity.plural, Entity]` because that's one of the most likely confusions. It's possible that the syntax is itself a source of confusion and surprise, and thus the decision is a net negative.

openfisca_core/entities.py
Morendil: @fpagnoux We are refactoring towards code that obeys the Law of Demeter, to reduce the entanglement of dependencies which, to me, is a main component of our technical debt.

The state of play before this PR is that `nb_persons` of `GroupEntity` is implemented via `has_role` of `PersonEntity` (for the branch where a role is supplied). This creates a dependency `GroupEntity -> PersonEntity`. But `has_role` of `PersonEntity` is implemented via a dependency on `Simulation`, to get the `GroupEntity` corresponding to the role, and finally to access `entity.members_role` which should be an internal implementation detail of GroupEntity. We have created a dependency `PersonEntity -> Simulation -> GroupEntity`.

Composing these dependencies we get `GroupEntity -> PersonEntity -> Simulation -> GroupEntity`, a circular dependency. This is one of the reasons, based on my experience, why things are hard to unit test.

After this PR, `nb_persons` can be called on a `GroupEntity` and do its job without relying on these collaborators, using exclusively the information it has and which it owns, `members_role`. This is easy to unit test, more maintainable.

I think it should also be possible to improve `has_role` (because of the dependency on `Simulation`, which seems very awkward to me) but that's a later topic.

openfisca_core/entities.py
Morendil: @fpagnoux Consistency, and easier debugging later. Using a property makes it easier to detect who is accessing this attribute. It makes sense to encapsulate this attribute somewhat "eagerly" because it's clearly an implementation detail that has its own structure and must be kept in sync with others, not something it's natural to expose to the world (like e.g. a `name`).

openfisca_core/entities.py
Morendil: This was mostly for unit testing purposes, the earlier tests that built partial functionality could work without roles (and still can, if you pass `None`). I've made it mandatory. We are still making use of the lazy initialization to default roles for the `build_default_simulation` case, however.

openfisca_core/simulation_builder.py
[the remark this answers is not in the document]
Morendil: Unless I've misunderstood something, the `np.select` approach is actually a lot slower.

```
>>> flattened_roles = ['b','c','a']
>>> roles = np.repeat('a',10000000)
>>> start = time.time(); indices = np.select([roles == role for role in flattened_roles], flattened_roles); duration = time.time() - start
>>> duration
33.49795985221863
>>> start = time.time(); indices = [flattened_roles.index(role_name) for role_name in roles]; duration = time.time() - start
>>> duration
9.044532775878906
```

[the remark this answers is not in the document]
fpagnoux: NumPy is vectorial computation library, meaning it allows to make calculations on large arrays faster than Python native loops.  The benefits are typically visible on _large_ arrays.  For small arrays, the constant costs of NumPy operations usually outweight the benefits of the vectorial approach.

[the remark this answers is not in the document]
fpagnoux: Using a script that is more similar to what happens in the code, this is what I get:

```py
import numpy as np
import time

class Role(object):
  def __init__(self, name):
    self.name = name

flattened_roles = [Role('a'), Role('b'), Role('c')]

role_by_name = {role.name: role for role in flattened_roles}

# Randomly generate a vector of role name
flattened_roles_name = ['a', 'b', 'c']
role_names_array = np.random.choice(flattened_roles_name, 1000000)

# Generate the role object vector using numpy
start = time.time(); roles_array = np.select([role_names_array == role.name for role in flattened_roles], flattened_roles); duration_n = time.time() - start
print(duration_n)

# Generate the role object vector using a Python loop
start = time.time(); roles_array = [role_by_name[role_name] for role_name in role_names_array]; duration_p = time.time() - start
print(duration_p)

print(f'Python is {round(duration_p/duration_n, 2)} slower than NumPy')
```

Result:

```
0.05843806266784668
0.340930700302124
Python is 5.83 slower than NumPy
```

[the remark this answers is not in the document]
fpagnoux: > I don't think we have a dire need to optimize this as it's a one-time setup cost

The thing is a "one-time setup" is not _that_ different from a formula. In a typical simulation where I compute an annual `income_tax`, I'll also run the formula code only once. Yet there was always a strong push from economists like @benjello to avoid looping over any vector. 

I agree that the performance cost here is not a deal breaker, but it seems strange to me that we start breaking the well-established don't-loop-over-vector rule, so far consistently applied in the code.

openfisca_core/simulation_builder.py
```diff
@@ -159,9 +165,37 @@ def build_default_simulation(self, tax_benefit_system, count = 1, **kwargs):
...
+        return entity_instance
+
+    def nb_persons(self, entity_singular, role = None):
+        return self.entities_instances[entity_singular].nb_persons(role = role)
+
+    def join_with_persons(self, group_instance, persons_group_assignment, roles: Iterable[str]):
+        group_sorted_indices = np.unique(persons_group_assignment, return_inverse = True)[1]
+        group_instance.members_entity_id = np.argsort(group_instance.ids)[group_sorted_indices]
+
+        if roles is not None:
```
fpagnoux: If `roles` is now a mandatory argument of the method, is this condition still relevant?

openfisca_core/entities.py
fpagnoux: 👌 
The duplication of a non-trivial code bit seems a little unfortunate to me though. 
Could we wrap-up the `np.logical_or.reduce` in a group-entity method, so that both `has_role`  and `nb_persons` can use it?

openfisca_core/entities.py
fpagnoux: > 👌

openfisca_core/entities.py
fpagnoux: > 👌

openfisca_core/taxbenefitsystems.py
fpagnoux: > 👌

openfisca_core/simulation_builder.py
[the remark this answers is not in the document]
fpagnoux: > 👌 
> I stil wonder if what is currently called an `Entity` instance is not, deep down, an iterable, similar to a NumPy array, but that's a discussion for another day

openfisca_core/simulation_builder.py
Morendil: The *argument* being mandatory is not the same as the *data* being mandatory (on this commit you can still pass None). I've pushed a commit to move in this latter direction but I'm not 100% convinced it's a good idea, all of a sudden to input CSV into OpenFisca you have to know substantially more about the model and encode it into your data "cleanly".

openfisca_core/entities.py
Morendil: Short answer: no. Part of the problem is that `has_role` is a `PersonEntity` method, while `nb_persons` is a `GroupEntity` method. The data that is being reduced belongs to `GroupEntity`.

I can't see a way to solve this without first addressing the problems in `has_role` - specifically the fact that it requires access to all group entities via the Simulation instance.

This is partially a consequence of the signature for formulas - you get a `person` but not the other entities so that `person` is the only thing you can call `has_role` on. And implicitly, `has_role` takes **two** arguments, not one; you don't ask "what role a person has" generically, you're always asking "what role does a person fill in a particular group". The problem is that the Role values are too lightweight (they're [flyweights in the GoF patterns sense](https://en.wikipedia.org/wiki/Flyweight_pattern)) to carry this information.

So the underlying question is, what better approach can we find for filtering by role, and how does this affect the signature of formulas… which is way out of scope for this PR.

openfisca_core/simulation_builder.py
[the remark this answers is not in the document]
Morendil: Hmm. The results I'm getting on my home Mac Mini are going the opposite way from what I reported from last night on my Macbook, whether it's on your script or my original script, so I'm not quite sure what is going on here. Possibly effects of GC on a smaller RAM allocation?

I've pushed the `np.select` version, but note that this also incurs the cost of transforming the Iterable that is passed in to join_with_persons into a Numpy array in the first place, which was not modeled in the above comparisons.

> well-established don't-loop-over-vector rule

It bothers me that we frame performance-impacting choices as "following rules", as opposed to "understanding what's going on *and then* empirically verifying the effect with benchmarks". 

We still don't have a clear understanding of the performance cost we're paying for the "garden of forking paths" effect where we compute every choice of every `np.select` or `np.where`, for instance.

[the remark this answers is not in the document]
fpagnoux: > It bothers me that we frame performance-impacting choices as "following rules", as opposed to "understanding what's going on and then empirically verifying the effect with benchmarks".

Sure, but if the alternative is to use developers time to make benchmarks _each time_ we are considering iterating on a vector, like we've done here, I'm not sure it's worth the pain. 

As for the avoiding "garden of forking paths", this is for sure a very promising idea, but would still need to accommodate the tabular data usecase.

openfisca_core/entities.py
fpagnoux: > As much as I appreciate technical debt reduction and better adhesion to OOP good practices in Core, I'd be worried is this was the main driver to big changes on something as frontal to users as the formulas signature.
> Ease of use for users (including tech-savy economists that are not dev) and expressivity of the formulas (to be assessed by user tests and user feedback) is IMHO the most important aspect of formula syntax.

fpagnoux: > As much as I appreciate technical debt reduction and better adhesion to OOP good practices in Core, I'd be worried is this was the main driver to big changes on something as frontal to users as the formulas signature.
> Ease of use for users (including tech-savy economists that are not dev) and expressivity of the formulas (to be assessed by user tests and user feedback) is IMHO the most important aspect of formula syntax.

PR #839, opened 2019-03-11
Restrict some Pytest options to 'make test', not all Pytest runs

Morendil:
#### Technical changes

- Make running Pytest more useful and less annoying

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ format-style:
 	autopep8 `git ls-files | grep "\.py$$"`
 
 test: clean check-syntax-errors check-style
-	env PYTEST_ADDOPTS="--cov=openfisca_core --exitfirst" pytest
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -25,7 +25,7 @@ format-style:
 	autopep8 `git ls-files | grep "\.py$$"`
 
 test: clean check-syntax-errors check-style
-	pytest
+	env PYTEST_ADDOPTS="--cov=openfisca_core --exitfirst" pytest
```
sandcha: 🙌 for limiting coverage calls!

Not a big fan of `--exitfirst` (stopping at first error) on this tag though as, `make test` is our only simple command to get an overview of the situation (linting, all tests statuses...).

Morendil: …which it was not giving us, previously!

You've convinced me, let's maximize `make test`, and fail fast only in CI.

PR #840, opened 2019-03-12
Improve usability of Enum values

Morendil:
Allow the use of Enum values in comparisons, working around our abuse of the Python module system for TaxBenefitSystem loading.

Tested by updating `test_holders.py` to import HousingOccupancyStatus instead of getting `possible_values`. (Also tested, outside of CI, against the YAML test in [this branch](https://github.com/openfisca/openfisca-france/compare/double-import-error) of France.)

### DIFF
--- a/openfisca_core/indexed_enums.py
+++ b/openfisca_core/indexed_enums.py
@@ -75,7 +75,7 @@ class EnumArray(np.ndarray):
 
     def __eq__(self, other):
         # When comparing to an item of self.possible_values, use the item index to speed up the comparison
-        if other.__class__ is self.possible_values:
+        if str(other.__class__) == str(self.possible_values):
             return self.view(np.ndarray) == other.index  # use view(np.ndarray) so that the result is a classic ndarray, not an EnumArray
         return self.view(np.ndarray) == other
 
--- a/openfisca_core/indexed_enums.py
+++ b/openfisca_core/indexed_enums.py
@@ -50,6 +50,15 @@ class Enum(BaseEnum):
         if array.dtype.kind in {'U', 'S'}:  # String array
             array = np.select([array == item.name for item in cls], [item.index for item in cls]).astype(ENUM_ARRAY_DTYPE)
         elif array.dtype.kind == 'O':  # Enum items arrays
+            # Ensure we are comparing the comparable. The problem this fixes:
+            #  On entering "cls" this method will generally come from variable.possible_values,
+            #  while the array values may come from directly importing a module containing an Enum class.
+            #  However, variables (and hence their possible_values) are loaded by a call to load_module,
+            #  which gives them a different identity from the ones imported in the usual way.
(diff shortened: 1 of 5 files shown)


sandcha:
Might be tested with this code calling `openfisca-tunisia` :
```
from openfisca_tunisia import CountryTaxBenefitSystem as TunisiaTaxBenefitSystem

from openfisca_tunisia.model.prelevements_obligatoires.cotisations_sociales import TypesRegimeSecuriteSociale

nb_individus = 3
simulation_builder = SimulationBuilder()
simulation = simulation_builder.build_default_simulation(tax_benefit_system, count=nb_individus)

period = '2019-01'
simulation.set_input('salaire_de_base', period, [1000, 1000, 1000])

regimes = [TypesRegimeSecuriteSociale.rtns, TypesRegimeSecuriteSociale.raci, TypesRegimeSecuriteSociale.salarie_cnrps]
simulation.set_input('regime_securite_sociale', period, regimes)

ugtt = simulation.calculate('ugtt', period)
assert ugtt[2] == 3, ugtt[2]
```

Morendil:
@sandcha That looks like it would work, and OpenFisca Tunisia needs to be updated to Core 25 first.

openfisca_core/indexed_enums.py
```diff
@@ -50,6 +50,15 @@ def encode(cls, array):
         if array.dtype.kind in {'U', 'S'}:  # String array
             array = np.select([array == item.name for item in cls], [item.index for item in cls]).astype(ENUM_ARRAY_DTYPE)
         elif array.dtype.kind == 'O':  # Enum items arrays
+            # Ensure we are comparing the comparable. The problem this fixes:
+            #  On entering "cls" this method will generally come from variable.possible_values,
+            #  while the array values may come from directly importing a module containing an Enum class.
+            #  However, variables (and hence their possible_values) are loaded by a call to load_module,
+            #  which gives them a different identity from the ones imported in the usual way.
+            #  So, instead of relying on the "cls" passed in, we use only its name to check that
+            #  the values in the array, if non-empty, are of the right type.
```
sandcha: Shorten comment or keep in PR only?
```suggestion
            # "cls" generally comes from variables.possible_values (thus, from load_module)
            # while the array values may come from directly importing a module containing an Enum class. 
```

openfisca_core/simulations.py
```diff
@@ -141,6 +141,8 @@ def calculate(self, variable_name, period, **parameters):
             if array is None:
                 array = holder.default_array()
 
+            array = self._cast_formula_result(array, variable)
```
sandcha: 👍

openfisca_core/indexed_enums.py
Morendil: As you know I'm not a fan of long comments in code, even of comments in general. In this particular case I really feel that the somewhat long explanation is needed to make sense of the code that follows…

PR #843, opened 2019-03-15
Remove simulation_json argument to Simulation constructor

Morendil:
#### Breaking changes

- Remove argument `simulation_json` of `Simulation` constructor, which was deprecated as of Core 25
- Remove keyword arguments from Simulation constructor, which should be called only from SimulationBuilder; introduce a property for `trace`
- Remove `period` attribute of Simulation

#### Migration notes

- As of Core 25, the preferred way of constructing new Simulation instances is via SimulationBuilder, any remaining uses of scenarios should be migrated to that API first.
- Any period attribute of the Simulation was coming from the simulation data (test case or JSON structure), use that instead of the attribute in the Simulation instance.
- Any keyword arguments of Simulation that you were using (or passing to Simulation-constructing methods) can now be accessed directly or as properties, `trace` being the most widely used. Example below:

**Before**

```Python
simulation = SimulationBuilder().build_from_entities(tax_benefit_system, input_data, trace = True)
```

**After**

```Python
simulation = SimulationBuilder().build_from_entities(tax_benefit_system, input_data)
simulation.trace = True
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,32 @@
 # Changelog
 
+# 29.0.0 [#843](https://github.com/openfisca/openfisca-core/pull/843)
+
+#### Breaking changes
+
+- Remove argument `simulation_json` of `Simulation` constructor, which was deprecated as of Core 25
+- Remove keyword arguments from Simulation constructor, which should be called only from SimulationBuilder; introduce a property for `trace`
+- Remove `period` attribute of Simulation
+
+#### Migration notes
+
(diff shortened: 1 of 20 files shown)


Morendil:
@fpagnoux Thanks for approving!

On second thought, this PR for me is not entirely consistent because we're still using new_scenario internally and the point of Core 25 was to move towards use of SimulationBuilder. So I'm adding commits to turn uses of `new_scenario` into uses of SimulationBuilder; to future changes easier I'm also turning those into PyTest fixtures.

Letting you know in case that changes your "Approve" decision.

PR #845, opened 2019-03-19
Always use safe_load to load YAML data

Morendil:
#### Technical changes

- Consistently use the safe approach to YAML loading, fixing [this deprecation warning](https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation) introduced in PyYAML 5.1

### DIFF
--- a/openfisca_core/scripts/migrations/v24_to_25.py
+++ b/openfisca_core/scripts/migrations/v24_to_25.py
@@ -52,7 +52,7 @@ class Migrator(object):
         print('Migrating {}.'.format(path))
 
         with open(path) as yaml_file:
-            tests = yaml.load(yaml_file, Loader=yaml.FullLoader)
+            tests = yaml.safe_load(yaml_file)
         if isinstance(tests, CommentedSeq):
             migrated_tests = [self.convert_test(test) for test in tests]
         else:
--- a/openfisca_core/scripts/migrations/v24_to_25.py
+++ b/openfisca_core/scripts/migrations/v24_to_25.py
@@ -52,7 +52,7 @@ class Migrator(object):
         print('Migrating {}.'.format(path))
 
         with open(path) as yaml_file:
-            tests = yaml.load(yaml_file)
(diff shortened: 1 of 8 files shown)


openfisca_core/scripts/migrations/v24_to_25.py
```diff
@@ -52,7 +52,7 @@ def migrate(self, path):
         print('Migrating {}.'.format(path))
 
         with open(path) as yaml_file:
-            tests = yaml.load(yaml_file)
+            tests = yaml.load(yaml_file, Loader=yaml.FullLoader)
```
fpagnoux: I don't think we really need to use the unsafe loader.

We only use YAML files as human-readable JSONs, and none of the more advanced unsafe features.

We could just replace the `yaml.load` by `yaml.safe_load`, and avoid having to specify a loader, which is a bit heavy. I tried on [this branch](https://github.com/openfisca/openfisca-core/compare/explicit-yaml-loader...explicit-yaml-loader-safe) and the [tests pass](https://circleci.com/gh/openfisca/openfisca-core/4692).

openfisca_core/tools/test_runner.py
```diff
@@ -174,7 +174,7 @@ def _generate_tests_from_directory(tax_benefit_system, path_to_dir, options):
 def _parse_test_file(tax_benefit_system, yaml_path, options):
     with open(yaml_path) as yaml_file:
         try:
-            tests = yaml.load(yaml_file)
+            tests = yaml.safe_load(yaml_file)
```
maukoquiroga:
```suggestion
            tests = yaml.load(yaml_file, Loader = Loader)
```

Then in https://github.com/openfisca/openfisca-core/pull/845/files#diff-68d30ead1f6e6a74cfa83304a419aaa8R44 change to:

```python
from yaml import SafeLoader as Loader
```

Otherwise, test loading time will increase dramatically.

PR #859, opened 2019-03-28
Do not ignore inputs for several months or years

### LINKED ISSUE #841: Take into account input declared for several months
Connected to #460
Connected to #813 

As noticed by @Morendil [here](https://github.com/openfisca/openfisca-core/pull/813#issuecomment-469718590), inputs declared for several months are not properly taken into account.

fpagnoux:
Fixes #841 

When a user tries to set an input for a variable for several months (or several years), but the variable has no `set_input` declared, raise an error instead of ignoring the input.

> I couldn't resist removing some nose dependencies on the way. The real changes brought by this PR are in `holders.py` and `test_holders.py`.

### DIFF
--- a/tests/web_api/test_spec.py
+++ b/tests/web_api/test_spec.py
@@ -4,20 +4,20 @@ import json
 from http.client import OK
 
 import dpath
-from nose.tools import assert_equal, assert_in
+from nose.tools import assert_in
 
 from . import subject
 
 
 def assert_items_equal(x, y):
-    assert_equal(sorted(x), sorted(y))
+    assert sorted(x) == sorted(y)
 
 
 openAPI_response = subject.get('/spec')
 
 
 def test_return_code():
-    assert_equal(openAPI_response.status_code, OK)
+    assert openAPI_response.status_code == OK
 
 
 body = json.loads(openAPI_response.data.decode('utf-8'))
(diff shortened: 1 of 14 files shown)


Morendil:
@fpagnoux Thanks! That unlocks #813 - I'll be merging both today.

tests/web_api/test_spec.py
```diff
@@ -42,7 +42,7 @@ def test_entity_definition():
     assert_in('children', dpath.get(body, 'definitions/Household/properties'))
     assert_in('salary', dpath.get(body, 'definitions/Person/properties'))
     assert_in('rent', dpath.get(body, 'definitions/Household/properties'))
-    assert_equal('number', dpath.get(body, 'definitions/Person/properties/salary/additionalProperties/type'))
+    assert 'number', dpath.get(body == 'definitions/Person/properties/salary/additionalProperties/type')
```
Morendil: Whoa, nope. This transformation does not do what you think it does.

I'm surprised this test even passes with this change… I think I see why: what comes after `assert` here is in fact a tuple and considered truthy by Python.

tests/web_api/test_spec.py
```diff
@@ -51,5 +51,5 @@ def test_situation_definition():
     for situation in situation_input, situation_output:
         assert_in('households', dpath.get(situation, '/properties'))
         assert_in('persons', dpath.get(situation, '/properties'))
-        assert_equal("#/definitions/Household", dpath.get(situation, '/properties/households/additionalProperties/$ref'))
-        assert_equal("#/definitions/Person", dpath.get(situation, '/properties/persons/additionalProperties/$ref'))
+        assert "#/definitions/Household", dpath.get(situation == '/properties/households/additionalProperties/$ref')
```
Morendil: Same here.

Morendil: BTW I think Pytest also gives us `assert 'households' in dpath.get(situation, '/properties')`

tests/web_api/test_spec.py
fpagnoux: Good catch, thanks! Abusive use of regexps...

Issue #706, opened 2018-08-03
Adopt a code formatter


magopian:
Following #702, the question was raised regarding the adoption of an automatic code formatter like [black](https://github.com/ambv/black).

Some discussion already started in https://github.com/openfisca/openfisca-france/pull/1060#issuecomment-410091351

fpagnoux:
There are at least 3 potential candidates:

- [black](https://github.com/ambv/black)
   - ~fascist~ _highly opinionated_: doesn't allow any configuration except for line length. I could possibly see it adopted for Core, but some choices are really rejected by economists, at least in France.
- [yapf](https://github.com/google/yapf)
    - Offers more configuration, ans is able to enforce spaces around `=`
    - Seems a little less well maintained than black. 
    - Doesn't handle the `u` prefix in strings
    - Hanging indent for closing brackets doesn't seem to work right now.
- [autopep8](https://github.com/hhatto/autopep8)
    - Does much less than the other ones
    - Can't enforce spaces around `=`
    - Share conf with `flake8`
    - Weirdly super slow on my machine

fpagnoux:
Looking at the output of the 3 on an already flaked code:
- The main impact of `black` and `yapf` seems to be to enforce a max line length and to handle that consistently. In some cases, the result definitely looks nicer. In others, it's arguable. 
- `autopep8` would just allow to automate the fix of flake8 errors we usually deal with manually. 

I must say I'm so far not fully sure to see the benefits of an advanced code formatter. I can totally be convinced, and adapt if other regular contributors see benefits in it, or if we think that formatters are so standard that not having one makes our codebase look dirty. But if I come back to #702, the two things we wanted to automate were:
- Making sure no `u'` are added to the code.
    - Only `black` can automatically fix that, so we'd have to adopt all its conventions
    - A `grep -e \\bu\' -e \\bu\" **/*.py` could detect them, and there may be a way to autofix in shell to. Agreed, it's much more artisanal than a dedicated formatter. 
- Automatically add future imports to the code
    - Neither `black` nor `yapf` can do that
    - In 4 months, we won't care as we'll stop supporting Python 2

fpagnoux:
So to summarize my position: no strong opinion, will not reject a PR that enforce any code formatting. So if this matters to you (@magopian or anyone) and that a formatter would make you contribute more, go ahead!

maukoquiroga:
Take a look at #727 and let me know of you think that could be a first step towards a stricter solution.

PR #860, opened 2019-03-29
Apply general spring style formatting 🌻

maukoquiroga:
#### Spring is here

- Apply `flake8-bugbear` recommendations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 30.0.2 [#860](https://github.com/openfisca/openfisca-core/pull/860)
+
+- Apply `flake8-bugbear` recommendations and enforce same in continuous integration.
+
 ### 30.0.1 [#855](https://github.com/openfisca/openfisca-core/pull/855)
 
 - Remove Python 2 compatibility code.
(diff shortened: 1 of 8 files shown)


maukoquiroga:
Hi @Morendil !

> I'm 👍 on adding bugbear and following its recommendations

I'll split in two different pull requests.

> Make this a very noisy PR

What could we do to solve that issue? Split in several different pull requests? Or maybe it is just not worth the effort and we should live with it.

> contain changes not mandated by our official linting checks

I did a catch-all to go fast. My objective is to

- Correct no spaces around `=` : this is [not enforceable](https://github.com/openfisca/openfisca-france/pull/1060#issuecomment-410288610) and is [not PEP endorsed](https://github.com/openfisca/openfisca-france/pull/1060#issuecomment-410064151), but it is the style used majoritarily by this repo.

- Correct hang-closing indentation : here again, auto-formatter do not enforce rules 100% of the time.

I'm OK removing all the rest.

See openfisca/openfisca-core#706 and openfisca/openfisca-france#1060

Morendil:
> What could we do to solve that issue?

My best suggestion is to make the changes when you touch these files for some other reason. This spreads the changes over time and keeps them correlated with functional changes.

PR #863, opened 2019-04-10
Display symbolic values of Enums in /trace and print_computation_log

### LINKED ISSUE #822: Return the symbolic values of Enums in /trace and print_com putation_log rather than integers
The trace for strings that belong to a limited list of values appear in the trace as a number instead of a string.

For example, a test that involves computing the family situation of 3 people can look like this :

   statut_marital<2017-01> >> [5 5 5]

While it could be  clearer to show 

   statut_marital<2017-01> >> [pacse  pacse  pacse]

("pacse" being the 5th value of the enum according to the documentation : https://fr.openfisca.org/legislation/statut_marital?q=statut_marital   )

Note that this problem seems to affect both the trace made in python (using the command  "openfisca test  myyaml.yaml -v",  and the tracer tool that appears in another git https://github.com/betagouv/openfisca-tracer

The problem can be reproduced by running a test on the following yaml : 

[cpleyaml.txt](https://github.com/openfisca/openfisca-core/files/2837276/cpleyaml.txt)

Morendil:
#### New features

- Display symbolic values of Enums in /trace and print_computation_log

Fixes #822

### DIFF
--- a/tests/core/test_tracer.py
+++ b/tests/core/test_tracer.py
@@ -47,7 +47,7 @@ def test_trace_enums():
     tracer.record_calculation_end("A", 2017, HousingOccupancyStatus.encode(np.array(['tenant'])))
 
     lines = tracer.computation_log()
-    assert lines[0] == '  A<2017> >> [\'tenant\']'
+    assert lines[0] == "  A<2017> >> ['tenant']"
 
 
 #  Tests on tracing with fancy indexing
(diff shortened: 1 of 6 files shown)


tests/core/test_tracer.py
```diff
@@ -40,6 +41,15 @@ def test_log_format():
     assert lines[1] == '    B<2017> >> 1'
 
 
+def test_trace_enums():
+    tracer = Tracer()
+    tracer.record_calculation_start("A", 2017)
+    tracer.record_calculation_end("A", 2017, HousingOccupancyStatus.encode(np.array(['tenant'])))
+
+    lines = tracer.computation_log()
+    assert lines[0] == '  A<2017> >> [\'tenant\']'
```
fpagnoux:
```suggestion
    assert lines[0] == "  A<2017> >> ['tenant']"
```

> (I know, we're not very consistent about simple vs double quotes)

PR #865, opened 2019-04-16
Improve error messages for entity role issues

fpagnoux:
Connected to #782

### DIFF
--- a/openfisca_core/simulation_builder.py
+++ b/openfisca_core/simulation_builder.py
@@ -285,10 +285,19 @@ class SimulationBuilder(object):
                     persons_to_allocate.discard(person_id)
 
             entity_index = entity_ids.index(instance_id)
-            for person_role, person_id in iter_over_entity_members(entity, roles_json, instance_id):
-                person_index = persons_ids.index(person_id)
-                self.memberships[entity.plural][person_index] = entity_index
-                self.roles[entity.plural][person_index] = person_role
+            role_by_plural = {role.plural or role.key: role for role in entity.roles}
+
+            for role_plural, persons_with_role in roles_json.items():
+                role = role_by_plural[role_plural]
+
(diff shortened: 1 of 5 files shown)


openfisca_core/tools/test_runner.py
```diff
@@ -232,8 +232,6 @@ def _parse_test(tax_benefit_system, test, options, yaml_path):
         builder.set_default_period(period)
         simulation = builder.build_from_dict(current_tax_benefit_system, input)
         simulation.trace = verbose
-    except Exception:
-        raise ValueError("Unexpected error while parsing {}".format(test['file_path']))
```
fpagnoux: > Having a broad exception-catching block before a specific makes the latter impossible to reach

openfisca_core/simulation_builder.py
```diff
@@ -557,25 +557,25 @@ def transform_to_strict_syntax(data):
     return data
 
 
-def iter_over_entity_members(entity_description, scenario_entity):
-    # One by one, yield individu_role, individy_legacy_role, individu_id
-    legacy_role_i = 0
-    for role in entity_description.roles:
+def iter_over_entity_members(entity, roles_json, instance_id):
```
Morendil: It feels very awkward to add instance_id as a parameter to this function, which does not need it other than to compose an error message.

In fact, I suspect this would be simpler, and this function wouldn't be needed at all, if (in `add_group_entity`) we iterated over the keys in the JSON rather than iterating over the roles in the entity.

fpagnoux: Agreed, this function is a remain from the scenarios and could be better integrated in the rest of the class. Trying to find something better.

fpagnoux: Done

PR #866, opened 2019-04-17
Raise an error when keys are duplicated in parameters files

fpagnoux:
#### Breaking changes

- Duplicate keys in YAML parameters files now raise an error
    - Before, only one of the two values declared was taking into account, while the other was silently ignored

#### Rationale
- Duplicate keys are easy to introduce by mistake while editing parameters files. They can cause bugs that are hard to detect immediately, and result in ambiguities in the code base that are hard to fix afterwards.

#### Performance impact

This implementation has no performance impact on the loading of the parameters file.

> For the record, a first approach trying to use `ruamel.yaml` instead of `pyyaml` ended up degrading the loading time (with C loader) by a factor of almost 2, and was abandoned.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+# 33.0.0
+
+### Breaking changes
+
+- Duplicate keys in YAML parameter files now raise an error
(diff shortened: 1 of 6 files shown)


fpagnoux:
Thanks for the quick review!

PR #867, opened 2019-04-19
Use Pytest in the test runner

fpagnoux:
Depends on  #865 

_PR temporarily opened against branch `error-roles` for diff readability_

Use PyTest instead of nose in `openfisca test`


Morendil:
Thanks @fpagnoux this is an idea whose time has finally come :) I'm looking at the impacts on France, a few things to fix (some yield tests for instance), I'll have a PR soon.

PR #868, opened 2019-05-07
Fix test runner

Morendil:
#### Bug fix

- Allow both `*.yaml` and `*.yml` extensions for YAML tests

This was generating a cryptic error message from Pytest when attempting to run YAML tests with the `.yml` extension, which was previously supported by `openfisca-run-test` and `openfisca test`. Thanks @guillett for noticing the problem :)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+# 34.0.1
```
guillett: Il faudrait 3 #.

Issue #862, opened 2019-04-04
Cut calculation time by half with population


sandcha:
LexImpact has to run two calculations on bulk data (50 000 households) in less than 15 seconds.

Definition of done:
Explore in 2 days max
* extracting vectors in calculation (shorten vectors)
* neutralising variables
* or by optimising byte code production

fpagnoux:
Some experimentation on this issue, going more in depth into an [idea](https://github.com/openfisca/openfisca-core/issues/749#issuecomment-433630197) initially introduced by @Morendil.

**TLDR: There seems to be a way to speed up the calculations by 33% to 50% in common usecases**

### Introduction

OpenFisca (OF) runs calculations vectorially. Concretely, when given a large population, OF never loops on the individuals, but run each operation on the global population.
This approach is much faster that looping over the population when the latter is large enough. It however leads to performing a lot of useless calculations, as there is no conditional branching.
For instance, even if a benefit only applies to a small portion of population, it will _still_ be computed for everyone, then zeroed for the ineligible individuals. 

The experiment presented here tries to improve performances by reducing these useless calculations by introducing some kind of conditional branching, while respecting the vectorial paradigm.

The idea is, when facing a condition, to "split and combine" (SC):
- Split the population between the individuals who respect the population, and the one who don't
- Run different calculations on these 2 sub-populations
- Combine the results to get it for the whole population


### Experiment 1

The code is available on [this gist](https://gist.github.com/fpagnoux/e14098a222ec74b78440502465cbd8dc).

- We introduce a benefit that is calculated differently whether individuals are handicapped or not. 
- We calculate it both with OF current's vectorial approach, and with the SC approach, and measure the ratio of execution time between the 2.
- We run this 10 times, and take the average ratio. 
    - A ratio of 2 for instance means that the SC approach is on average twice faster than the current approach
- We study the impact of 3 factors:
    - The **size** `N` of the population
    - The **frequency**  `F` of handicap in the population. 
    - The **complexity** `C` of calculations in each case. This is measured in number of basic vectorial operations in a formula. 3 different versions of the formulas are introduced to that extend.

### Result


C=13 | N=1 | N=1M
-- | -- | --
F=0.1 | 2 | 1.5
F=0.5 | 2 | 1.1


C=6 | N=1 | N=1M
-- | -- | --
F=0.1 | 2 | 1.35
F=0.5 | 2 | 0.9


C=2 | N=1 | N=1M
-- | -- | --
[...]

PR #871, opened 2019-05-07
[WIP] Introduce subpopulations

fpagnoux:
Connected to #862 

This PR follows up on the discussion on #862 and aims at building the base needed to allow using the split and combine approach.

I've tried to experiment using the new operator in a couple formulas on a France branch (`xp-split-combine`). I noticed  a 10% reduction of the time needed to run`test_basics.py`.

-------------------------------

#### Breaking changes

- `population.ids` is now a NumPy array, and not a regular Python list.

#### New features

- Introduce `population.get_subpopbulation(condition)`
  - This allow to extract a supopulation based on a condition
  - Subpopulations can be used the same way than regular populations
- Introduce `population.if_()`, which allows ternary conditions using the split and combine approach.

#### Technical notes

- Subpopulations don't handle their own cache: they write and read in their super_population cache

#### Minor changes

- Add type-checking to the build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,25 @@
 # Changelog
 
+## 35.0.0
+
+#### Breaking changes
+
+- `population.ids` is now a NumPy array, and not a regular Python list.
+
+#### New features
+
+- Introduce `population.get_subpopbulation(condition)`
+  - This allow to extract a supopulation based on a condition
+  - Subpopulations can be used the same way than regular populations
+
+#### Technical notes
+
+- Subpopulations don't handle their own cache: they write and read in their super_population cache
+
+#### Minor changes
+
+- Add type-checking to the build
+
 ### 34.0.1
 
 #### Bug fix
--- a/openfisca_core/data_storage.py
+++ b/openfisca_core/data_storage.py
@@ -2,50 +2,43 @@
 
 import shutil
 import os
+from typing import Dict, Optional
 
 import numpy as np
 
 from openfisca_core import periods
-from openfisca_core.periods import ETERNITY
+from openfisca_core.periods import Period, ETERNITY_PERIOD
 from openfisca_core.indexed_enums import EnumArray
-
+from openfisca_core.commons import PartialArray
 
 class InMemoryStorage(object):
     """
     Low-level class responsible for storing and retrieving calculated vectors in memory
     """
 
-    def __init__(self, is_eternal = False):
-        self._arrays = {}
+    def __init__(self, is_eternal:bool = False):
+        self._arrays: Dict = {}
         self.is_eternal = is_eternal
 
-    def get(self, period):
+    def get(self, period: Optional[Period]) -> Optional[PartialArray]:
         if self.is_eternal:
-            period = periods.period(ETERNITY)
-        period = periods.period(period)
+            period = ETERNITY_PERIOD
 
         values = self._arrays.get(period)
         if values is None:
             return None
-        if isinstance(values, dict):
-            return next(iter(values.values()))
         return values
 
-    def put(self, value, period):
+    def put(self, value: PartialArray, period: Optional[Period]):
         if self.is_eternal:
-            period = periods.period(ETERNITY)
-        period = periods.period(period)
+            period = ETERNITY_PERIOD
 
         self._arrays[period] = value
 
-    def delete(self, period = None):
-        if period is None:
+    def delete(self, period: Period = None):
+        if period is None or self.is_eternal:
             self._arrays = {}
             return
-
-        if self.is_eternal:
-            period = periods.period(ETERNITY)
-        period = periods.period(period)
 
         self._arrays = {
             period_item: value
@@ -77,12 +70,20 @@ class OnDiskStorage(object):
     Low-level class responsible for storing and retrieving calculated vectors on disk
     """
 
-    def __init__(self, storage_dir, is_eternal = False, preserve_storage_dir = False):
-        self._files = {}
-        self._enums = {}
+    def __init__(self, storage_dir: str, is_eternal: bool = False, preserve_storage_dir: bool = False):
+        self._files: Dict = {}
+        self._enums: Dict = {}
         self.is_eternal = is_eternal
         self.preserve_storage_dir = preserve_storage_dir
         self.storage_dir = storage_dir
+        self.values_dir = os.path.join(storage_dir, 'values')
+        self.masks_dir = os.path.join(storage_dir, 'masks')
+        if not os.path.isdir(self.storage_dir):
+            os.mkdir(self.storage_dir)
+        if not os.path.isdir(self.values_dir):
+            os.mkdir(self.values_dir)
+        if not os.path.isdir(self.masks_dir):
+            os.mkdir(self.masks_dir)
 
     def _decode_file(self, file):
         enum = self._enums.get(file)
@@ -91,46 +92,51 @@ class OnDiskStorage(object):
         else:
             return np.load(file)
 
-    def get(self, period):
+    def get(self, period: Optional[Period]) -> Optional[PartialArray]:
         if self.is_eternal:
-            period = periods.period(ETERNITY)
-        period = periods.period(period)
+            period = ETERNITY_PERIOD
 
         values = self._files.get(period)
         if values is None:
             return None
-        if isinstance(values, dict):
-            return self._decode_file(next(iter(values.values())))
-        return self._decode_file(values)
+        values_file, mask_file = values
+        values = self._decode_file(values_file)
+        mask = self._decode_file(mask_file) if mask_file else None
 
-    def put(self, value, period):
+        return PartialArray(values, mask)
+
+    def put(self, partial_array: PartialArray, period: Optional[Period]):
         if self.is_eternal:
-            period = periods.period(ETERNITY)
-        period = periods.period(period)
+            period = ETERNITY_PERIOD
+
+        value = partial_array.value
+        mask = partial_array.mask
 
         filename = str(period)
-        path = os.path.join(self.storage_dir, filename) + '.npy'
-        if isinstance(value, EnumArray):
-            self._enums[path] = value.possible_values
-            value = value.view(np.ndarray)
-        np.save(path, value)
-        self._files[period] = path
+        value_file = os.path.join(self.values_dir, filename) + '.npy'
+        mask_file =  os.path.join(self.masks_dir, filename) + '.npy'
 
-    def delete(self, period = None):
-        if period is None:
+        if isinstance(value, EnumArray):
+            self._enums[value_file] = value.possible_values
+            value = value.view(np.ndarray)
+
+        np.save(value_file, value)
+        if mask is not None:
+            np.save(mask_file, mask)
+            self._files[period] = (value_file, mask_file)
+        else:
+            self._files[period] = (value_file, None)
+
+    def delete(self, period: Optional[Period] = None):
+        if period is None or self.is_eternal:
             self._files = {}
             return
 
-        if self.is_eternal:
-            period = periods.period(ETERNITY)
-        period = periods.period(period)
-
-        if period is not None:
-            self._files = {
-                period_item: value
-                for period_item, value in self._files.items()
-                if not period.contains(period_item)
-                }
+        self._files = {
+            period_item: value
+            for period_item, value in self._files.items()
+            if not period.contains(period_item)
+            }
 
     def get_known_periods(self):
         return self._files.keys()
@@ -138,13 +144,16 @@ class OnDiskStorage(object):
     def restore(self):
         self._files = files = {}
         # Restore self._files from content of storage_dir.
-        for filename in os.listdir(self.storage_dir):
+        for filename in os.listdir(self.values_dir):
             if not filename.endswith('.npy'):
                 continue
-            path = os.path.join(self.storage_dir, filename)
+            values_file = os.path.join(self.values_dir, filename)
+            masks_file = os.path.join(self.masks_dir, filename)
+            if not os.path.isfile(masks_file):
+                masks_file = None
             filename_core = filename.rsplit('.', 1)[0]
             period = periods.period(filename_core)
-            files[period] = path
+            files[period] = (values_file, masks_file)
 
     def __del__(self):
         if self.preserve_storage_dir:
--- a/openfisca_core/data_storage.py
+++ b/openfisca_core/data_storage.py
@@ -11,12 +11,13 @@ from openfisca_core.periods import Period, ETERNITY_PERIOD
 from openfisca_core.indexed_enums import EnumArray
 from openfisca_core.commons import PartialArray
 
+
 class InMemoryStorage(object):
     """
     Low-level class responsible for storing and retrieving calculated vectors in memory
     """
 
-    def __init__(self, is_eternal:bool = False):
+    def __init__(self, is_eternal: bool = False):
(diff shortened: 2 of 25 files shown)


openfisca_core/data_storage.py
```diff
@@ -2,51 +2,45 @@
...
-        self._arrays = {}
+    def __init__(self, is_eternal: bool = False):
+        self._arrays: Dict = {}
         self.is_eternal = is_eternal
 
-    def get(self, period):
+    def get(self, period: Optional[Period]) -> Optional[PartialArray]:
         if self.is_eternal:
-            period = periods.period(ETERNITY)
-        period = periods.period(period)
```
fpagnoux: > I've tightened the type accepted by the data storages methods, as they are internal classes and don't need to be flexible on their input.
> I don't think we need to make that a breaking change, as they are not user-facing.

openfisca_core/data_storage.py
```diff
@@ -2,51 +2,45 @@
...
+            period = ETERNITY_PERIOD
 
         values = self._arrays.get(period)
         if values is None:
             return None
-        if isinstance(values, dict):
```
fpagnoux: > This was a legacy from the extra params, not needed anymore

openfisca_core/data_storage.py
```diff
@@ -64,18 +58,11 @@ def get_memory_usage(self):
                 cell_size = np.nan,
                 )
 
-        nb_arrays = sum([
-            len(array_or_dict) if isinstance(array_or_dict, dict) else 1
```
fpagnoux: > This condition was another legacy from extra params

openfisca_core/holders.py
```diff
@@ -60,39 +64,55 @@ def create_disk_storage(self, directory = None, preserve = False):
...
         """
+        if period is not None:
+            period = make_period(period)
+        cached_array = self.get_cached_array(period)
+        if cached_array is None:
+            return None
+        if cached_array.mask is None:
+            return cached_array.value
+
+        return np.ma.masked_array(  # Masked arrays are a subclass of NumPy arrays that have missing values.
```
fpagnoux: > This guarantees backward compatibility, as the signature of this method hasn't changed.

openfisca_core/populations.py
```diff
@@ -204,10 +233,22 @@ def get_rank(self, entity, criteria, condition = True):
         # Return -1 for the persons who don't respect the condition
         return np.where(condition, result, -1)
 
+    def if_(self, condition: np.ndarray[bool], formula_if_true: Callable):
```
fpagnoux: > This should probably be merged in this PR, but is an example of what kind of helpers could be built on subpopulations.
> There is for sure discussions to have about the interface we want to offer formula writers

Morendil:
@fpagnoux I'm feeling in the dark about the intent here, is this related to the split & combine angle of attack on performance ?

openfisca_core/populations.py
```diff
@@ -204,10 +232,28 @@ def get_rank(self, entity, criteria, condition = True):
         # Return -1 for the persons who don't respect the condition
         return np.where(condition, result, -1)
 
+    def if_(self, condition: np.ndarray[bool], formula_if_true: Callable, formula_if_false: Optional[Callable] = None) -> np.ndarray:
```
fpagnoux: This helper is using the split and combine approach for a ternary condition case.

This doesn't necessarily need to  be merged in this PR, but is an example of what kind of helpers could be built on subpopulations.
There is for sure discussions to have about the interface we want to offer formula writers

fpagnoux:
@Morendil yep, exactly. I've added more contexts in the description.

Morendil:
I've submitted a couple comments that I had pending, because I think it's best if they can be seen; I'm planning to take some time to review this more thoroughly but because of the topic it could take some time and I think it's best if I give my overall feeling right now: I feel very unwilling to take over maintenance if this were merged at this stage. The interaction between subpopulations and caching for instance does not make sense for me in this implementation.

I think this PR is "off-tempo", i.e. it tries to do the right thing at the wrong moment.

The underlying problem has to do with the choice of implementing split-and-combine without first reviewing how computations are scheduled and performed, and information passed between formulas; in particular the fact that we are giving formula authors a sub-optimal programming language in which all variables are de facto global. It seems to me that solving this latter problem should take precedence, but its interaction with split-and-combine is very much not trivial.

More detailed thoughts to come.

fpagnoux:
> The main thing I don't understand is, what seems to make it necessary to cache "subpopulations"?

Do you mean caching the results of a variable calculated for a subpopulation? If so, to avoid recalculating a basic income for a sub-population if we already know the value for all its members.

CHANGELOG.md
```diff
@@ -1,5 +1,25 @@
...
+  - This allow to extract a supopulation based on a condition
+  - Subpopulations can be used the same way than regular populations
+
+#### Technical notes
+
+- Subpopulations don't handle their own cache: they write and read in their super_population cache
+
+#### Minor changes
+
+- Add type-checking to the build
```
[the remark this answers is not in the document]
fpagnoux: 👌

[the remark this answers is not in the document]
fpagnoux: This is minor from the user's point of view, probably not for developers. It's relatively easy to unplug the type checking and just keep the annotation though.

openfisca_core/populations.py
```diff
@@ -552,14 +616,124 @@ def transform(self, result):
...
-    else:
-        if shortcut == 'first_person':
-            return FirstPersonToEntityProjector(population, parent)
-        role = next((role for role in population.entity.flattened_roles if (role.max == 1) and (role.key == shortcut)), None)
-        if role:
-            return UniqueRoleToEntityProjector(population, role, parent)
+class SubPopulation(Population):
+
+    def __init__(self, super_population: Population, condition: np.ndarray, **kw):
+        super().__init__(super_population.entity, **kw)  # type: ignore # (see https://github.com/python/mypy/issues/5887)
```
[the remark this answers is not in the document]
fpagnoux: I don't love that either, but it seems to be the Python way to make a class "multi-inheritance friendly" in Python. If you remove it, constructors will throw an error as they get an unexpected argument meant for an other superclass. 
See [this post](https://stackoverflow.com/questions/34884567/python-multiple-inheritance-passing-arguments-to-constructors-using-super).
An alternative is to not rely on `super`, but explicit which constructor is used. See [this commit](https://github.com/openfisca/openfisca-core/commit/72a6cddf353bf425d9e8f02f668e3e6b73211e35).

I don't have a strong preference. Maybe the alternative makes it easier to understand the intent?

openfisca_core/simulations.py
```diff
@@ -183,11 +202,15 @@ def calculate_add(self, variable_name, period, **parameters):
...
-            self.calculate(variable_name, sub_period, **parameters)
+            self.calculate_(population, variable_name, sub_period, **parameters)
             for sub_period in period.get_subperiods(variable.definition_period)
             )
 
     def calculate_divide(self, variable_name, period, **parameters):
+        population = self.get_variable_population(variable_name)
+        return self.calculate_divide_(population, variable_name, period, **parameters)
+
+    def calculate_divide_(self, population, variable_name, period, **parameters):
```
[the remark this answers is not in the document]
fpagnoux: I agree, and would like to find better naming. 

With this PR's naming: 

- `calculate` is the current `calculate`. It calculates a variable for the whole simulation population. 
- `calculate_` does the same thing, but for a specific population, that may be a subset of the simulation population. It is lower-level, and should not be directly used by a user. Yet it needs to be accessible by some `population`  methods, so it can't be fully private. 

It looks like a [single `_` prefix means "protected" in Python](https://stackoverflow.com/questions/797771/python-protected-attributes), so maybe just moving the `_` as a prefix would be good enough?

fpagnoux:
> I feel very unwilling to take over maintenance if this were merged at this stage. The interaction between subpopulations and caching for instance does not make sense for me in this implementation.

What more precisely do you feel does't make sense? The approach here is to always cache results, so that a variable is never calculated twice for the same individual at the same period. What different approach would you suggest?

> I think this PR is "off-tempo", i.e. it tries to do the right thing at the wrong moment.

> The underlying problem has to do with the choice of implementing split-and-combine without first reviewing how computations are scheduled and performed, and information passed between formulas; in particular the fact that we are giving formula authors a sub-optimal programming language in which all variables are de facto global. It seems to me that solving this latter problem should take precedence, but its interaction with split-and-combine is very much not trivial.

I don't think I agree. Needs for better performance are regularly expressed, both for [small-populations use cases](https://github.com/openfisca/openfisca-core/pull/786#issuecomment-444075042), and [large ones](https://github.com/openfisca/openfisca-core/issues/862). The IPP also has plans to run OpenFisca on unprecedentedly large populations soon.

It seems to me that your post is suggesting postponing an improvement that could, on the short (or middle) run, improve the life of many users, to wait for a much larger refactoring that I'm not sure will materialize soon, if at all. Is a non-sub-optimal language to write formulas emerging as of today? Are there open issues on "how computations are scheduled and performed, and information passed between formulas" that are starting to reach conclusions? 

We can definitely delay one issue if prioritizing another one makes more sense, but literal timing (when are features likely to be delivered?) and the progressive delivery of value to users must be taken into account.

Morendil:
@fpagnoux I realize that my feedback here boils down to "no", and I could be more constructive.

I think there's valuable insight in this PR of the job to be done to adapt caching to partial computations, and the OO design as it is in this PR makes it very hard to understand what's going on. (For instance resorting to multiple inheritance, which is [inherently problematic](https://en.wikipedia.org/wiki/Multiple_inheritance#The_diamond_problem), or multiple classes with the "[refused bequest](https://refactoring.guru/smells/refused-bequest)" anti-pattern.) I'd like to separate the two.

My practical recommendation is to take this PR as a first draft and aim for a do-over where we spend more time discussing, as a team, what decisions we are making. In particular I'd prefer to introduce the split-and-combine approach first *without* changing anything about caching, and then work out precisely what needs to be done to adapt caching.

Morendil:
> The approach here is to always cache results, so that a variable is never calculated twice for the same individual at the same period. What different approach would you suggest?

The above does make sense (although it leaves open the question of whether the cache is valid). It's a crisp expression of the expected result. It helps me make sense of the implementation (although I question the use of recursion in `calculate` - it's not obvious that it is guaranteed to terminate).

What doesn't make sense is to have that formulation of the caching policy, once translated into code, spread across a bunch of classes which do not primarily bear that responsibility, but that of describing the assignment of individuals to entities and roles within these entities. This was already a problem with the original design; the Simulation needed to delegate caching to the population's entities' holders and this PR seems to compound that problem, making future refactoring harder not easier.

Instead, it would probably make more sense to introduce an entirely separate Cache class and give it what it needs to do its job.

This job can be expressed in terms of what `calculate` needs to do now:
- receive a "target" mask (which by default is "all of the population"), a variable and a period
- are there any known values for this variable at this period ?
  - if yes:
    - do these values cover the array mask for which we are calculating ?
      - if yes: return the cached array, limited to that mask
      - if no: set the target mask to the difference between the original target mask and the cached mask
- if we haven't yet returned a value, compute the value for the target mask, if we can
- if we have computed any new values, merge these values at the target mask into the cache.

My intuition is that we can do all that at lower complexity cost than this PR does, with a more straightforward OO design.

One tricky part is that the target mask needs to be propagated down the stack (possibly narrowing with further conditions), and the natural way to do this would be to add a parameter to `calculate`, and that clearly has impact on existing models which we prefer to avoid. To do this by exploiting the protocol of the "population" parameter to `calculate` might be doable, but subclassing isn't our only option to do so.

openfisca_core/populations.py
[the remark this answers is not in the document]
Morendil: I hadn't noticed the multiple inheritance. That's the greater concern. We might be able to avoid it, which would get rid of the problem entirely.

Morendil:
Following up on our conversation IRL, and to start dumping some of my thinking turned into code experiments… Consider the following change to `_run_formula`:

```Python
def _run_formula(self, variable, population, period):
    # …some lines omitted…
    class Calculator:
        def __init__(self, population):
            self.population = population

    calculator = Calculator(population)
    if formula.__code__.co_argcount == 2:
        array = formula(calculator, period)
    else:
        array = formula(calculator, period, parameters_at)

    return array
```

One of the things we need to do if we want to perform computations on partial vectors (or partial populations) is to propagate a "mask" down the stack.

One approach to do this is to **augment the responsibilities of Population**, giving it this mask as an extra attribute. This becomes tricky, as observed, because it seems to give rise to a need to subclass Population into SubPopulation. However, Population is already subclassed into GroupPopulation, and we end up with multiple inheritance, which I see as a smell.

An alternate approach is to [prefer composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance). We compose a Population with a Mask; the Population knows "who is there" and the Mask is responsible for knowing "who we are currently interested in".

In the snippet above, Calculator shows how this composition would work. Instead of passing it just a population, we could pass it a population and a mask. We figure out what formulas need from this object and arrange to delegate its responsibilities to the underlying Population, possibly taking the mask into account.

We get something great out of doing this: we are forced to make explicit the contract that formulas rely on when they receive a Population.

It turns out that all we need to do to keep Core's unit tests passing at 100% is the following:

```Python
class Calculator:
    def __init__(self, population):
        self.population = population

    def __call__(self, variable_name, period = None, options = None, **parameters):
        return self.population(variable_name, period, options)

    def __getattr__(self, attribute):
        if attribute == 'members':
            return self.population.members
        return self.population.__getattr__(attribute)

    def sum(self, array, role = None):
        return self.population.sum(array, role)

    def empty_array(self):
        return self.population.empty_array()

    def filled_array(self, value):
        return self.population.filled_array(value)

    def nb_persons(self, role = None):
        return self.population.nb_persons(role)
```

…which is obviously not enough - this does not include `reduce` or `sum`, for instance, even though it's used in France. So already one benefit of this exercise is that we've identified a gap in test coverage. (In fact `sum` has a unit test in test_entities but `reduce` has no tests at all in Core.)

So we now have to run the next meaningful experiment: run France tests against this modified Core.

When I do this, the next error to turn up is a bit surprising:
```
  File "/app/openfisca_france/model/prelevements_obligatoires/impot_revenu/ir.py", line 68, in formula
    has_birth = individu.get_holder('date_naissance').get_known_periods()
  File "/app/openfisca-core/openfisca_core/simulations.py", line 256, in __getattr__
    return self.population.__getattr__(attribute)
  File "/app/openfisca-core/openfisca_core/populations.py", line 52, in __getattr__
    raise AttributeError("You tried to use the '{}' of '{}' but that is not a known attribute.".format(attribute, self.entity.key))
AttributeError: You tried to use the 'get_holder' of 'individu' but that is not a known attribute.
```

It bothers me a lot that we have formulas using Holder methods directly! It's a low-level abstraction that should not be used or exposed. It also seems to expose a desire for a Cache API available in formulas. On this other hand, this use in France seems to be very restricted, only `age` needs it.

Next, we have `has_role` which is used a lot, for I think usually some wrong reasons (see #874). (And to mention a problem with the Population/GroupPopulation class hierarchy, by the way: `has_role` does not make sense at all for a GroupPopulation.)

So I'll keep doing that until I have France tests running and I know the full contract…

maukoquiroga:
Thanks a lot @fpagnoux for this contribution and @Morendil for the review and comments. And closing this for now as it seems it has not aged well, however please do not hesitate to reopen this issue were this become a priority again (there are a lot of good ideas and contributions to pick from this proposal, and some of them are already integrated into master since) 😃

Issue #853, opened 2019-03-21
Allow composite parameters definition in YAML files


guillett:
Some parameters in the legislation are defined relatively to other parameters in the legislation.
Currently, I do not know how to defined such a _composite_ parameter. Well actually, the current hack is [to create a formula](https://github.com/betagouv/openfisca-bacASable/blob/master/openfisca_bacASable/garantie_jeunes.py#L26-L44) but that's not really pretty.

benjello:
You can use a helper tat depends only on the period or the parameter sub_node tp avoid creating a variable.

Morendil:
@guillett I'm not sure I understand, do you have an example in mind?

guillett:
Thank you both.

@benjello I don't understand your proposal, do you have reference I can have a look at?

@Morendil, I would like [that parameter file (`openfisca_bacASable/parameters/garantie_jeunes.yaml`) in OpenFisca-BacASable](https://github.com/betagouv/openfisca-bacASable/blob/master/openfisca_bacASable/parameters/garantie_jeunes.yaml) to be (or close to):
```yaml
age:
  reference: 'Article L5131-3 du code du travail - https://www.legifrance.gouv.fr/affichCodeArticle.do;jsessionid=DED54A598193DDE1DF59E0AE16BDE87D.tplgfr21s_3?idArticle=LEGIARTI000033023861&cidTexte=LEGITEXT000006072050&dateTexte=20190131'  
  minimum:
    values:
      2009-01-01: 16
  maximum:
    values:
      2009-01-01: 25
montant_maximum:
  values:
    2009-01-01: prestations.minima_sociaux.rsa.montant_de_base_du_rsa * (1 - prestations.minima_sociaux.rsa.forfait_logement.taux_1_personne)
```

benjello:
@guillett :  a solution for the composite parameter (but it is not testable) is a helper like in [here](https://github.com/openfisca/openfisca-france/blob/591b1dce02f4e5b4b1471cf1b574cb5271d8fb57/openfisca_france/model/prestations/prestations_familiales/af.py#L334)

Morendil:
Closing this with #872 which solves the underlying issue that made it awkward to use formulas for that purpose.

PR #872, opened 2019-05-13
Allow formulas to return scalar values and broadcast

Morendil:
#### Technical changes

- Allow formulas to return scalar values; broadcast these values to population-sized arrays

Note that the check for "formulas must return an array" was not unit tested and included, also untested, functionality to check for NaNs in debug mode; this is removed by that PR.

This fixes the underlying issue of #853 - a formula can now compute a value derived from a parameter, for use in a vector computation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 34.2.0
+
+#### New features
+
+- Allow formulas to return scalar values; broadcast these values to population-sized arrays
+
 ## 34.1.0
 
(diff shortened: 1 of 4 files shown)


Morendil:
What is done for neutralized variables?

No, there is no specific space optimization for this particular case. Since it's not used at all (because it seemed like an "ugly hack") I'm reasonably confident we don't need to optimize yet.

benjello:
The idea is that you only stock the value and do the conversion to array when doing the computation.

PR #873, opened 2019-05-13
Fix default group entity

### LINKED ISSUE #849: Non helping ValueError on missing group entity in test case
Hi there!

### Here is what I did:

A `TEST_CASE` missing one group entity (`foyers_fiscaux`).

```python
from openfisca_france import FranceTaxBenefitSystem

tax_benefit_system = FranceTaxBenefitSystem()

TEST_CASE = {
    'individus': {
        'parent1': {
            'age': {'2015-01': 30},
            'salaire_de_base': {'2015': 20000}
        },
        'enfant1': {
            'age': {'2015-01': 12}
        },
        'enfant2': {
            'age': {'2015-01': 18}
        }
    },
    'familles': {
        'famille1': {
            'parents': ['parent1'],
            'enfants': ['enfant1', 'enfant2']
        }
    },
}

simulation_builder = SimulationBuilder()
simulation = simulation_builder.build_from_entities(tax_benefit_system, TEST_CASE)

simulation.calculate('af', '2015-01')
```

A message error saying that I need to define the information Z 
to calculate variable X on the entity Y.

> Z might be that I need to explicitly dispatch persons (individus) on group entity Y.

```python
(...)
~/.local/share/virtualenvs/tut78/lib/python3.7/site-packages/openfisca_core/entities.py in check_array_compatible_with_entity(self, array)
     98         if not self.count == array.size:
     99             raise ValueError("Input {} is not a valid value for the entity {} (size = {} != {} = count)".format(
--> 100                 array, self.key, array.size, self.count))
    101 
    102     def check_role_validity(self, role):

ValueError: Input [False False] is not a valid value for the entity individu (size = 2 != 3 = count)
```

Seen during Core upgrade on openfisca/tutorial#78, `notebooks/getting_started.ipynb`, first simulation building.

The error might come from this `af` > ... > `af_eligibilite_base` calculation step:

```python
residence_dom = famille.demandeur.menage('residence_dom', period)
``` 

## Context

I identify more as a:

Morendil:
#### Bug fixes

- Fix incomplete initialization of group entities provided by default when not supplied in YAML tests

Fixes #849

### DIFF
--- a/openfisca_core/simulation_builder.py
+++ b/openfisca_core/simulation_builder.py
@@ -248,7 +248,8 @@ class SimulationBuilder(object):
         persons_count = len(persons_ids)
         self.entity_ids[entity.plural] = persons_ids
         self.entity_counts[entity.plural] = persons_count
-        self.memberships[entity.plural] = np.arange(0, persons_count - 1, dtype = np.int32)
+        self.memberships[entity.plural] = np.arange(0, persons_count, dtype = np.int32)
+        self.roles[entity.plural] = np.repeat(entity.flattened_roles[0], persons_count)
 
     def add_group_entity(self, persons_plural, persons_ids, entity, instances_json):
         """
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+### 34.2.2
+
+#### Bug fixes
+
+- Fix incomplete initialization of group entities provided by default when not supplied in YAML tests
+
 ### 34.2.1
 
 #### Technical changes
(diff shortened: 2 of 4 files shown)


openfisca_core/simulation_builder.py
```diff
@@ -243,7 +243,8 @@ def add_default_group_entity(self, persons_ids, entity):
         persons_count = len(persons_ids)
         self.entity_ids[entity.plural] = persons_ids
         self.entity_counts[entity.plural] = persons_count
-        self.memberships[entity.plural] = np.arange(0, persons_count - 1, dtype = np.int32)
+        self.memberships[entity.plural] = np.arange(0, persons_count, dtype = np.int32)
```
sandcha: Good catch!

PR #876, opened 2019-05-16
Support role indices in join_with_persons

Morendil:
#### New features

- Support role indices in SimulationBuilder.join_with_persons
  - This broadens the range of input data we can handle, at some risk of misattributing roles

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 34.1.0
+
+#### New features
+
+- Support role indices in SimulationBuilder.join_with_persons
+  - This broadens the range of input data we can handle, at some risk of misattributing roles
+
 ### 34.0.1
 
 #### Bug fix
--- a/openfisca_core/simulation_builder.py
+++ b/openfisca_core/simulation_builder.py
@@ -186,8 +186,10 @@ class SimulationBuilder(object):
         return self.populations[entity_singular].nb_persons(role = role)
 
     def join_with_persons(self, group_population, persons_group_assignment, roles: Iterable[str]):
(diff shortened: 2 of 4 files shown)


Morendil:
@benjello Yes, the unique/argsort maneuver is a Numpy-ism and not obvious to grasp; I had to re-read the test to bring it back to mind:

https://github.com/openfisca/openfisca-core/blob/caee09cb32ce98ea28c314d701141bd7a0f8bee8/tests/core/test_simulation_builder.py#L357-L369

We're saying that the group entity's identifiers
- may be symbols rather than integers
- may be in arbitrary order in the group entity's table

So, we require that the identifiers are given to us in the order of the group entity's table, and then we ensure that we deal properly with the join column in the invididuals' table, when converting into the internal representation (which maps these identifiers to a 0-based range of integers, for indexing into members_roles).

I guess that's a possible comment
```Python
# Maps identifiers to a 0-based integer range, for indexing into members_roles
```

benjello:
@Morendil : I would give it a bit more of context but it is fine for me.

The group entity's identifiers may be symbols rather than integers and may be in arbitrary order in the group entity's table
We require that the identifiers are given to us in the order of the group entity's table, and then we ensure that we deal properly with the join column in the invididuals' table when converting into the internal representation (which maps these identifiers to a 0-based range of integers, for indexing into members_roles).

But you know better than me if these long messages are clarifying or obfuscating the code.
So whatever is good for you (and @sandcha  and @maukoquiroga and @magemax ) is fine for me.

Morendil:
I linked to the PR so the longer explanation is discoverable, and put in the short comment.

Issue #877, opened 2019-05-16
[EPIC] Create a toolbox to check performance


sandcha:
[In progress...]

bonjourmauko:
Are we still looking forward to make this happen? Recently, another initiative has emerged based on inputs and contributions by @benjello @sandcha @benoit-cty and @clallemand. We could maybe merge/supercede #1315

PR #884, opened 2019-05-22
Unify tracing

sandcha:
Connected to #877 

#### Technical changes

- Unifies the disparate data structures previously used for cycle detection, `print_computation_log` and `/trace` endpoints into a single collection contract (`enter_calculation`, `record_xxx`, `exit_calculation`)
- Further work will clarify the contracts for extraction according to use case (log, trace endpoint, performance profiling)


openfisca_core/tracers.py
```diff
@@ -269,27 +269,26 @@ def exit_calculation(self):
         self.stack.pop()
 
 
-class FullTracer:
+class FullTracer(SimpleTracer):
```
Morendil: > Je privilégierais la composition à l'héritage. Cela dit ici ça a l'air de bien marcher -  c'est plus concis en tout cas que l'implémentation que j'imaginais, qui aurait reconstitué la pile à partir du noeud courant de l'arbre en remontant la chaîne des parents.

openfisca_core/tracers.py
```diff
@@ -269,27 +269,26 @@ def exit_calculation(self):
...
 
     def __init__(self):
+        SimpleTracer.__init__(self)
```
Morendil: > Je m'attendrais plutôt à `super().__init()__`, tu as une raison de préférer cette formulation ?

openfisca_core/tracers.py
```diff
@@ -269,27 +269,26 @@ def exit_calculation(self):
...
+        self._current_node = None
 
     @property
     def trees(self):
         return self._trees
 
     def enter_calculation(self, variable: str, period):
-        node = {'name': variable, 'period': period, 'children': [], 'parent': self._next}
-        if self._next is None:
+        SimpleTracer.enter_calculation(self, variable, period)
```
Morendil: > Idem, c'est peut-être mon passé Java mais ici je m'attendrais plutôt à `super().enter_calculation(variable, period)`, est-ce que cette écriture est standard en Python ?

openfisca_core/tracers.py
Morendil: Sorry, mixed up my languages ;)

openfisca_web_api/handlers.py
```diff
@@ -43,23 +40,18 @@ def trace(tax_benefit_system, input_data):
...
 
-    trace = deepcopy(simulation.tracer.trace)
-    for _vector_key, vector_trace in trace.items():
-        value = vector_trace['value'].tolist()
-        if isinstance(vector_trace['value'], EnumArray):
-            value = [item.name for item in vector_trace['value'].decode()]
-        if isinstance(value[0], bytes):
-            value = [str(item) for item in value]
-        vector_trace['value'] = value
+    trace = simulation.tracer.get_flat_trace()
```
fpagnoux: > 👍

openfisca_core/tracers.py
Morendil: @sandcha Up

openfisca_core/tracers.py
Morendil: @sandcha Up

Morendil:
This seems to break "openfisca serve", to be investigated.

Morendil:
> This seems to break "openfisca serve", to be investigated.

Apparently we enter `openfisca_web_api/loader/__init__.py::build_data` but it takes longer than the configured worker timeout to complete. The timeout interrupts during the `build_openAPI_specification`call.

Morendil:
> The timeout interrupts during the `build_openAPI_specification` call.

On further investigation: the root cause seems to be an extreme slowdown when we enter _check_for_cycles after a certain point (specifically, when computing `chomage_imposable`, way down the stack for the Swagger spec example of the `/trace` endpoint on France).

This slowdown does not occur when using the SimpleTracer, but only for FullTracer; my hypothesis is that the FullTracer/SimpleTracer design (with inheritance and what I see as some awkward breaches of encapsulation) might explain this, but I'm investigating still further before reaching conclusions.

Morendil:
> my hypothesis is that the FullTracer/SimpleTracer design (with inheritance and what I see as some awkward breaches of encapsulation) might explain this

Confirmed. This line:
https://github.com/openfisca/openfisca-core/blob/41eedcbf1d6ae3cc60d1a40c65db7c644a856106/openfisca_core/simulations.py#L303

now takes about 10 whole seconds to execute. The variable `chomage_imposable` has a spiral, and is computed 12 times when computing `revenu_assimile_salaire`, so the `/trace` call takes at least 120 seconds to run, well in excess of the timeout threshold.

sandcha:
The performance issue on `openfisca serve` is resolved. 
Reopening for reviews 🙂

sandcha:
@benjello Could you check that there is no regression on aggregated trace? 🙂 
We didn't find tests on it so we added [these checks](https://github.com/openfisca/openfisca-core/pull/884/files#diff-4f941ec7aa147685703d5451446b42a2R40).

benjello:
@sandcha : what do you mean by aggregate trace ?

sandcha:
@benjello Sorry. I meant, when you define `aggregate=True` in `simulation.tracer.print_computation_log(aggregate=True)` for a simulation with bulk data.

benjello:
@sandcha  Never tried that.

tests/core/test_unified_tracing.py
```diff
@@ -0,0 +1,226 @@
...
+
+    assert simulation.tracer.entered
+    assert simulation.tracer.exited
+
+
+@mark.parametrize("tracer", [SimpleTracer(), FullTracer()])
+def test_cycle_error(tracer):
+    simulation = StubSimulation()
+    simulation.tracer = tracer
+    tracer.enter_calculation('toto', 2017)
```
fpagnoux: > Maybe we should replace the 'toto', 'tata' and 'titi' by something less French. Maybe 'x' ('y' and 'z') ?  'variable_x' ?

openfisca_core/tracers.py
```diff
@@ -1,216 +1,198 @@
...
-                  },
-                'global_income<2017-01>': {...}
-              }
 
-        .. py:attribute:: usage_stats
+class FullTracer:
 
-            ``dict`` containing, for each variable computed, the number of times the variable was requested.
+    def __init__(self):
+        self._stacker = SimpleTracer()
```
fpagnoux: Rename `self._stacker` to `self._simple_tracer` ? I'm not sure what a "stacker" is.

sandcha: The idea here is to say that we use the SimpleTracer for its stack.

openfisca_core/simulations.py
```diff
@@ -238,10 +236,7 @@ def _run_formula(self, variable, population, period):
         if formula is None:
             return None
 
-        if self.trace:
-            parameters_at = self.trace_parameters_at_instant
-        else:
-            parameters_at = self.tax_benefit_system.get_parameters_at_instant
+        parameters_at = self.trace_parameters_at_instant
```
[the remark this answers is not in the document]
sandcha: Reset to previous code as it's not easy to move `trace_parameters_at_instant` into `tracers.py` (needs the tax_benefit_system).
It could be improved in an other PR.

tests/core/test_unified_tracing.py
sandcha: Done

openfisca_core/simulations.py
```diff
@@ -238,7 +236,7 @@ def _run_formula(self, variable, population, period):
         if formula is None:
             return None
 
-        if self.trace:
+        if self._trace:
```
fpagnoux: I'd keep `self.trace`

PR #888, opened 2019-06-09
Misc ipp fix

benjello:
#### Technical changes

- Define a trim option for average rate function.
- Convert a dict_values to a list (left-over by the python 3 migration)
- Use a less stric version dependency for `psutils`. Problem arising with `psutils` are mainly due to freeze at install. Use `pip` option `--no-cache-dir` if you face troubles at install.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+### 34.2.5 [#888](https://github.com/openfisca/openfisca-core/pull/888)
+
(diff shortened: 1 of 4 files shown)


benjello:
@claireleroy : what is the psutils version we use at CASD. I want to include it as a lower bound in these PR.

claireleroy:
@benjello I use 5.4.7 at CASD

benjello:
Merci @sandcha !

Issue #887, opened 2019-06-07
Provide an explicit Cache API


Morendil:
Following discussions on #871 (on here and IRL with @fpagnoux) we've identified a few ["sticks"](http://mikadomethod.info/) that should be moved out of the way before implementing some performance optimizations that seem quite promising (split-and-combine, and possibly later taking advantage of static analysis to prune the computation tree ahead of time).

These include #885 and #886 to warm up. The next significant bit is to make the caching API explicit.

maukoquiroga:
A tl;dr of previous efforts was to:

1. Move the cache from the variables to the simulation.
2. Have a way of rewinding or itemising calculations.
3. Separate cache and storage layers.

PR #891, opened 2019-06-21
[WIP/RFC] Provide an explicit Cache API

fpagnoux:
Connected to #887 

## PR Content

This PR is WIP as it only defines some methods contracts rather than implementing them. Only one method is pseudo-implemented, to better explicit the separation of tasks.

Some unit tests are also added, to explicit the behavior of the introduces methods.

The intent is to reach a decision on the big picture before getting drowned into implementation details.

## Approach

The suggestion here is to:

- Dismantle and remove the Holder class
      - Caching calculated variables would be moved to a new class `Cache`
      - The rest of the logic would be moved to other existing classes
- Make `simulation.set_input` the user-facing method to set and input for a variable at a given time

### DIFF
--- a/openfisca_core/cache.py
+++ b/openfisca_core/cache.py
@@ -39,6 +39,6 @@ class Cache:
 
     def get_memory_usage(self) -> Dict[str, Dict]:
         """
-            Get data about the virtual memory usage of the holder.
+            Get data about the virtual memory usage of the cache.
         """
         pass
--- a/openfisca_core/cache.py
+++ b/openfisca_core/cache.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+
+from typing import Optional, List, Dict
+
+from openfisca_core.types import Array
+from openfisca_core.periods import Period
+
+
+class Cache:
+
+
+    def get(self, variable: str, period: Period) -> Optional[Array]:
+        """
+            Get the cached value of ``variable`` for the given period.
+
+            If the value is not known, return ``None``.
+        """
+        pass
+
+    def put(self, variable: str, period: Period, value: Array) -> None:
+        """
+            Store ``value`` in cache for ``variable`` at period ``period``.
+        """
+        pass
+
+    def delete(self, variable: str, period: Optional[Period] = None) -> None:
+        """
+            If ``period`` is ``None``, remove all known values of the variable.
+
+            If ``period`` is not ``None``, only remove all values for any period included in period (e.g. if period is "2017", values for "2017-01", "2017-07", etc. would be removed)
+        """
+        pass
+
+    def get_known_periods(self, variable: str) -> List[Period]:
+        """
+            Get the list of periods the ``variable`` value is known for.
+        """
+        pass
+
+    def get_memory_usage(self) -> Dict[str, Dict]:
+        """
+            Get data about the virtual memory usage of the holder.
+        """
+        pass
--- a/openfisca_core/simulations.py
+++ b/openfisca_core/simulations.py
@@ -426,11 +426,29 @@ class Simulation(object):
 
             If a ``set_input`` property has been set for the variable, this method may accept inputs for periods not matching the ``definition_period`` of the variable. To read more about this, check the `documentation <https://openfisca.org/doc/coding-the-legislation/35_periods.html#automatically-process-variable-inputs-defined-for-periods-not-matching-the-definitionperiod>`_.
         """
-        variable = self.tax_benefit_system.get_variable(variable_name, check_existence = True)
-        period = periods.period(period)
-        if ((variable.end is not None) and (period.start.date > variable.end)):
+        variable = self.get_variable(variable_name)
+        population = self.get_variable_population(variable_name)
+
+        # If the variable is constant over time, we ignore the period parameter and set it for ETERNITY
+        if variable.definition_period == periods.ETERNITY:
+            period = periods.ETERNITY_PERIOD
+        if period is not None:
+            period = periods.period(period)
+
+        if variable.is_inactive(period):
             return
-        self.get_holder(variable_name).set_input(period, value)
+        array = variable.cast_to_array(value)
+
+        if len(array) != population.count:
+            raise ValueError(
+                f'Unable to set value "{value}" for variable "{variable.name}", as its length is {len(array)} while there are {population.count} {population.entity.plural} in the simulation.'
+                )
+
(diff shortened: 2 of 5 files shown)


openfisca_core/cache.py
```diff
@@ -0,0 +1,44 @@
+# -*- coding: utf-8 -*-
+
+from typing import Optional, List, Dict
+
+from openfisca_core.types import Array
+from openfisca_core.periods import Period
+
+
+class Cache:
```
fpagnoux: > This class interface looks very similar to the current `InMemoryStorage`.  Maybe these classes could be merged ?

openfisca_core/cache.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+    def get(self, variable: str, period: Period) -> Optional[Array]:
```
fpagnoux: > I initially thought about names longer like `get_cached_array` or `put_in_cache`, but as this class is just doing cache, it seemed redondant.

openfisca_core/simulations.py
```diff
@@ -426,11 +426,29 @@ def set_input(self, variable_name, period, value):
...
+        variable = self.get_variable(variable_name)
+        population = self.get_variable_population(variable_name)
+
+        # If the variable is constant over time, we ignore the period parameter and set it for ETERNITY
+        if variable.definition_period == periods.ETERNITY:
+            period = periods.ETERNITY_PERIOD
+        if period is not None:
+            period = periods.period(period)
+
+        if variable.is_inactive(period):
```
fpagnoux: > This replaces the check on `variable.end`

openfisca_core/simulations.py
```diff
@@ -426,11 +426,29 @@ def set_input(self, variable_name, period, value):
...
             return
-        self.get_holder(variable_name).set_input(period, value)
+        array = variable.cast_to_array(value)
```
fpagnoux: > This replaces the `_to_array` method of `Holder`

openfisca_core/simulations.py
```diff
@@ -426,11 +426,29 @@ def set_input(self, variable_name, period, value):
...
+
+        if len(array) != population.count:
```
fpagnoux: > I was tempted to abstract this check into a method somewhere else, but the error message is pretty specific to the `set_input` context, and needs access both the `variable` and `population`.

openfisca_core/simulations.py
```diff
@@ -426,11 +426,29 @@ def set_input(self, variable_name, period, value):
...
+            raise ValueError(
+                f'Unable to set value "{value}" for variable "{variable.name}", as its length is {len(array)} while there are {population.count} {population.entity.plural} in the simulation.'
+                )
+
+        if variable.set_input:
+            return variable.set_input(self, variable, period, array)
```
fpagnoux: > The `set_input` hooks need access to both `simulation` and `variable`. It could for sure be done better, but I'd rather do the minimum adaptation and keep further improvement out of scope as there have been recent work on that by @Morendil.

openfisca_core/simulations.py
```diff
@@ -426,11 +426,29 @@ def set_input(self, variable_name, period, value):
...
+
+        variable.check_input_period(period)
```
fpagnoux: > This replaces the period checked that were spread on `holder.set_input` and `holder._set`

tests/core/test_u_cache.py
```diff
@@ -0,0 +1,40 @@
...
+from numpy.testing import assert_equal
+
+from openfisca_core.cache import Cache
+from openfisca_core import periods
+
+period = periods.period(2018)
+
+
+@fixture
+def cache():
```
fpagnoux: > Some methodology question for @Morendil :
> 
> Let's assume that `Cache` uses, in its implementation, a bunch of `InMemoryStorage` instances to store data. This is `Cache` internal business, and is not reflected in the public methods.
> 
> To write a proper **unit** test on Cache, do I need to stub/mock `InMemoryStorage`, or can I just write tests on the Cache methods, as below?
> 
> In other words, can a unit test indirectly run some code that is outside of the class under test, or does it make it an integration test ?
> 
> Thanks a lot in advance!

openfisca_core/cache.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def delete(self, variable: str, period: Optional[Period] = None) -> None:
+        """
+            If ``period`` is ``None``, remove all known values of the variable.
+
+            If ``period`` is not ``None``, only remove all values for any period included in period (e.g. if period is "2017", values for "2017-01", "2017-07", etc. would be removed)
+        """
+        pass
+
+    def get_known_periods(self, variable: str) -> List[Period]:
```
Morendil: I'm not convinced this method needs to be part of the public contract. Who needs it? Under what circumstances?

tests/core/test_u_cache.py
Morendil: I try not to get too hung up on unit/integration test differences, it's a tactical matter rather than one of principle. It's fine if the methods of the object under test rely on some other collaborators, as long as you're primarily testing that object's behaviour, rather than indirectly testing the collaborators.

fpagnoux: Ok, thanks a lot for the explanation 🙂

openfisca_core/cache.py
fpagnoux: Not sure about the exact circumstances, but it is definitely used in several places in IPP code, when they want to figure out _all_ what's in the cache.

tests/core/test_u_cache.py
```diff
@@ -0,0 +1,40 @@
...
+    return Cache()
+
+
+def test_does_not_retrieve(cache):
+    value = cache.get_cached_array('toto', period)
```
sandcha:
```suggestion
    value = cache.get('toto', period)
```

openfisca_core/simulations.py
```diff
@@ -426,11 +426,29 @@ def set_input(self, variable_name, period, value):
...
+        self.cache.put_in_cache(variable.name, period, array)
```
sandcha:
```suggestion
        self.cache.put(variable.name, period, array)
```

maukoquiroga:
Thanks again @fpagnoux for this contribution, which inspired greatly a proposal of my own on #947. I'm closing as this proposal stales a bit, although the issue itself remains open as, at some point, I agree we need to get rid of holders 😃

PR #895, opened 2019-07-02
Generate flame chart for performance analysis

sandcha:
Connected to #877

#### New features

- Introduce `openfisca test myTest.yaml --performance`
  - Generates a flame chart in a web page to view the time taken by every calculation in a simulation

To test it:
1. Run a yaml test:
`openfisca test tests/formulas/irpp.yaml --performance -c openfisca_france`
2. Run a Python web server with:
`python -m http.server 5000`
3. See its result in your browser at `http://localhost:5000`

When your yaml file contains multiple tests, only the last one is displayed in the flame chart.
You can use [openfisca test --name_filter option](https://openfisca.org/doc/openfisca-python-api/openfisca-run-test.html) to choose a specific test case.


alexsegura:
Looks cool 🙂 

What is the format of `performance.json`? Is it proprietary?
I mean, can it be consumed by another program, or only the HTML file with D3 is able to consume it?

Another implementation for the HTML rendering would be to communicate through WebSockets, so that the performance report is updated _live_! But it's more complicated.

Morendil:
@alexsegura The format is quite straightforward, it's a JSON object representing a tree, with dictionary keys `name`, `value` and `children` (the latter recursively containing child nodes), as specified by the [library](https://github.com/spiermar/d3-flame-graph) we're using.

benjello:
Is it possible to format the output in [RunSnakeRun](http://www.vrplumber.com/programming/runsnakerun/)

Morendil:
@benjello No, RunSnakeRun takes cProfile files as input, we don't know how to output that format.

maukoquiroga:
😍

benjello:
@Morendil there is also this other visualisation tool https://jiffyclub.github.io/snakeviz/

maukoquiroga:
@Morendil I'd need some help for the review and rebase, great job though!

fpagnoux:
Rebased

openfisca_core/tracers.py
```diff
@@ -203,3 +218,22 @@ def print_log(self, aggregate = False):
...
+        last_tree = self._full_tracer.trees[-1]
+        simulation_total_time = last_tree['end'] - first_tree['start']
+
+        children = [self.json_tree(tree) for tree in self._full_tracer.trees]
+        return {'name': 'simulation', 'value': simulation_total_time, 'children': children}
+
+    def json_tree(self, tree):
+        calculation_total_time = tree['end'] - tree['start']
+        children = [self.json_tree(child) for child in tree['children']]
+        return {'name': f"{tree['name']}<{tree['period']}>", 'value': calculation_total_time, 'children': children}
```
fpagnoux: Could we rename `value` to `calculation_time` or `time`? I initially thought it was the value of the variable.

openfisca_core/tools/test_runner.py
```diff
@@ -131,38 +135,51 @@ def parse_test(self):
...
+                self.print_computation_log(tracer)
+            if performance:
+                self.create_performance_log(tracer)
+
+    def print_computation_log(self, tracer):
+        print("Computation log:")  # noqa T001
+        tracer.print_computation_log()
+
+    def create_performance_log(self, tracer):
+        f = open("performance.json", "w")
```
fpagnoux: This could be wrapped in a `tracer.performance_log.generate_graph(dir_path)` method. 
That would allow to generate the graph in another context than a text execution (e.g. a calculation on a large population), and I think would better respect the Single Responsibility Principle.

openfisca_core/scripts/tools/index.html
```diff
@@ -0,0 +1,151 @@
+<!DOCTYPE html>
```
fpagnoux: I'd rename the directory containing this file `assets` rather than `tools`

openfisca_core/scripts/tools/index.html
```diff
@@ -0,0 +1,151 @@
...
+      .sort(true)
+      //Example to sort in reverse order
+      //.sort(function(a,b){ return d3.descending(a.name, b.name);})
+      .title("")
+      .onClick(onClick)
+      .differential(false)
+      .selfValue(false);
+
+
+    // Example on how to use custom tooltips using d3-tip.
```
fpagnoux: Dead code

openfisca_core/scripts/tools/index.html
```diff
@@ -0,0 +1,151 @@
...
+    //   .offset([8, 0])
+    //   .attr('class', 'd3-flame-graph-tip')
+    //   .html(function(d) { return "name: " + d.data.name + ", value: " + d.data.value; });
+
+    // flameGraph.tooltip(tip);
+
+    var details = document.getElementById("details");
+    flameGraph.setDetailsElement(details);
+
+    // Example on how to use custom labels
```
fpagnoux: Dead code

openfisca_core/scripts/tools/index.html
fpagnoux: With Chrome, I can't simply open the `index.html` page on my browser, as the request to `performance.json` is blocked by the CORS policy:

I'm able to get the graph only with a small HTTP server like `python -m http.server 8000`

How to open the graph should be documented, at least in the Changelog and PR description. 

image

openfisca_core/tracers.py
Morendil: That's part of the interface contract with d3-flame-graph, so the name is a constraint.

tests/core/tools/test_runner/test_yaml_runner.py
```diff
@@ -143,16 +143,20 @@ def test_performance_option_output():
...
-    if os.path.isfile("./index.html"):
-        os.remove("./index.html")
+    clean_performance_files()
     test_item.runtest()
 
     assert test_item.simulation.trace
     assert os.path.isfile("./index.html")
-    assert os.path.isfile(performance_path)
-    f = open(performance_path, "r")
+    assert os.path.isfile("./performance.json")
```
Morendil: The previous version, with a local variable holding the path instead of two coincidentally equal string constants, was better.

tests/core/tools/test_runner/test_yaml_runner.py
```diff
@@ -143,16 +143,20 @@ def test_performance_option_output():
...
+    f = open("./performance.json", "r")
     assert json.loads(f.read()).get('name')
+    clean_performance_files()
+
+
+def clean_performance_files():
+    if os.path.isfile("./performance.json"):
```
Morendil: Make the JSON and HTML paths arguments to this function, again to avoid relying on coincidental equality of some string constants.

openfisca_core/scripts/tools/index.html
sandcha: 👍 It will be copied to the CHANGELOG. 
> It's already in this PR description.

openfisca_core/simulations.py
```diff
@@ -106,13 +106,15 @@ def calculate(self, variable_name, period):
             period = periods.period(period)
 
         self.tracer.enter_calculation(variable_name, period)
+        self.tracer.record_start(time.time_ns() / (10**9))
```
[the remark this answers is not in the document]
sandcha: I'm in favour of moving `time.time_ns()` to the tracer but:
* we need to test how sub-calculations measure their durations (aka `test_tracers.py::test_calculation_time_with_depth`)
* time precision / os clocks differ (so, a simulation on bulk data might need some other time unit) 🤔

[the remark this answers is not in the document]
sandcha: `record_start` and `enter_calculation` have specific and different missions so I would keep them separated. 

> Besides, this is a first flame chart test based on timestamps but the chart values might take other performance information like the number of calls to the same variable which leads to a call to `enter_calculation` but no`record_start` (Even if, I know, it's bad to involve a possible future to a PR 😬).

openfisca_core/tracers.py
fpagnoux: 👌

openfisca_core/simulations.py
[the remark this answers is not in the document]
fpagnoux: > `record_start` and `enter_calculation` have specific and different missions so I would keep them separated. 

Keeping them separated as 2 different functions in the `tracers` module makes total sense 👍 . 

But do they really have different missions from the simulation's point of view ? The fact that the simulation needs to tell the tracer to first enter a calculation, then to record the time feels to me like the simulation is telling the tracer how to do its job. And this makes me feel like it's an SRP violation.

I also find that the names read very similar at first sight:
- `tracer.record_start `: Record that something has started. Probably a calculation? 
- `tracer.enter_calculation`: Make the tracer enter a calculation. So tell the tracer a calculation has started, how is this different?

> I don't really understand how capturing more information would lead to call `enter_calculation` but not `record_start`, since we decided call `record_start` in all cases, even if we are just reading from the cache. 

This is not a hill I'm willing to die on, but I think we could have a single exposed method explicitly called `record_calculation_start` that would then call private methods to both `record_start_time` and `enter_calculation`.

openfisca_core/tracers.py
```diff
@@ -203,3 +224,31 @@ def print_log(self, aggregate = False):
...
             print(line)  # noqa T001
+
+
+class PerformanceLog:
+
+    def __init__(self, full_tracer):
+        self._full_tracer = full_tracer
+
+    def generate_graph(self, dir_path):
+        f = open(os.path.join(dir_path, 'performance.json'), 'w')
```
fpagnoux: > Using `with` to open and close a file is more Pythonic

openfisca_core/simulations.py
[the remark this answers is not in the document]
fpagnoux: I added a commit containing [the 1st suggested change](https://github.com/openfisca/openfisca-core/pull/895#discussion_r315853136) on this branch, and [another commit](https://github.com/openfisca/openfisca-core/commit/60b3b3b338a21f6486d0f690d1422e7ced4b2c40) addressing my [2nd remark](https://github.com/openfisca/openfisca-core/pull/895#discussion_r315857967)
on a separate branch `trace-performance-suggestion` . 

Feel free to adopt or drop the second one (I think it simplifies the tracer interface and keep responsibilities more separated, but up to you).

openfisca_core/scripts/assets/index.html
```diff
@@ -96,6 +96,16 @@ <h3 class="text-muted">d3-flame-graph</h3>
     var details = document.getElementById("details");
     flameGraph.setDetailsElement(details);
 
+    flameGraph.label(function(d) {
```
sandcha: 😍

sandcha:
Rebasing on `master` branch.

openfisca_core/tracers.py
```diff
@@ -242,13 +242,13 @@ def __init__(self, full_tracer):
         self._full_tracer = full_tracer
 
     def generate_graph(self, dir_path):
-        f = open(os.path.join(dir_path, 'performance.json'), 'w')
-        f.write(json.dumps(self._json()))
-        f.close()
+        with open(os.path.join(dir_path, 'performance.json'), 'w') as f:
+            f.write(json.dumps(self._json()))
+            f.close()
```
fpagnoux: With `with`, the file is automatically closed in the end, so that line is not necessary 🙂

openfisca_core/tracers.py
```diff
@@ -1,9 +1,12 @@
...
 
+import os
+import json
+import time
 import numpy as np
-
-from typing import List, Dict
+from typing import List, Dict, Optional
 from collections import ChainMap
+import importlib.resources as pkg_resources
```
guillett: :exclamation: This removes the 3.6 backward compatibility

New in version 3.7.
From https://docs.python.org/3/library/importlib.html#module-importlib.resources

guillett: Thanks @Morendil  for suggesting `pip install importlib_resources`

openfisca_core/tracers.py
```diff
@@ -99,9 +120,19 @@ def trees(self):
     def computation_log(self):
         return ComputationLog(self)
 
+    @property
+    def performance_log(self):
+        return PerformanceLog(self)
+
+    def _get_time_in_sec(self) -> float:
+        return time.time_ns() / (10**9)
```
guillett: :exclamation: This removes the 3.6 backward compatibility

New in version 3.7.
From https://docs.python.org/3/library/time.html#time.time_ns

openfisca_core/tracers.py
guillett: However, `importlib_resources` still requires this line to be changed into
```python
import importlib_resources as pkg_resources
```

PR #906, opened 2019-09-10
Generate calculation times CSV

fpagnoux:
- Make the tracer generate CSV tables containing details about calculation times
- Make `tracer.get_flat_trace` scale for large populations
- Round calculations times to keep only 4 digits

Example:

`performance_table.csv`:

| name                                  | calculation_time | formula_time | 
|---------------------------------------|------------------|--------------| 
| disposable_income<2017-01>            | 0.001064         | 0.000118     | 
| salary<2017-01>                       | 1.311e-05        | 1.311e-05    | 
| basic_income<2017-01>                 | 0.000428         | 0.000232     | 
| age<2017-01>                          | 0.000196         | 0.0001431    | 
| birth<2017-01>                        | 5.293e-05        | 5.293e-05    | 
| income_tax<2017-01>                   | 7.987e-05        | 6.7e-05      | 
| social_security_contribution<2017-01> | 0.0004251        | 0.0004091    | 

`aggregated_performance_table.csv`:

| name                         | nof_calculations | calculation_time | formula_time | avg_calculation_time | avg_formula_time | 
|------------------------------|------------------|------------------|--------------|----------------------|------------------| 
| age                          | 1                | 0.000196         | 0.0001431    | 0.000196             | 0.0001431        | 
| basic_income                 | 1                | 0.000428         | 0.000232     | 0.000428             | 0.000232         | 
| birth                        | 1                | 5.293e-05        | 5.293e-05    | 5.293e-05            | 5.293e-05        | 
| disposable_income            | 1                | 0.001064         | 0.000118     | 0.001064             | 0.000118         | 
| income_tax                   | 1                | 7.987e-05        | 6.7e-05      | 7.987e-05            | 6.7e-05          | 
| salary                       | 1                | 1.311e-05        | 1.311e-05    | 1.311e-05            | 1.311e-05        | 
| social_security_contribution | 1                | 0.0004251        | 0.0004091    | 0.0004251            | 0.0004091        |


openfisca_core/tools/test_runner.py
```diff
@@ -155,14 +155,14 @@ def runtest(self):
             if verbose:
                 self.print_computation_log(tracer)
             if performance:
-                self.generate_performance_graph(tracer)
+                self.generate_performance_details(tracer)
```
fpagnoux: > Not sure about this method name, would happily take a better suggestion.
> `generate_performance_data` ?

openfisca_core/tracers.py
```diff
@@ -62,6 +68,32 @@ def stack(self):
         return self._stack
 
 
+@dataclass
+class TraceNode:
```
fpagnoux: > Using a dataclass rather than a `dict` makes the code more readable and allow for more semantic typing

openfisca_core/tracers.py
```diff
@@ -62,6 +68,32 @@ def stack(self):
...
+        if round_:
+            return self.round(result)
+        return result
+
+    def formula_time(self):
+        result = self.calculation_time(round_ = False) - sum(child.calculation_time(round_ = False) for child in self.children)
+        return self.round(result)
+
+    @staticmethod
+    def round(time):
```
fpagnoux: > I don't think it makes sense to have calculation times like `8.106231689453125e-06`. This is hard to read and not really relevant. 
> Python profiler keeps only 4 digits, and this PR suggests doing the same

openfisca_core/tracers.py
```diff
@@ -130,53 +162,125 @@ def _get_time_in_sec(self) -> float:
...
-        period = node['period']
+    def key(self, node: TraceNode) -> str:
+        name = node.name
+        period = node.period
         return f"{name}<{period}>"
 
     def get_flat_trace(self):
         trace = {}
-        for tree in self._trees:
-            trace = {**self._get_flat_trace(tree), **trace}
```
fpagnoux: > This was very inefficient for a large trace.

openfisca_core/tracers.py
```diff
@@ -130,53 +162,125 @@ def _get_time_in_sec(self) -> float:
...
+        for node in self.browse_trace():
+            trace.update({  # We don't want cache read to overwrite data about the initial calculation. We therefore use a non-overwriting update.
+                key: node_trace
+                for key, node_trace in self._get_flat_trace(node).items()
+                if key not in trace
+                })
         return trace
 
-    def serialize(self, value):
+    def get_serialized_flat_trace(self):
```
fpagnoux: > Serializing is very expensive for large populations. This PR splits the flattening from the serialization so that we can get one without the other.

openfisca_core/tracers.py
```diff
@@ -130,53 +162,125 @@ def _get_time_in_sec(self) -> float:
...
+            key: {
+                **flat_trace,
+                'value': self.serialize(flat_trace['value'])
+                }
+            for key, flat_trace in self.get_flat_trace().items()
+            }
+
+    def serialize(self, value: np.ndarray) -> List:
         if isinstance(value, EnumArray):
-            value = [item.name for item in value.decode()]
```
fpagnoux: > Using a for makes this function practically unusable for large populations.

openfisca_core/tracers.py
```diff
@@ -62,6 +68,32 @@ def stack(self):
...
+    start: float = 0
+    end: float = 0
+
+    def calculation_time(self, round_ = True):
+        result = self.end - self.start
+        if round_:
+            return self.round(result)
+        return result
+
+    def formula_time(self):
```
fpagnoux: > Also happy to find a different name here. The `calculation_time` is the time needed to calculate a variable for a given period. 
> The `formula_time` is the time spent inside the formula of the given variable for a given period. It does **not** include time spent calculating dependencies.

openfisca_core/tracers.py
```diff
@@ -130,53 +162,125 @@ def _get_time_in_sec(self) -> float:
     def print_computation_log(self, aggregate = False):
         self.computation_log.print_log(aggregate)
 
-    def generate_performance_graph(self, dir_path):
+    def generate_performance_graph(self, dir_path: str) -> None:
         self.performance_log.generate_graph(dir_path)
 
+    def generate_performance_tables(self, dir_path: str) -> None:
```
Morendil: This and the supporting methods (aggregate_calculation_times, _write_csv) should not be in FullTracer, but in the PerformanceLog class. Keep the responsibilities separate.

openfisca_core/tracers.py
```diff
@@ -130,53 +162,125 @@ def _get_time_in_sec(self) -> float:
...
+        with open(path, 'w') as csv_file:
+            writer = csv.DictWriter(csv_file, fieldnames = fieldnames)
+            writer.writeheader()
+            for row in rows:
+                writer.writerow(row)
+
+    def aggregate_calculation_times(self, flat_trace: Dict) -> Dict[str, Dict]:
+
+        def _aggregate_calculations(calculations):
+            nof_calculations = len(calculations)
```
Morendil: I'm not a fan of `nof_` as an abbreviation. `nb_calculations` is more standard, or `count_calculations'.

openfisca_core/tracers.py
```diff
@@ -130,53 +162,125 @@ def _get_time_in_sec(self) -> float:
...
+        self._write_csv(os.path.join(dir_path, 'performance_table.csv'), csv_rows)
+
+        aggregated_csv_rows = [
+            {'name': key, **aggregated_time}
+            for key, aggregated_time in self.aggregate_calculation_times(flat_trace).items()
+            ]
+
+        self._write_csv(os.path.join(dir_path, 'aggregated_performance_table.csv'), aggregated_csv_rows)
+
+    def generate_performance_data(self, dir_path: str) -> None:
```
Morendil: I'm not sure it's a good idea to tie these two tasks together, as the output format is very different and will likely be of interest in different circumstances.

Before this PR, we generate the HTML visualization by invoking the test runner with the` --performance` option. This is unambiguous.

After this PR, the `--performance` option does two things, and possibly we're going to be interested in only one of them. It would seem more useful to me to have either two variants (e.g. `--performance:graph` and `--performance:tables`) or two entirely distinct commands.

openfisca_core/tracers.py
Morendil: It's best to go one step further and move these responsibilities out of the Tracer class, just as we did with ComputationLog and PerformanceLog (and as I argued should be done for the new functionality).

The point of `get_serialized_flat_trace` is to perform the work of the `/trace` endpoint. I'm not sure I understand what role this PR gives to the non-serialized version, `get_flat_trace`. But I don't think they belong in the Tracer class.

openfisca_core/tracers.py
fpagnoux: I used to think the same, until I recently learned from native English speakers that `nb` is [**not** an abbreviation for number in English](https://forum.wordreference.com/threads/abbreviation-of-number-n-n%C2%B0-nr-nbr-no.264328/), but only for _nota bene_.
"No" is the standard abbreviation, but `no_calculations` would be awfully confusing. 

I'm ok with switching to `calculation_count` though

openfisca_core/tools/test_runner.py
fpagnoux: Switched to `generate_performance_data`

openfisca_core/tracers.py
fpagnoux: 👌
Extracted a `FlatTrace` class

openfisca_core/tracers.py
fpagnoux: Done

openfisca_core/tracers.py
```diff
@@ -75,24 +107,24 @@ def record_calculation_start(self, variable: str, period):
         self._record_start_time()
 
     def _enter_calculation(self, variable: str, period):
-        new_node = {'name': variable, 'period': period, 'children': [], 'parent': self._current_node, 'parameters': [], 'value': None}
+        new_node = TraceNode(name = variable, period = period, parent = self._current_node)
         if self._current_node is None:
             self._trees.append(new_node)
         else:
-            self._current_node['children'].append(new_node)
+            self._current_node.children.append(new_node)
```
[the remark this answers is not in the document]
fpagnoux: 👌, just created a `append_child` method.

> I have a theoretical question about this specific case though. To me, dataclasses and dictionnaries in Python are very similar objects, so I'd say that using `self._current_node['children'].append` or `self._current_node.children.append(new_node)` comes with the same level of coupling.
> Do you agree with that, or do you have a different analysis? 
> In other words, was the idea of your suggestion more "Now that we have dataclasses, let's reduce coupling" or "Using a dataclass increases coupling, and we need to mitigate that" ?
> Thanks in advance 🙂

openfisca_core/tracers.py
fpagnoux: 👌, split the 2 options.

PR #908, opened 2019-09-20
Fix parameter-clone

benjello:
#### New features

Fix Parameters.clone() to yield true clones

### DIFF
--- a/openfisca_core/parameters.py
+++ b/openfisca_core/parameters.py
@@ -4,11 +4,12 @@
 """Handle legislative parameters."""
 
 
+import copy
+from typing import Iterable, Optional, Dict, List, Union
+import logging
 import os
 import sys
-import logging
 import traceback
-from typing import Iterable, Optional
 
 import yaml
 import numpy as np
@@ -133,12 +134,12 @@ class Parameter(object):
     """
 
     def __init__(self, name, data, file_path = None):
-        self.name = name
-        self.file_path = file_path
+        self.name: str = name
+        self.file_path: str = file_path
         _validate_parameter(self, data, data_type = dict)
-        self.description = None
-        self.metadata = {}
-        self.documentation = None
+        self.description: str = None
+        self.metadata: Dict = {}
+        self.documentation: str = None
         self.values_history = self  # Only for backward compatibility
 
         # Normal parameter declaration: the values are declared under the 'values' key: parse the description and metadata.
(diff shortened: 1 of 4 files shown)


openfisca_core/parameters.py
```diff
@@ -191,6 +192,18 @@ def __eq__(self, other):
...
 
+    def clone(self):
+        new = empty_clone(self)
+        new_dict = new.__dict__
+
+        for key, value in self.__dict__.items():
+            if key not in ['values_list', 'metadata']:
+                new_dict[key] = value
+
+        new.metadata = copy.deepcopy(self.metadata)
```
fpagnoux: > metadata is a basic Python `dict`, so we can rely on the built-in deepcopy

openfisca_core/parameters.py
```diff
@@ -191,6 +192,18 @@ def __eq__(self, other):
...
+        new.values_list = [parameter_at_instant.clone() for parameter_at_instant in self.values_list]
```
fpagnoux: > Using `clone` here since `parameter_at_instant ` is a custom class

openfisca_core/parameters.py
```diff
@@ -191,6 +192,14 @@ def __eq__(self, other):
     def __call__(self, instant):
         return self.get_at_instant(instant)
 
+    def clone(self):
+        clone = empty_clone(self)
+        clone.__dict__ = self.__dict__.copy()
+
+        clone.metadata = copy.deepcopy(self.metadata)
```
fpagnoux: > metadata is a basic Python `dict`, so we can rely on the built-in `deepcopy`

openfisca_core/parameters.py
```diff
@@ -191,6 +192,14 @@ def __eq__(self, other):
...
+        clone.values_list = [parameter_at_instant.clone() for parameter_at_instant in self.values_list]
```
fpagnoux: > `parameter_at_instant ` is a custom class, so using `clone`

tests/core/parameter_validation/test_parameter_clone.py
```diff
@@ -0,0 +1,69 @@
+# -*- coding: utf-8 -*-
```
fpagnoux: > I'd probably move this test outside of the `parameter_validation` folder.

benjello: Donc

benjello:
@Morendil : could we go this way for now but we wight adopt a solution along the lines you suggested on openfisca-core slack ?

benjello:
cc @maukoquiroga

PR #909, opened 2019-10-24
Introduce variable annualization

fpagnoux:
- Introduce `tax_benefit_system.annualize_variable(variable_name, period)` method
  - Used in reform, this method allows to speed up variable calculations when a month-to-month calculation is not relevant, for instance if the input data are annual.


openfisca_core/variables.py
```diff
@@ -146,7 +147,7 @@ class Variable(object):
            Free multilines text field describing the variable context and usage.
     """
 
-    def __init__(self, baseline_variable = None):
+    def __init__(self, baseline_variable = None, enforce_no_unexpected_attr = False):
```
fpagnoux: > Adding this parameter allows to keep the current behavior of checking for unexpected parameters in variables created by users, but relaxes this constraints by default for easier testing.

tests/core/variables/test_annualize.py
```diff
@@ -0,0 +1,81 @@
...
+from openfisca_core.variables import get_annualized_variable
+
+
+@fixture
+def monthly_variable():
+    class monthly_variable(Variable):
+        value_type = int
+        entity = Person
+        definition_period = MONTH
+        nof_call = 0
```
Morendil: Is this an abbreviation for "number of calls" ? If so, I'd suggest a different name, this leaves too much room for interpretation.

openfisca_core/variables.py
Morendil: I would take this as a hint that the original behavior was not a good idea in the first place. Instead of adding complexity to counter complexity, consider removing the original complexity, i.e. removing the check for extra attributes. What was the value of this check in the first place?

Morendil:
I'm not clear on what the context of this PR is and what benefits are expected, can you add some more information on that? Also perhaps a brief summary of how it's implemented?

tests/core/variables/test_annualize.py
fpagnoux: I used to think the same, until I recently learned from native English speakers that `nb` is [**not** an abbreviation for number in English](https://forum.wordreference.com/threads/abbreviation-of-number-n-n%C2%B0-nr-nbr-no.264328/), but only for _nota bene_.
"No" is the standard abbreviation, but `no_calculations` would be awfully confusing.

I'm ok with switching to `calculation_count` though

openfisca_core/variables.py
fpagnoux: The value is that, as someone who models the legislation, if I mistype `defaut_value = 1` instead of `default_value = 1`, OpenFisca raises an immediate error message, rather than silently ignoring this unexpected input. 

Now, clearly having this check in the constructor was not a good idea

fpagnoux: Change reverted and test adapted

fpagnoux:
> I'm not clear on what the context of this PR is and what benefits are expected, can you add some more information on that? Also perhaps a brief summary of how it's implemented?

I've added some doc about the implementation in the code.

For the context and expected benefits:
- OpenFisca variables often tend to use `MONTH` as a `definition_period` to allow calculations with a fine temporal granularity
- However, in some cases, this fine granularity is not necessary and causes irrelevant heavy calculations. For instance, even if a gross salary is only known globally for the year, a monthly social contribution will be calculated for each month of the year, with the same formula doing the same calculation 12 times. 
- Annualization gives the ability to a reform writer to indicate that calculations for a specific variable and period should be run annually.
- This can result in very significant performance improvements

PR #918, opened 2019-11-28
[WIP] Expression variables

benjello:
Introduce a new way of creating variables through expression_variables and decomposition 

- [ ] Declarative formulas
    - [ ] Expression variables
        - No parameters nor change in period 
        - use https://github.com/pydata/numexpr
        - [x] simplest case : no projections from an entity on another one
        - [x] project person-entity on a group-entity
        - [x] project a group-entity on another group-entity
            - [x] specify the role of the source entity to project upon (after aggregation see below) before applying person-entity projection on a group-entity (see above)
            - [x] sum by default
            - [ ] specify aggregation method
        - [ ] project a group-entity on a person-entity
        - [ ] deal with dated expressions (`expression_2019_01_15`)
        - [ ] extensively test in openfisca-core (@cbenz)
    - [x] USe a decomposition as source of truth
        - write a composition as a YAML file  et generate variables as sum of children
            - [x] introduce variables metadata (`entity: menage`)
        - [x] implement `TaxBenefitSystem.add_variables_from_decomposition_tree`
            - [x] do not overwrote an existing variable (on if `update = True`)
            - [x] error if a leave do not correspond to an existing variable
            - [x] create an expression variable for each node
            - [x] ignore leaves
            - [x] test openfisca-core decompostion using country-template
    - [x] In openfisca-france, use and load `decomposition.yml` (@benjello)
        - [x] See https://github.com/openfisca/openfisca-france/pull/1383
        - [x] From a `decomposition.yml` reproduce `revenu_disponible` and its children
        - [x] Compare results with and without decomposition


sandcha:
Thank you @benjello and @cbenz for this proposition!

I would be happy to have some additional description : 
* what problem dit you experience and are trying to solve here?
* do you have a use case example to describe this problem?
* do you have a code example of the solution you implemented?

> And, if you made some important choices along the road, it could be great to describe them too.

benjello:
> * what problem dit you experience and are trying to solve here?

We wanted to add a faster way of creating new "simple" variables from an expression and use that to 
create variables from a decomposition which is an tree where the nodes are some algebraic sum of their children

> * do you have a use case example to describe this problem?

See decompositions test within this PR and openfisca/openfisca-france#1383 for a real use-case example using decompostion to compute `revenu_disponible` and 

> * do you have a code example of the solution you implemented?

see above 
 
> > And, if you made some important choices along the road, it could be great to describe them too.

The choices are described in the expression variable section above. I acknowledge we lack tests for expression variables alone.

MattiSG:
From what I understand, this PR demonstrates a possible implementation for a way to declare some simple formulas in a declarative way instead of imperative, through “expressions”. Examples can be seen in https://github.com/openfisca/openfisca-france/pull/1383, in particular https://github.com/openfisca/openfisca-france/pull/1383/files#diff-541d996c379ab02b2c6033fb149401918da8618c60682d7fce974051ce4e73deR55-R59:

```python
     expression = "revenus_nets_du_travail + impots_directs + pensions_nettes + ppe + prestations_sociales + revenus_nets_du_capital"
     expression_options = {
         "ppe": {"filter": FoyerFiscal.DECLARANT_PRINCIPAL},
         "prestations_sociales": {"filter": Famille.DEMANDEUR},
         }
```

Such “expressions” surely are appealing compared to writing explicit entity navigation code 😃

However, I believe that this PR should be understood as a way to demonstrate feasibility and to start a discussion, and not as a request to merge the code. The test coverage for such a massive new feature seems insufficient to me, and should be reflected on the Country Template as well as the documentation.
I believe it would be important to assess the potential usage of such expressions on a large codebase such as OpenFisca-France. Only by knowing the proportion of formulas whose declaration would be made simpler can we have good judgement on whether it is worth adding the massive complexity (in terms of learning, documentation and maintenance) of offering a second way to declare formulas on top of the existing one.
I also believe it would be necessary to define precisely what are the limits of these expressions; otherwise, it will be too tempting to expand them until they allow most of the current formulas. However, them being pure text means they will become exponentially complex to validate. I am afraid we would create a text-based DSL on top of the current OpenFisca DSL, with the huge risk of having to handle compilation and tokenisation-level problems, as all text-based systems end up doing.

For me, the way forward with this proposal is rather steep, and would require:

1. Putting together a formal proposal in the form of an RFC with a scope of expressions, examples of formulas written as expressions, and potential impact on a large codebase.
2. Identifying users who can _not_ currently contribute and who _would_ contribute if such a feature was introduced.
3. Leading a discussion on the right level of abstraction for such declaration extensions: do we consider that there would now be two independent DSLs, one based on expressions and the other in the “OpenFisca Classic” format, and there could be more in the future? Or is this just a “shorthand syntax” on top of the “OpenFisca Classic” format, meaning the former could always be expanded into the latter?

Considering this proposal has not seen work in 18 months, I:

1. Mark it as draft.
2. Close it in order to clean up the list of open PRs to make it easier for maintainers to focus on active issues.
3. Encourage the original authors to create an issue as an RFC as defined above, referencing this PR and https://github.com/openfisca/openfisca-france/pull/1383 to demonstrate feasibility.

If anyone disagrees with this assessment and considers this code is ready to be merged and would provide value as such, please feel free to reopen 🙂

PR #923, opened 2020-01-02
Fix enum scalar array bug

### LINKED ISSUE #922: Error when trying to encode scalar Enum arrays
The following code fails for scalar Enum arrays (without dimension):

https://github.com/openfisca/openfisca-core/blob/871bf2007545236bc814b49dea27559bef87c584/openfisca_core/indexed_enums.py#L60

A real-life example of this bug:

https://github.com/openfisca/openfisca-france/blob/7918eec5be4c2cdeb65da9fb023cfbfcce537156/openfisca_france/model/prestations/logement_social.py#L71-L81

You can reproduce the issue with the following code:

```py
from numpy import array, atleast_1d

class Greetings(Enum):
    hola = "Hola"
    chao = "Chao"

greeting = array(Greetings.hola)
greeting.size # 1
len(greeting) # TypeError: len() of unsized object
atleast_1d(greeting).size # 1
len(atleast_1d(greeting)) # 1
```

What's the expected behaviour here?

1. We're expecting an array with at least one dimension.
2. We should make sure we cast or take into account scalar arrays.

cc @Morendil @fpagnoux

maukoquiroga:
Fixes #922 

#### Bug fix

- Fix a `TypeError` in `Enum.encode` when trying to encode a scalar Numpy array.
  - Use `numpy.take` to retrieve the first element of any array (including not indexed).


openfisca_core/indexed_enums.py
```diff
@@ -1,98 +1,142 @@
...
+            # variable.possible_values, while the array values may come from directly
+            # importing a module containing an Enum class. However, variables (and
+            # hence their possible_values) are loaded by a call to load_module, which
+            # gives them a different identity from the ones imported in the usual way.
+            # So, instead of relying on the "cls" passed in, we use only its name to
+            # check that the values in the array, if non-empty, are of the right type.
+            array_size: int = size(array)
+            array_name: str = take(array, 0).__class__.__name__
+
+            if array_size > 0 and array_name is cls.__name__:
```
maukoquiroga: @Morendil @benjello Here :)

maukoquiroga:
This PR adds tests, code itself doesn't change, except for :

```py
len(x)
```

to

```py
size(x)
```

Question is : handling scalar vectors is the expected behaviour ?

maukoquiroga:
Hey thanks @Morendil and @benjello!

Now PR is scoped to document expected behaviour of `Enum` and `EnumArray`.

tests/core/test_indexed_enums.py
```diff
@@ -0,0 +1,142 @@
...
+
+def test_enum_encode_when_array_is_array_of_values(my_enum):
+    values = array([my_enum.bar.value])
+
+    result = my_enum.encode(values)
+
+    assert result[0] == 1
+
+
+def test_enum_encode_when_array_is_scalar_array_of_value(my_enum):
```
maukoquiroga: This "just works", should we just assume it is an implementation side-effect and delete the test?

tests/core/test_indexed_enums.py
```diff
@@ -0,0 +1,142 @@
...
+
+def test_enum_encode_when_array_is_array_of_indices(my_enum):
+    values = array([1])
+
+    result = my_enum.encode(values)
+
+    assert result[0] == 1
+
+
+def test_enum_encode_when_array_is_scalar_array_of_indices(my_enum):
```
maukoquiroga: Similarly here.

tests/core/test_indexed_enums.py
maukoquiroga: For me it should throw a `TypeError`

tests/core/test_indexed_enums.py
maukoquiroga: For me it should throw a `TypeError`

tests/core/test_indexed_enums.py
```diff
@@ -0,0 +1,142 @@
...
+    values = array(1)
+
+    result = my_enum.encode(values)
+
+    assert result == 1
+
+
+def test_enum_encode_when_array_is_another_data_structure(my_enum):
```
maukoquiroga: A contrario, this one should work IMHO, by _duck typing_, but it doesn't.

maukoquiroga:
Bump @Morendil @benjello

Issue #924, opened 2020-01-03
Compatibility with Numpy 1.18


maukoquiroga:
[Numpy bumped to 1.18](https://numpy.org/devdocs/release/1.18.0-notes.html), [are we prepared?](https://github.com/openfisca/openfisca-france/pull/1395#pullrequestreview-337568319)

I did analysis by going through each change announced in the Numpy's 1.18 changelog, checking against OpenFisca Core's code base, and here's what I found.

This is a "theme" or "meta" issue to deal with each part of the code base that needs some work in order to ensure compatibility with Numpy 1.18.

- [x] https://github.com/openfisca/openfisca-core/blob/d47f65ad21c2de683a324378fc80a1177f0efa04/openfisca_core/simulation_builder.py#L515

- [x] https://github.com/openfisca/openfisca-core/blob/d47f65ad21c2de683a324378fc80a1177f0efa04/openfisca_core/simulation_builder.py#L520

- [x] https://github.com/openfisca/openfisca-core/blob/92ce9396e29ae5d9bac5ea604cfce88517c6b35c/openfisca_core/scripts/measure_numpy_condition_notations.py#L56

- [x] https://github.com/openfisca/openfisca-core/blob/273e17b65ccbe6941aa5576140d9c880990483bc/openfisca_core/formula_helpers.py#L31

- [x] https://github.com/openfisca/openfisca-core/blob/273e17b65ccbe6941aa5576140d9c880990483bc/openfisca_core/formula_helpers.py#L47

- [x] https://github.com/openfisca/openfisca-core/blob/5a1fd43eb16e55710d9dd84c90ff7408c765e3e6/openfisca_core/indexed_enums.py#L51

- [x] https://github.com/openfisca/openfisca-core/blob/5a1fd43eb16e55710d9dd84c90ff7408c765e3e6/openfisca_core/indexed_enums.py#L62

- [x] https://github.com/openfisca/openfisca-core/blob/5a1fd43eb16e55710d9dd84c90ff7408c765e3e6/openfisca_core/indexed_enums.py#L116

- [x] https://github.com/openfisca/openfisca-core/blob/5a1fd43eb16e55710d9dd84c90ff7408c765e3e6/openfisca_core/indexed_enums.py#L128

- [x] https://github.com/openfisca/openfisca-core/blob/c64f9083c1fae9c572282370c3d1f4164c312e1a/openfisca_core/tools/simulation_dumper.py#L80

[...]

PR #925, opened 2020-01-03
Document that we expect axis count to be an int

maukoquiroga:
Part of #924 

#### Documentation

- Annotates the expected type of the axis count in `SimulationBuilder`.


maukoquiroga:
Hey @benjello, I'm moving discussion on https://github.com/openfisca/openfisca-core/issues/924#issuecomment-570505486 here.

I think we should just :

- Make sure code is properly tested (we're using those methods as we should)
- Make sure code is properly documented (type annotations, docstrings)

In this particular case, type annotating the variable is enough.

PR #926, opened 2020-01-03
Downgrade Numpy to 1.17

maukoquiroga:
#### Technical change

- Downgrade numpy version's upper bound to 1.17
- Details:
  - Numpy 1.18 deprecates the use of several of its methods.
  - Changes in `numpy.select` have impacted other packages depending on OpenFisca Core.


maukoquiroga:
Thanks @Morendil !!!

PR #934, opened 2020-01-03
Handle values defined on different period structures

### LINKED ISSUE #564: Handle individuals using different period structures for the same variable.
Hi there!

### Here is what I did:

I sent the following payload to the preview API:
```json
{
  "familles": {
    "famille_1": {
      "enfants": [
      ],
      "parents": [
        "Bill",
        "Bob"
      ]
    }
  },
  "foyers_fiscaux": {
    "foyer_fiscal_1": {
      "declarants": [
        "Bill",
        "Bob"
      ],
      "personnes_a_charge": [
      ]
    }
  },
  "individus": {
    "Bill": {
      "salaire_de_base": {
        "2017": 20000
      },
      "tns_auto_entrepreneur_chiffre_affaires": {
        "2016": 0,
        "2016-08": 0,
        "2016-09": 0,
        "2016-10": 0,
        "2016-11": 0,
        "2016-12": 0,
        "2017-01": 0,
        "2017-02": 0,
        "2017-03": 0,
        "2017-04": 0,
        "2017-05": 0,
        "2017-06": 0,
        "2017-07": 35000,
        "2017-08": 35000,
        "2015-01": 2916.6666666666665,
        "2015-02": 2916.6666666666665,
        "2015-03": 2916.6666666666665,
        "2015-04": 2916.6666666666665,
        "2015-05": 2916.6666666666665,
        "2015-06": 2916.6666666666665,
        "2015-07": 2916.6666666666665,
        "2015-08": 2916.6666666666665,
        "2015-09": 2916.6666666666665,
        "2015-10": 2916.6666666666665,
        "2015-11": 2916.6666666666665,
        "2015-12": 2916.6666666666665
      }
    },
    "Bob": {
      "tns_auto_entrepreneur_chiffre_affaires": {
        "2015": 6000,
        "2016": 6000,
        "2016-08": 0,
        "2016-09": 0,
        "2016-10": 0,
        "2016-11": 0,
        "2016-12": 0,
        "2017-01": 0,
        "2017-02": 0,
        "2017-03": 0,
        "2017-04": 0,
        "2017-05": 0,
        "2017-06": 0,
        "2017-07": 0,
        "2017-08": 0
      }
    }
  },
  "menages": {
    "menage_1": {
      "conjoint": [
        "Bob"
      ],
      "enfants": [
      ],
      "impots_directs": {
        "2017": null
      },
      "personne_de_reference": [
        "Bill"
      ],
      "revenu_disponible": {
        "2017": null
      }
    }
  }
}
```

To return without error.

I get:
```json
{
  "error": "Internal server error: Inconsistent input: variable tns_auto_entrepreneur_chiffre_affaires has already been set for all months contained in period 2015, and value [    0.  6000.] provided for 2015 doesn't match the total ([ 34999.99609375      0.        ]). This error may also be thrown if you try to call set_input twice for the same variable and period."
}
```

tns_auto_entrepreneur_chiffre_affaires [legislation.openfisca.fr](https://legislation.openfisca.fr/tns_auto_entrepreneur_chiffre_affaires)/[code](
https://github.com/openfisca/openfisca-france/blob/18.9.0/openfisca_france/model/revenus/activite/non_salarie.py#L1464-L1469) relies on `set_input_divide_by_period`.

For Bob a normalization is required as data is provided annually on 2015.
For Bill a normalisation is not required as data is provided monthly on 2015.

That discrepancy seems to be the origin of the problem.

## Context

I identify more as a:

guillett:
Based on #933 
Fixes #564

The new function `solve_input_data` is hacky but I would like to start a conversation to have that feature in Core soon.

### DIFF
--- a/openfisca_core/holders.py
+++ b/openfisca_core/holders.py
@@ -149,7 +149,7 @@ class Holder(object):
             periodMap = {k: periods.period(k) for k in keys}
             periodValues = periodMap.values()
             starts = [p.start for p in periodValues]
-            ends = [p.offset(p.size_in_months - 1).start for p in periodValues]
+            ends = [p.offset(p.size_in_months - 1, unit=MONTH).start for p in periodValues]
             start = min(starts)
             end = max(ends)
 
@@ -161,20 +161,24 @@ class Holder(object):
             full_size = (self.population.count, size)
             presence = np.full(full_size, False)
             values = self.variable.default_array(full_size)
+            dim1, dim2 = np.indices(full_size)
 
             # Set single period values
             for k in keys:
                 p = periodMap[k]
                 if p.size_in_months == 1:
                     column = month_index(p.start)
-                    presence[:, column], values[:, column] = buffer[k]
+                    p_presence, p_values = buffer[k]
+                    tile_count = self.population.count // p_values.size
+                    presence[:, column] = np.tile(p_presence, tile_count)
+                    values[:, column] = np.tile(p_values, tile_count)
 
             # Set values for multiple period input
             for k in keys:
                 p = periodMap[k]
                 if p.size_in_months != 1:
                     # Determine period indexes
-                    idx = np.array(range(p.size_in_months)) + month_index(p.start)
+                    idx = slice(month_index(p.start), p.size_in_months)
 
                     counts = p.size_in_months - presence[:, idx].sum(axis=1)
                     current_sum = values[:, idx].sum(axis=1)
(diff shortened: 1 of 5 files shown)


Morendil:
This is not the way to go IMO, it makes Holder more complex, when what we want is for them to become simpler… and even simpler, until eventually they go away. Getting rid of Holders is high priority in improving the architecture of Core.

Ask yourself - why does this need to be in Holder, where it will be in common use among fine-grained YAML-test-style data *and* tabular data, when only the former needs it ?

Do the Cache API first, migrate to it, get rid of Holders, and then it will be more obvious how to implement this where it should be implemented - in SimulationBuilder.

guillett:
@Morendil many thanks for your feedback. There are multiple issues to be tacked with such a feature/bug fix
- have a working version of it
- put it in the right place.

I worked in the first one we'll have to work on the second one as a community.
@benjello I am eager to get your insight/feedback on that PR too. :)

benjello:
@guillett : I do understand the need for a feature handling different period range for the same input_variable and I see how it can be useful.
I am not versed in this part openfisca-core code enough to make any insightful comment at this point, sorry about that.

I would suggest not to use camel case and use more long names for local variables so it can be more readable by a unexperienced coder as myself and I would definitively give it a second try since this question is important.

PR #943, opened 2020-02-02
Use Python's ABC for abstracts

maukoquiroga:
#### Deprecations

- Deprecate `Dummy`.
  - The functionality is now directly provided by `empty_clone`.

#### Technical changes

- Refactor abstract scales to use Python's `abc` lib.
  - This allows for consistent inheritance, which wasn't the case before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## 34.7.0 [#943](https://github.com/openfisca/openfisca-core/pull/943)
+
+#### Deprecations
+
+- Deprecate `Dummy`.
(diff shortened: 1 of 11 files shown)


maukoquiroga:
@benjello could you test and let me know? I've tried to respect all contracts as I understood them, although as in the previous implementation abstractness wasn't really enforced, it may be that you hacked your way away with those two classes. Nonetheless, even in that case, I tried to preserve everything as it was before, even those classes, in an attempt to remove them in a major.

maukoquiroga:
(to support my claim, see that tests are not changed, only completed).

PR #947, opened 2020-02-08
[WIP/RFC] Provide an explicit caching  API

maukoquiroga:
Connected to #887 
Connected to #891 

#### What's inside?

A possible implementation of #891, trying to keep current contracts as they're now.

This is a proposition made to separate concerns between caching and storing logic, so as to move all caching related tasks from `Holder` to `Cache`, and all storing logic from everywhere to `MemoryStorage` and `DiskStorage`.

Assuming those low-level classes were meant to be private, the new `Cache` class respects the previous classes' contracts, and warns the user of deprecated methods to be definitely removed in the future.

#### Technical changes

- [x] Create `MemoryStorage` and move storage logic here.
- [x] Create `DiskStorage` and move storage logic here.
- [x] Extract eternalness checks somewhere else.
- [x] Merge current caching implementations into a single `Cache` class.
- [x] Add warning messages for deprecated methods.
- [x] Keep current contracts by dispatching to the new ones.

#### What's next?

- [ ] Somehow get rid of the eternal / non-eternal variable check.
- [ ] Move caching logic out of `Holder` to `Cache`, notably the 'lookup'
- [ ] Preserve contracts of `calculate`, `set_input*`, 
- [ ] One simulation, one cache
- [ ] Cache.lookup(function(variable, period)))
- [ ] Git rid of `Holder`.

---

Thanks for your feedback!


.circleci/config.yml
```diff
@@ -24,25 +24,25 @@ jobs:
             # pip install --editable git+https://github.com/openfisca/country-template.git@BRANCH_NAME#egg=OpenFisca-Country-Template  # use a specific branch of OpenFisca-Country-Template
             # pip install --editable git+https://github.com/openfisca/extension-template.git@BRANCH_NAME#egg=OpenFisca-Extension-Template  # use a specific branch of OpenFisca-Extension-Template
 
-      - save_cache:
-          key: v1-py3-{{ .Branch }}-{{ checksum "setup.py" }}
-          paths:
-            - /tmp/venv/openfisca_core
+      - run:
```
Morendil: I would advise moving these CircleCI changes to a separate PR and explaining the intent in more detail there.

openfisca_core/data_storage.py
```diff
@@ -1,136 +1,191 @@
...
 
+from openfisca_core import indexed_enums
 from openfisca_core import periods
-from openfisca_core.periods import ETERNITY
-from openfisca_core.indexed_enums import EnumArray
 
+from typing_extensions import Protocol
+
+
+class StorageLike(Protocol):
```
Morendil: Mixing abstract base classes and structural subtyping feels weird here.

openfisca_core/data_storage.py
```diff
@@ -1,136 +1,191 @@
...
+            period = periods.period(periods.ETERNITY)
+
         period = periods.period(period)
 
-        self._arrays = {
-            period_item: value
-            for period_item, value in self._arrays.items()
-            if not period.contains(period_item)
-            }
+        if period is not None:
```
Morendil: This null check is redundant given the constructor call above. (Which itself is redundant given the guard clause.)

openfisca_core/data_storage.py
```diff
@@ -1,136 +1,191 @@
...
+            self._arrays = self._pop(period, list(self._arrays.items()))
 
-    def get_known_periods(self):
-        return self._arrays.keys()
+    def get_known_periods(self) -> List[periods.Period]:
+        return list(self._arrays.keys())
```
Morendil: Be careful here: this is a change of behaviour. `dict.keys()` returns a (dynamic) view, the `list()` calls freezes it. (The change is not necessary, as this was already being frozen by the calling code in Holder.)

I don't think the change is consequential, but it makes the title of the PR a lie, as we're adding tests *and* changing behaviour - *and* changing the CI config *and* adding type annotations *and* changing the style of imports… PRs that do many things at once are harder to review and more likely to contain unnoticed errors.

tests/core/data_storage/test_on_disk_storage.py
```diff
@@ -0,0 +1,174 @@
...
+
+from openfisca_core import data_storage
+from openfisca_core import periods
+
+import pytest
+
+
+@pytest.fixture
+def storage():
+    return functools.partial(data_storage.OnDiskStorage, storage_dir = "/tmp")
```
Morendil: That style of import isn't an advantage when dealing with classes. Using
```
from openfisca_core.data_storage import OnDiskStorage
```
is just fine here.

The meta-rule for imports is, I should be able to look at the name of a method or class and if I'm not familiar with it, glance back to the imports section of the source file and immediately see where it is imported from. This saves a lot of time when debugging, compared to wildcard imports.

Systematically prefixing can be advantageous in particular cases like Numpy methods, but I wouldn't make it a general rule. It feels burdensome.

tests/core/data_storage/test_in_memory_storage.py
```diff
@@ -0,0 +1,113 @@
...
+    result = storage.get(period)
+
+    assert result == value
+
+
+def test_put_when_is_eternal(eternal_storage, value):
+    storage = eternal_storage()
+    storage.put(value, "foo")
+
+    result = storage.get("bar")
```
Morendil: This doesn't really make sense. The argument of `get` is a period, not an arbitrary string. This test commits the implementation to happily support arbitrary strings, which isn't the intent; the intent is to return the same value for all periods.

tests/core/data_storage/test_in_memory_storage.py
```diff
@@ -0,0 +1,113 @@
...
+
+def test_delete_when_period_is_not_specified(storage, period, value):
+    storage = storage()
+    storage.put(value, period)
+    storage.put(value, period.last_year)
+    storage.delete()
+
+    result = storage.get(period)
+
+    assert not result
```
Morendil: When juxtaposed with the previous test, this seems to suggest that we still have a value for `period.last_year`.

tests/core/data_storage/test_in_memory_storage.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def test_delete_when_is_eternal(eternal_storage, value):
+    storage = eternal_storage()
+    storage.put(value, "qwerty")
+    storage.put(value, "azerty")
+    storage.delete("asdf1234")
```
Morendil: See above; better to use periods.

tests/core/data_storage/test_on_disk_storage.py
```diff
@@ -0,0 +1,174 @@
...
+def test_get(storage, period, file, mocker):
+    storage = storage()
+    files = {period: file}
+    mocker.patch.dict(storage._files, files)
+    mocker.patch.object(storage, "_decode_file", autospec = True)
+    storage.get(period)
+
+    result = files.get(period)
+
+    storage._decode_file.assert_called_once_with(result)
```
Morendil: It's really difficult for me to make any sense of this test. I don't have any intuitive sense of what is being guaranteed. I have no idea what `autospec` does or why it's relevant here. The test demands detailed knowledge of the implementation and thus fails to provide much value as documentation.

.circleci/config.yml
maukoquiroga: It just started failing randomly.

openfisca_core/data_storage.py
maukoquiroga: Why does it feel weird ? In this case, as I'm not actually implementing anything, I could effectively just pick structural subtyping at not cost.

https://www.python.org/dev/peps/pep-0544/#explicitly-declaring-implementation

openfisca_core/data_storage.py
maukoquiroga: > We're adding tests

Check :)

> Changing behaviour

Should be avoided, thanks for the pointer.

> Changing the CI config

Was randomly changing

> Adding type annotations

That's documentation.

As I understand this project already adopts type annotations, and a progressive strategy to annotate code. I'm just doing what the linter does: check as I go.

I feel you don't like them, and that's fine. Just take one thing into consideration:

- I don't feel the current storage implementation is  good one
- In order to change it, I'm trying to get a sense of their behaviours
- so I can later move things around with easy, for example to provide an explicit cache API
- I type annotate before writing tests as contract hypothesis that I correct after testing

> Changing the style of imports

Should be avoided, thanks for the pointer.

> PRs that do many things at once are harder to review and more likely to contain unnoticed errors.

I'll try to reduce the noise, thanks !

tests/core/data_storage/test_in_memory_storage.py
maukoquiroga: Hello, I'm just trying to map the current behaviour, in an effort of exposing redundancies. Why? Because those are IMHO low-hanging fruits that'll helps us do what we want to do after: get rid of the current implementation.

tests/core/data_storage/test_in_memory_storage.py
maukoquiroga: Thanks, you're right.

tests/core/data_storage/test_in_memory_storage.py
maukoquiroga: I'm trying to expose the uselessness of some parts the current implementation to other contributors who can come after me. Would it be clearer if I add an explanation ? The objective of the test is to say "whatever you do, as long as it is an eternal definition period we're dealing with, won't have no effect whatsoever, so wonder why does this code is necessary at all".

tests/core/data_storage/test_on_disk_storage.py
maukoquiroga: It feels ok to me, although I'll stick to current convention for this library, wasn't intended, thanks for the pointer.

tests/core/data_storage/test_on_disk_storage.py
maukoquiroga: >  I don't have any intuitive sense of what is being guaranteed.

I'm trying to separate the caching logic with the data I/O logic, so the intention is to guarantee that, given the expected conditions, responsibility will be handed over from the caching layer to the disk I/O layer.

I'm not really interested on what happens after caching is done, so I'm just guaranteeing the correct message is being handed. That'll help us change the caching contract without affecting the data I/O one.

>  I have no idea what autospec does or why it's relevant here.

It makes sure you're respecting the actual contract of the method you're mocking. Which is exactly what we want : we can to make sure that three conditions are respected :

- when responsibility is handed over
- arity
- type

If those 3 are respected, I feel confident going and tinkering around with the caching layer.

> The test demands detailed knowledge of the implementation and thus fails to provide much value as documentation.

That's correct, do you prefer if I add some docstring to the test itself ? All this is a roadmap to get rid of these classes.

maukoquiroga:
> I would seek to eliminate them rather than to reinforce them.

The objective is to do so. 

I wanted to propose a pull request before I started actually getting rid of them.

But it may feel like just _work in progress_, which in fact is, a I continued working after this point on dismantling this two classes (hint: it is not easy).

I understand if you feel this PR doesn't add any value by itself that justifies the LOC+, you can just reject it.

IMHO being clear about the actual contracts is what this adds as value to facilitate dismantling them later (tests + annotations then).

Morendil:
> it may feel like just work in progress

Yup, I tend to work in the same way, "refactor to understand". I acknowledge that I have mostly negative things to say about this PR - this isn't to belittle the work itself. It's good to see progress on that bit of a tangle around inputs and cache !

As discussed earlier I don't think I'm in a position, organisationally speaking, to either approve or reject PRs to Core. We had a quick chat about @sandcha today about this and my perspective is that there are still two distinct styles of organisation at work here, one is an "internally funded" team (temporarily at DINUM and perhaps soon elsewhere), one an open source community in the usual sense. The overlap of the two is a bit awkward, at least at the moment, at least to me.

openfisca_core/data_storage.py
Morendil: > I feel you don't like them

Nah. I love explicit types, otherwise I wouldn't be doing so much Haskell at the moment :)

tests/core/data_storage/test_in_memory_storage.py
Morendil: OK, I get where you're coming from.

Keep in mind that all the information I have when you ping me for review is "I'm being asked for a review". If there is a type of feedback that's more (or less) useful to you, I'm happy to consider that before I start reviewing - I always make a stop at the PR's description before diving into the code itself, so that's a useful place (for me) to have information about intent and context.

By default, I'm going to assume that the last contract we discussed still applies: i.e. "would you be happy maintaining this code as it will be after this PR".

maukoquiroga: Thanks a lot! Yeah I think I'll have to go the extra mile for this to be mergeable then.

tests/core/data_storage/test_on_disk_storage.py
maukoquiroga: I'm reworking the API otherwise those tests won't make sense, major bump unavoidable ...

maukoquiroga:
@Morendil Thanks a lot for your review! This is my weekend's contribution :P

tests/core/data_storage/test_in_memory_storage.py
maukoquiroga: Changed !

tests/core/data_storage/test_on_disk_storage.py
maukoquiroga: Changed !

openfisca_core/data_storage.py
maukoquiroga: ❤️ tried to remove as much noise as I can! :)

maukoquiroga:
@Morendil gave it another try after your initial feedback :) any further advise would be highly appreciated.

openfisca_core/data_storage.py
maukoquiroga: Picked base classes in the end.

tests/core/data_storage/test_in_memory_storage.py
maukoquiroga: I did use periods in the end.

tests/core/data_storage/test_on_disk_storage.py
maukoquiroga: I changed of approach radically, so this code is no longer updated.

Morendil:
This still isn't the right tack IMO. It leaves the existing relationships between instances untouched, with the Variable -> Holder -> Cache linkage roughly reproducing the Variable -> Holder -> Storage scheme.

To sketch it out briefly, I believe the Cache belongs to the Simulation (one Cache per Simulation), and functions as a "two-dimensional dictionary" - or to put it another way, a dictionary whose keys are tuples, specifically `(variable, period)` tuples; the values are Numpy arrays as usual. The logic to deal with eternal values will likely be a thin adapter in front of this dictionary. To preserve the functionality afforded by OnDiskStorage we want to be able to persist and reload this dictionary but this is a relatively trivial side-feature, not part of the core logic.

The Cache functionality should be relatively compact, in LOC terms, at parity with existing functionality; where it starts getting potentially more complex is when we add masking.

So the way I'd approach this work would be to take the point of view of the Simulation, and pretend Holders and Storage and all this stuff don't exist at all. From the perspective of the Simulation, all I need is a way to answer "do I already have a value for this variable at this period that I can use, or do I need to compute it", and of course a way to register "I've just computed a value for this variable at this period so I'll store it to avoid computing it again".

Separately, test that the Cache state of a simulation can be persisted and loaded again. This might be as simple as serializing and deserializing the array; while we're still not clear on how that feature is used in practice and exactly what needs it supports, I suspect the persistence medium is an implementation detail, *and* the current implementation doesn't best reflect those needs. So shunting that feature "to the side" gives us more flexibility.

Since this is all essentially a matter of side-effects mocks would make some sense to me in testing this; request the same computation twice on a Simulation and check that only the first request causes a call to `calculate` for instance, with the second instead resulting in a Cache lookup.

maukoquiroga:
Thanks a lot, starts looking like a roadmap.

> So the way I'd approach this work would be to take the point of view of the Simulation, and pretend Holders and Storage and all this stuff don't exist at all. From the perspective of the Simulation, all I need is a way to answer "do I already have a value for this variable at this period that I can use, or do I need to compute it", and of course a way to register "I've just computed a value for this variable at this period so I'll store it to avoid computing it again".

> Separately, test that the Cache state of a simulation can be persisted and loaded again. This might be as simple as serializing and deserializing the array; while we're still not clear on how that feature is used in practice and exactly what needs it supports, I suspect the persistence medium is an implementation detail, and the current implementation doesn't best reflect those needs. So shunting that feature "to the side" gives us more flexibility.

DiskStorage always persists on disk, and you can dump and restore as expected, it works quite nicely (there's a test for that™). What happens is that all stored files are destroyed once the cache object is garbage collected (`__del__(self)`), _unless_ you tell the cache to do not do so. Again here, most of this caching is `variable` caching, `entities` and so on are handled somewhere else ...

As for array serializing and deseriarializing, I think it's being do like :

```py
numpy.dump(enum.posible_values).
EnumArray(file, enum).
```

In my humble prespespective, It would be simply to just:

```py
hashed_array = tuple(map(tuple, array)
array = numpty.asarray(hashed_arraty)

### in the case you'd like to have

{(variable, periosd, hashed_mask): numpy.ndarray)}
```

One additional test would be to do a restore from a set of fixture files that are independent of the test case.

maukoquiroga:
Here's what I've been thinking about masks, I can't grasp if I truly understood the idea @Morendil and @sandcha :

```py
In [95]: v = 1                                                                                                                                                                                              

In [96]: p = 2                                                                                                                                                                                              

In [97]: m = tuple(map(bool, numpy.array([True, False, True])))                                                                                                                                             

In [98]: key = (v,p,m)                                                                                                                                                                                      

In [99]: state = {key: k}                                                                                                                                                                                   

In [100]: k = numpy.array([100, 200, 500])                                                                                                                                                                  

In [101]: numpy.array(key[2]) * state[key]                                                                                                                                                                  
Out[101]: array([100,   0, 500])

```

maukoquiroga:
Another example :

```py
import numpy

from typing import NamedTuple


class Key(NamedTuple):
    variable = 2
    period = 3
    mask = numpy.array([True, True, False, True])


class Cache:

    state: {}

    def __init__(self):
        self.state = {}


    def put(self, key: NamedTuple, value: numpy.ndarray) -> None:
        self.state[(key.variable, key.period, tuple(map(bool, key.mask)))] = value


    def get(self, key: NamedTuple) -> {}:
        return self.state[(key.variable, key.period, tuple(map(bool, key.mask)))]


cache = Cache()
value = numpy.array([10, 20, 30, 40])
cache.put(Key(), value)
result = cache.get(Key())

print(cache.state)
print(result)
print(value)

```

Morendil:
@maukoquiroga With respect to masking see the [prior discussion](https://github.com/openfisca/openfisca-core/pull/871#issuecomment-497712017) which outlined the algorithm.

When we spoke yesterday I implied that the mask would be part of the cache key but that was an oversimplification, so it's possible we'll end up dealing with just `(variable, period)` as the cache key and have to overlay the masking logic on top of that. Sorry for being misleading.

That doesn't diminish the value of simplifying the Holders/Storage tangle.

Also one important thing missing from that roadmap, which should have high priority, is figuring out how we're going to remove all the calls to `get_holder` from the France model (which as a side-effect will answer the question "how do we advise other country package maintainers deal with the breaking change of removing Holders").

There are two clusters: one of them around age computations (which is why I worked on the age-related PR in France in the first place), the other around "inversions", such as the Reform that computes gross salary based on net.

As @sandcha noted the way of dealing with ages that I suggested has some downsides, but it helps getting rid of get_holder calls; @sandcha has proposed that we implement masking so that we can perform the age computations in France without the awkwardness but implementing masking requires having that Cache API which requires getting rid of holders: we can't allow circular dependencies in our [Mikado map](http://www.methodsandtools.com/archive/mikado.php)! So, perhaps a better plan is to temporarily change the age computations to that awkward method and later make them more flexible again.

I would suggest that inversions should be removed entirely from France (and explicitly documented as unsupported) for the time being as they are not properly "reforms" (speculative encodings of counterfactual law) but explorations of an alternative `calculate` protocol. Inversions require fine control of the computation cache, since the entire point of them is to repeatedly calculate the same target value based on different inputs - which invalidates the choice of `(variable, period)` as a cache key. They are a feature which Core lacks proper support for at the moment but will become easier to support as a result of two things, a cache API and a way to compute dependencies statically.

Morendil:
> there'll be no a single line of caching code inside the functions

I don't think of that as a goal; it would certainly be nice. In practice, the caching logic is currently interleaved with the Spiral detection logic in `Simulation::_calculate` and I'm not quite convinced we can cleanly separate the two yet.

maukoquiroga:
Thanks a lot @Morendil for your feedback, I'm closing as this RFC showed there might be another better strategy to deal with caching and getting rid of holders 😃

PR #951, opened 2020-03-29
Avoid Web API failure with dynamic variable generation

guillett:
#### Technical changes

- Avoid Web API failure with _dynamic_ variable generation

Using reforms to create _dynamic_ variables can lead to failure to start the API because introspection is failing to get a `source code` section


guillett:
Any changes that PR get reviewed ? It is preventing me from relying on master. Thanks.

sandcha:
@guillett Do you have an example that failed before this PR?
The last commit starts a test. You can update it if it helps.

maukoquiroga:
Hi @guillett! Same as @sandcha if you could provide a failing scenario that would be ideal.

maukoquiroga:
(I'm just guessing but that may be the same situation we faced a while ago with the LexImpact API)

guillett:
Many thanks to both of you.

I updated the reform to generate the bug.
`pytest tests/web_api/case_with_reform/reforms.py` does not fail even without a fix but `openfisca serve --country-package openfisca_country_template --reforms tests.web_api.case_with_reform.reforms.add_variable_reform` does fail and highlight the need of the suggested fix.

maukoquiroga:
So thanks to you @guillett, I adapted and moved the failing example to openfisca/country-template#98 just to be sure it was not a namespace problem (it is not).

maukoquiroga (approved):
Thanks for this @guillett 

So after a quick check it seems to be no satisfactory way of retrieving the source of a dynamically created class.

Regarding the solution, it terms of the WebAPI, `source_code` is an optional field so there's no breaking change.

Please let me know of you're ok with my changes, otherwise to roll them back.

guillett:
Thanks @maukoquiroga. I am OK with your change even if the use case is further from the test then before.
Yes, it is not a breaking change so we should be able to merge that easily.

Thank you to both of you.

maukoquiroga:
Yes sorry for that I didn't realise it was a problem with dynamic variables, but just with addîg new ones.

Sad news is that getting "rules as code" from a user interface generated reform will be far from evident (you only have the ast and bytecode to pull out some code-generating magic).

Could you review my country-package's PR? If it's all OK I can tale care of the integration a bit later today.

PR #964, opened 2020-11-18
Add dependencies endpoint

sarafalamaki:
#### Purpose

This new endpoint allows a user to find the dependent input variables to a variable. It's useful if you want to calculate a variable, and need to know what inputs you need. It allows you to find the answer to that question without having to manually go through all the variables (and their dependencies) until you get to the inputs.

It's also useful for gathering statistics about the use of input variables. If I want to design a form to get information from a user to assess their eligibility for a series of benefits, I could use this to find which input variables are used most often across different formulas, and put those at the start of the form. This could avoid having to ask the user too many questions, as you could design the form to exit early if the user isn't eligible.

I made it particularly for the Canadian project, where they wanted to get a list of input variables, and sort them based on use. You could do it in the front-end, but you'd have to parse python to do so. It makes more sense in the backend. It would have also been useful for NSW Community Gaming project, where we had to do this by hand to make the form.

#### New features

- New dependencies api end point that finds the dependent input variables of a variable.
- New property in variables.py to enable this.

#### Technical changes

- Made the API documentation for the trace endpoint a little clearer.

### DIFF
--- a/openfisca_core/entities.py
+++ b/openfisca_core/entities.py
@@ -40,17 +40,12 @@ class Entity(object):
             raise ValueError("{} is not a valid role".format(role))
 
     def get_variable(self, variable_name, check_existence = False):
+        if not self._tax_benefit_system:
+            raise ValueError("No tax_benefit_system set. Please set it using set_tax_benefit_system")
         return self._tax_benefit_system.get_variable(variable_name, check_existence)
 
     def check_variable_defined_for_entity(self, variable_name):
-        try:
-            variable_entity = self.get_variable(variable_name, check_existence = True).entity
-        except AttributeError as e:
-            import sys
-            import traceback
-            traceback.print_tb(sys.exc_info()[2])
-            print(e)
-            raise ValueError("FIXME: tax_benefit_system is null")
+        variable_entity = self.get_variable(variable_name, check_existence = True).entity
         # Should be this:
         # if variable_entity is not self:
         if variable_entity.key != self.key:
--- a/openfisca_core/entities.py
+++ b/openfisca_core/entities.py
@@ -43,7 +43,14 @@ class Entity(object):
         return self._tax_benefit_system.get_variable(variable_name, check_existence)
 
     def check_variable_defined_for_entity(self, variable_name):
-        variable_entity = self.get_variable(variable_name, check_existence = True).entity
+        try:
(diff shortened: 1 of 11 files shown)


sandcha:
Thanks @sarafalamaki for this contribution!
Could you describe the use cas that needs this feature? 
> Which user profile? What becomes possible with this feature? etc.

openfisca_core/entities.py
```diff
@@ -43,7 +43,14 @@ def get_variable(self, variable_name, check_existence = False):
...
     def check_variable_defined_for_entity(self, variable_name):
-        variable_entity = self.get_variable(variable_name, check_existence = True).entity
+        try:
+            variable_entity = self.get_variable(variable_name, check_existence = True).entity
+        except AttributeError as e:
+            import sys
+            import traceback
+            traceback.print_tb(sys.exc_info()[2])
+            print(e)
+            raise ValueError("FIXME: tax_benefit_system is null")
```
sandcha, line 53: Update the error message by replacing "where the problem is" by "how you can solve the problem"?
You can find some information on the error message in the [contribution guide](https://github.com/openfisca/openfisca-core/blob/master/CONTRIBUTING.md#error-messages) of the repository.

sandcha: > If this error impacts the `/dependencies` endpoint and when we deliver it to the web API response, we have to keep in mind that the reader may not know precisely how openfisca is built (following the _principle of least astonishment_).

sarafalamaki: The error says FIXME, because I wasn't sure if this is the right place to raise it.  I'm going to move it to the function above (where it originates) and push changes again. 

It shouldn't arise in the dependency API endpoint anymore, I had to set tax_benefit_system explicitly to mitigate it.

Please let me know if there's anything I've missed about the semantics of using the tax benefit system, as I hacked at it till it worked.

Br3nda:
@sarafalamaki this looks very very very useful!

Br3nda:
Definitely needs some tests

sarafalamaki:
> Definitely needs some tests

Do you know how to get the code coverage check to work? It requires some sort of token?

sarafalamaki:
I opened another pull request with identical changes so that the submit_coverage test would pass. They didn't work here as I didn't have the appropriate token in my personal github org.

PR #965, opened 2020-12-09
Increase verbosity

benjello:
#### Technical changes

- Improve error message when laoding parameters file to detect the problematic file

### DIFF
--- a/openfisca_core/parameters.py
+++ b/openfisca_core/parameters.py
@@ -835,7 +835,13 @@ def _load_yaml_file(file_path):
                 file_path,
                 stack_trace
                 )
-
+        except Exception:
+            stack_trace = traceback.format_exc()
+            raise ParameterParsingError(
+                "Invalid parameter file content. Check the traceback above for more details.",
+                file_path,
+                stack_trace
+                )
 
 def load_parameter_file(file_path, name = ''):
     """
@@ -847,12 +853,7 @@ def load_parameter_file(file_path, name = ''):
         raise ValueError("{} doest not exist".format(file_path))
     if os.path.isdir(file_path):
         return ParameterNode(name, directory_path = file_path)
-    try:
-        data = _load_yaml_file(file_path)
-    except Exception as e:
-        print("Problem when loading parameters file {}".format(file_path))
-        raise e
-
+    data = _load_yaml_file(file_path)
     return _parse_child(name, data, file_path)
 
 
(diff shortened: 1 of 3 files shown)


benjello:
@sandcha @maukoquiroga : i am pretty sure this is not the correct way of doing it. But this PR clearly makes my point and I am all ears to learn the right way to do this.

benjello:
Context: debugging utf8 troubles in parameters file.

sandcha:
@benjello Thank you for this contribution! 
In what circumstances did you get a bug? Do you have an example that reproduces this error?

openfisca_core/parameters.py
```diff
@@ -847,7 +847,12 @@ def load_parameter_file(file_path, name = ''):
         raise ValueError("{} doest not exist".format(file_path))
     if os.path.isdir(file_path):
         return ParameterNode(name, directory_path = file_path)
-    data = _load_yaml_file(file_path)
+    try:
+        data = _load_yaml_file(file_path)
+    except Exception as e:
+        print("Problem when loading parameters file {}".format(file_path))
```
sandcha, line 853: Update the message to suggest to the user how he could solve the problem?
> Your use case description will help to do so.

benjello: A problem with _load_yaml_file without any mention of the problematic file. In fine I found a latin1 character in the file.
See https://github.com/openfisca/openfisca-france/commit/1a7f96ff4c9bdebc4f523e15763be7ddb11dfd1f

benjello: To reproduce the error just use the branch https://github.com/openfisca/openfisca-france/tree/cotisations-fix just before the applying commit openfisca/openfisca-france@1a7f96f and and run 
```bash
openfisca test tests/cotisations/csg_remplacement.yaml
```

sandcha: Thank you! Here is a commit including the following observations:
* As print usage is prohibited here, we should use an exception instead and we already have `ParameterParsingError` (defined above in this same file) that works for our situation 🙌  
* Then, we try to put the error detection near the line that produces the error so I would put the `except` in `_load_yaml_file` (that already tries to parse one file content) instead of `load_parameter_file` (mainly checking a file path).

WDYT?

> I could have added the detected `UnicodeDecodeError` but as its error message is clear and as we might still need to catch any `Exception` to avoid the issue of not knowing in which file we have a problem, I chose the general `Exception`.
 
> Error checked with this command: `openfisca test -c openfisca_france tests/cotisations/csg_remplacement.yaml`

sandcha (approved):
Thank you @benjello! 
LGTM with the last commits 😊 
I leave the merge to you, with the name of the PR as description of the merge commit, if you agree with them.

sandcha:
P.S: The installation error should be fixed by the rebase.

openfisca_core/parameters.py
benjello: Perfect

benjello:
Thanks @sandcha !

PR #975, opened 2021-02-15
Refactor dependencies function

cesco-fran:
Refactor dependencies function


sarafalamaki (approved):
Looks great to me. We should probably use _get_variable_names in the other endpoints too, but that can be a different PR. I don't want to add complications to the current one!

PR #976, opened 2021-02-16
Add Typings

cesco-fran:
#### Technical changes

-  Add typings to the source code.

### DIFF
--- a/openfisca_core/holders.py
+++ b/openfisca_core/holders.py
@@ -6,16 +6,28 @@ import warnings
 
 import numpy as np
 import psutil
+from typing import TYPE_CHECKING, List, Union, Any, Optional, Dict
 
 from openfisca_core import periods
 from openfisca_core.commons import empty_clone
 from openfisca_core.data_storage import InMemoryStorage, OnDiskStorage
 from openfisca_core.errors import PeriodMismatchError
 from openfisca_core.indexed_enums import Enum
-from openfisca_core.periods import MONTH, YEAR, ETERNITY
+from openfisca_core.periods import MONTH, YEAR, ETERNITY, Period
 from openfisca_core.tools import eval_expression
 
+
+if TYPE_CHECKING:
+    from openfisca_core.populations import Population
+    from openfisca_core.variables import Variable
+    from openfisca_core.simulations import Simulation
+
+
 log = logging.getLogger(__name__)
+
+
+#TODO change with a more specific type
+ArrayLike = Any
 
 
 class Holder(object):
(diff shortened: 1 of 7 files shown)


openfisca_core/holders.py
```diff
@@ -6,41 +6,57 @@
...
 from openfisca_core.commons import empty_clone
 from openfisca_core.data_storage import InMemoryStorage, OnDiskStorage
 from openfisca_core.errors import PeriodMismatchError
 from openfisca_core.indexed_enums import Enum
-from openfisca_core.periods import MONTH, YEAR, ETERNITY
+from openfisca_core.periods import MONTH, YEAR, ETERNITY, Period
 from openfisca_core.tools import eval_expression
 
+
+if TYPE_CHECKING:
```
maukoquiroga, line 20: Why conditional ? We already run `mypy` in Makefile.

maukoquiroga:
I love this ❤ it adds some lecture overhead at first but having explicit contracts will definitely ease refactoring, bug correction, debugging... I would at a minimum create types mapping each important class. It's a layer of indirection maybe but I think it would make types more "drop-in" and ease code comprehension for most newcomers.

cesco-fran (commented):
Since is possible to use the class as type ...there will be no indirection for type that correspond to class.

maukoquiroga:
I think we should move forward with this, any help needed?

openfisca_core/holders.py
```diff
@@ -6,41 +6,57 @@
...
+    from openfisca_core.populations import Population
+    from openfisca_core.variables import Variable
+    from openfisca_core.simulations import Simulation
+
+
 log = logging.getLogger(__name__)
 
 
+#TODO change with a more specific type
+ArrayLike = Any
```
maukoquiroga, line 30: This will be shipped out of the box in Numpy 1.20/21

openfisca_core/holders.py
```diff
@@ -6,41 +6,57 @@
...
     """
         A holder keeps tracks of a variable values after they have been calculated, or set as an input.
     """
 
-    def __init__(self, variable, population):
-        self.population = population
-        self.variable = variable
-        self.simulation = population.simulation
-        self._memory_storage = InMemoryStorage(is_eternal = (self.variable.definition_period == ETERNITY))
+    def __init__(self, variable:'Variable', population:'Population'):
```
maukoquiroga, line 38: Flake won't like the lack of spaces. 

Also you can forward-reference them no?

openfisca_core/holders.py
```diff
@@ -6,41 +6,57 @@
...
+        self.population:'Population' = population
+        self.variable:'Variable' = variable
+        # TODO change once decided if simulation is needed or not
```
maukoquiroga, line 41: I think this requires a bit more of examination. I'll take a look.

openfisca_core/holders.py
```diff
@@ -6,41 +6,57 @@
...
+        if population.simulation is None:
+            raise Exception('You need a simulation attached to the population')
```
maukoquiroga, line 43: Error messages in OpenFisca are meant to help the users discover not just what is wrong, but how can they solve their problems or reach for help. Could you develop this further?

Issue #720, opened 2018-08-30
Create a method `asof` that returns the state of the tax benefit system at that date 


benjello:
Hi there!

I really enjoy OpenFisca, but I recently encountered an issue.

I would like to extract a tax benefit system _as of_ some specific date potentially to use in the future _as is_  time never changed anything for that specific date.

The steps that should do the job are:
 - [ ] erase all parameters after some date
 - [ ] erase all the formulas after some date for all the variables

## Context

I identify more as a:

- Developer _(I create tools that use the existing OpenFisca code)_.
- Historian _(I accumulate data based on past legislation values)_.
- Lobbyist _(I model reforms to make them exist)_.

@fpagnoux do you see other things that may cause problems ?

cc @claireleroy @bfabre01

benjello:
I implemented it as a standalone function in [another package](https://github.com/openfisca/openfisca-survey-manager/blob/master/openfisca_survey_manager/utils.py#L156).

@openfisca/core-contributors : 
 - do you think that this tool and the inflator mentionned in this [france issue](https://github.com/openfisca/openfisca-france/issues/1095) should make it to core ?
 - if so could you have a quick look at the code and tell me what I should take care of when moving it to core as a new tax_benefit_system method

benjello:
@openfisca/core-contrib  @Morendil @sandcha @fpagnoux @maukoquiroga : I almost finished a couple of tax-benefit system and parameters methods that are implemented as independent functions in survey-manager:
 - [`asof`](https://github.com/openfisca/openfisca-survey-manager/blob/master/openfisca_survey_manager/utils.py#L127) which freeze the tax-benefit-system as of some date
 -  [`inflate_parameters`](https://github.com/openfisca/openfisca-survey-manager/blob/master/openfisca_survey_manager/utils.py#L18) that modify all the parameters starting from some date by in inflator value provided by the user.

[Tests](https://github.com/openfisca/openfisca-survey-manager/blob/master/openfisca_survey_manager/tests/test_legislation_inflator.py) are here so you can grasp what these functions actually do.
  
I think the best place for this are methods for Parameters and TaxBenefitSystem objects and thus belong to. What is your opinion ?

benjello:
@Morendil @sandcha @fpagnoux: any advice ?I really want to do that but not in vain (PR not accepted as out of scope) ;-)

benjello:
[...]

PR #978, opened 2021-02-17
Asof parameters

benjello:
#### New features

- Introduce `VectorialAsofDateParameterNodeAtInstant`
  - Allows for extracting serialized parameters as of some date introduced as a `np.datetime64` vector 

- Example : The [test file](https://github.com/openfisca/openfisca-core/blob/d57d615b99c2df1e2cff928e44356ab7ee3d44ea/tests/core/parameters_date_indexing/test_date_indexing.py) shows parameters that depends on your date of birth being after or before some dates. So you need to get the parameters as of some date.

### DIFF
--- a/openfisca_core/parameters/vectorial_asof_date_parameter_node_at_instant.py
+++ b/openfisca_core/parameters/vectorial_asof_date_parameter_node_at_instant.py
@@ -1,0 +1,70 @@
+import numpy
+
+from openfisca_core import parameters
+from openfisca_core.errors import ParameterNotFoundError
+from openfisca_core.indexed_enums import Enum, EnumArray
+from openfisca_core.parameters import helpers
+
+
+
+
+class VectorialAsofDateParameterNodeAtInstant(parameters.VectorialParameterNodeAtInstant):
+    """
+    Parameter node of the legislation at a given instant which has been vectorized along osme date.
+    Vectorized parameters allow requests such as parameters.housing_benefit[date], where date is a np.datetime64 type vector
+    """
+
+    @staticmethod
+    def build_from_node(node):
+        parameters.VectorialParameterNodeAtInstant.check_node_vectorisable(node)
+        subnodes_name = node._children.keys()
+        # Recursively vectorize the children of the node
+        vectorial_subnodes = tuple([
+            VectorialAsofDateParameterNodeAtInstant.build_from_node(node[subnode_name]).vector
+            if isinstance(node[subnode_name], parameters.ParameterNodeAtInstant)
+            else node[subnode_name]
+            for subnode_name in subnodes_name
+            ])
+        # A vectorial node is a wrapper around a numpy recarray
+        # We first build the recarray
+        recarray = numpy.array(
+            [vectorial_subnodes],
+            dtype = [
+                (subnode_name, subnode.dtype if isinstance(subnode, numpy.recarray) else 'float')
+                for (subnode_name, subnode) in zip(subnodes_name, vectorial_subnodes)
+                ]
+            )
+        return VectorialAsofDateParameterNodeAtInstant(node._name, recarray.view(numpy.recarray), node._instant_str)
+
+    def __getitem__(self, key):
+        # If the key is a string, just get the subnode
+        if isinstance(key, str):
+            key = numpy.array([key], dtype = 'datetime64[D]')
+            return self.__getattr__(key)
+        # If the key is a vector, e.g. ['1990-11-25', '1983-04-17', '1969-09-09']
+        elif isinstance(key, numpy.ndarray):
+            assert numpy.issubdtype(key.dtype, numpy.datetime64)
+            names = list(self.dtype.names)  # Get all the names of the subnodes, e.g. ['ne_avant_X', 'ne_apres_X', 'ne_apres_Y']
+            values = numpy.asarray([value for value in self.vector[0]])
+            names = [
+                name
+                for name in names
+                if not name.startswith("ne_avant")
+                ]
+            names = [
+                numpy.datetime64(
+                    "-".join(name[9:].split("_"))
+                    )
+                for name in names
+                ]
+            conditions = sum([
+                name <= key
+                for name in names
+                ])
+            result = values[conditions]
+
+            # If the result is not a leaf, wrap the result in a vectorial node.
+            if numpy.issubdtype(result.dtype, numpy.record):
+                return VectorialAsofDateParameterNodeAtInstant(self._name, result.view(numpy.recarray), self._instant_str)
+
+            return result
--- a/openfisca_core/parameters/vectorial_asof_date_parameter_node_at_instant.py
+++ b/openfisca_core/parameters/vectorial_asof_date_parameter_node_at_instant.py
@@ -39,16 +39,16 @@ class VectorialAsofDateParameterNodeAtInstant(parameters.VectorialParameterNodeA
         # If the key is a vector, e.g. ['1990-11-25', '1983-04-17', '1969-09-09']
         elif isinstance(key, numpy.ndarray):
             assert numpy.issubdtype(key.dtype, numpy.datetime64)
-            names = list(self.dtype.names)  # Get all the names of the subnodes, e.g. ['ne_avant_X', 'ne_apres_X', 'ne_apres_Y']
+            names = list(self.dtype.names)  # Get all the names of the subnodes, e.g. ['before_X', 'after_X', 'after_Y']
             values = numpy.asarray([value for value in self.vector[0]])
             names = [
                 name
                 for name in names
-                if not name.startswith("ne_avant")
+                if not name.startswith("before")
                 ]
             names = [
                 numpy.datetime64(
-                    "-".join(name[9:].split("_"))
+                    "-".join(name[len("after_"):].split("_"))
                     )
                 for name in names
                 ]
--- a/openfisca_core/parameters/vectorial_asof_date_parameter_node_at_instant.py
+++ b/openfisca_core/parameters/vectorial_asof_date_parameter_node_at_instant.py
@@ -1,11 +1,6 @@
 import numpy
 
 from openfisca_core import parameters
-from openfisca_core.errors import ParameterNotFoundError
-from openfisca_core.indexed_enums import Enum, EnumArray
-from openfisca_core.parameters import helpers
-
-
 
 
 class VectorialAsofDateParameterNodeAtInstant(parameters.VectorialParameterNodeAtInstant):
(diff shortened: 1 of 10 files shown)


benjello:
@sandcha @guillett @maukoquiroga @eraviart : I know this PR is extremely ugly and I intent to improve it. But I definitively need your advice on how to do that.

benjello:
@maukoquiroga : you went through a large refactoring and I am quite lost to rebase my branch on master ...
Could you rebase it for me? Thanks.

sandcha:
Does this PR answer this issue: https://github.com/openfisca/openfisca-core/issues/720 ? 🙂

benjello:
> Does this PR answer this issue: #720 ?

No, it is a complete different topic.

sandcha:
Rebasing the branch as discussed with @benjello.

benjello:
@nikhilwoodruff @eraviart @sandcha @MattiSG : 

I need your input to imporve this PR. In particular for the prefix of the generation (`ne_apres_YYYY_MM-DD`, `ne_avant_YYYY_MM-DD`).
Do `after_YYYY_MM-DD`, `before_YYYY_MM-DD`  would be more generic ?

nikhilwoodruff:
This looks nice to me. Of course my preference would definitely be `before`/`after` but feel free to outvote this.

nikhilwoodruff (approved):
OK @benjello think I understand this a bit better now: this is a very cool feature and I like the implementation and interface.

sandcha:
Thank you for this PR and for the examples!
Thanks to [the parameters_date_indexing/trimtp_rg.yml](https://github.com/openfisca/openfisca-core/blob/18124dbba4fd432c5f540f93976e036bfb947ace/tests/core/parameters_date_indexing/trimtp_rg.yml), I understand that what we call `asof` here could improve the modelization of some use cases (and that it has nothing to do with reforms 😅). Possible use cases: 
* Pensions that depend on the birth date or career of persons and therefore, they depend on different dates specific to the person/entity.
* In France for example, the merit benefit for students where the eligibility and amount depend on the year of the baccalaureate/general certificate of education of the student/entity ([description in French](https://www.service-public.fr/particuliers/vosdroits/F1010)). 
 
So far, when we filtered parameters by date, it was the simulation period or, more precisely, the date on which we observed the applicable law. Here, if I'm right, we have use cases where at the same time, we observe different applicable law options that depend on a date specific to the entity. So, we need to access a parameter by a double index: the usual "simulation period" and a new index, a "date" index.

So, it looks great to have some mechanism to manage these situations! 
But I have some doubts about the implementation:
* We introduce an entirely new parameter format while we already have double indexed parameters where one is the simulation period and the other is some entity information (the tax scales formats).
* The chosen filter here (`before_*`, [after_*](https://github.com/openfisca/openfisca-core/blob/18124dbba4fd432c5f540f93976e036bfb947ace/tests/core/parameters_date_indexing/trimtp_rg.yml#L9)) is not a simple string key. We need to parse it and the algorithm will have to adapt to the specific prefix of the item: it mixes a bit between the current abstraction levels in the parameter management and yaml specification. Besides, parsing a string and having keys to parse that depend on each other (like `before_1934_01_01` and `after_1934_01_01`) implies to manage more error messages.
* The typing of the parameters is managed between [parameters/](https://github.com/openfisca/openfisca-core/tree/master/openfisca_core/parameters/) and [taxscales/](https://github.com/openfisca/openfisca-core/tree/master/openfisca_core/taxscales). If we introduce the `VectorialAsofDateParameterNodeAtInstant`, it looks like we are bypassing some steps (between a `ParameterNodeAtInstant` and the Python class that defines the object like those we have for the scales).

@benjello @nikhilwoodruff What do you think about these implementation questions? cc @maukoquiroga @MattiSG

sandcha:
One solution might be to see the `asof` parameter as a scale that evolves by date: the thresholds values would be dates as they are naturally ordered and could be filtered with the usual `.calc(information from the entity)` method.

Here is a first example : 
This extract from [trimtp_rg.yml](https://github.com/openfisca/openfisca-core/blob/18124dbba4fd432c5f540f93976e036bfb947ace/tests/core/parameters_date_indexing/trimtp_rg.yml)):
```yaml
nombre_trimestres_cibles_par_generation:
  description: Nombre de trimestres cibles par génération
  before_1934_01_01:
    description: Avant 1934
    values:
      1983-01-01:
        value: 150.0
  after_1934_01_01:
    description: '1934-01-01'
    values:
      1994-01-01:
        value: 151.0
      1983-01-01:
        value: null
```

Could be modeled as:
```yaml
number_targeted_quarters_per_generation:
  description: The number of targeted contribution quarters per generation
brackets:
  - amount: 
      1983-01-01:
        value: 150.0
    1994-01-01:
        value: null
    threshold: 
      1983-01-01:
        value: 0001-01-01 # from python/openfisca default date to next threshold
  - amount: 
      1994-01-01:
         value: 151.0
    threshold: 
      1983-01-01:
        value: 1934-01-01 # born after January 1934
```

tests/core/parameters_date_indexing/aad_rg.yaml
```diff
@@ -0,0 +1,121 @@
+description: Age d'annulation de la décote
```
sandcha, line 1: As openfisca-core is an international repository, use [English as default language](https://openfisca.org/doc/contribute/language.html#language) for the filename and content?

tests/core/parameters_date_indexing/aad_rg.yaml
```diff
@@ -0,0 +1,121 @@
...
+age_annulation_decote_en_fonction_date_naissance:
```
sandcha, line 2: Do the parameters of `aad_rg.yaml` and `trimtp_rg.yml` show other mechanisms than the `before_YYYY_MM_DD` / `after_YYYY_MM_DD` format? 🤔 If they don't, we could shorten all the (translated) examples to highlight what they are describing.

tests/core/parameters_date_indexing/test_date_indexing.py
```diff
@@ -0,0 +1,38 @@
...
+
+def test_on_node():
+    date_de_naissance = numpy.array(['1950-01-01', '1953-01-01', '1956-01-01', '1959-01-01'], dtype = 'datetime64[D]')
+    parameter_at_instant = parameters.aad_rg('2012-03-01')
+    node = parameter_at_instant.age_annulation_decote_en_fonction_date_naissance[date_de_naissance]
+    assert_near(node.annee, [65, 66, 67, 67])
+    assert_near(node.mois, [0, 2, 0, 0])
+
+
+# def test_inhomogenous():
```
sandcha, line 32: Add this commented test?

tests/core/parameters_fancy_indexing/coefficient_de_minoration.yaml
```diff
@@ -0,0 +1,135 @@
+description: Coefficient de minoration ARRCO
```
sandcha, line 1: This parameter isn't used, is it?
Was it used as an example to explicit some use cas of the `asof` parameter as it doesn't contain `before...` and `after...` keys? Add it in the test or remove the file if it is no longer used?

sandcha (changes requested):
Thank you for reopening this subject!
Whatever we decide on the parameter format subject, here are some request changes on the examples.

> This review doesn't really address the `VectorialParameterNodeAtInstant` object and the tests as they depend on the parameter format.

tests/core/parameters_date_indexing/test_date_indexing.py
benjello: We should ...

tests/core/parameters_fancy_indexing/coefficient_de_minoration.yaml
benjello: Yes could be removed

tests/core/parameters_date_indexing/aad_rg.yaml
benjello: Yes a translation is needed

tests/core/parameters_date_indexing/aad_rg.yaml
benjello: Yes a translation is needed

benjello:
@sandcha @nikhilwoodruff : I think we should go for a more general multiindexed array parameters.
What do you think of something like the following:

```yaml
quarters_by_generation:
  description:  Quarters needed to get full retirement by generation
  index:
    generation:
      description: date of birth
      type: date
      selection: asof # could be exact by default or any other  
    # period is the implicit 0-axis
    # We can have more than one more axis.
  values:
    null:
      1983-01-01:
        value: 150.0
    1934-01-01:
      1994-01-01:
        value: 151.0
      1983-01-01:
        value: null
```

cc @eraviart @guillett @maukoquiroga

nikhilwoodruff:
> @sandcha @nikhilwoodruff : I think we should go for a more general multiindexed array parameters. What do you think of something like the following:
> 
> ```yaml
> quarters_by_generation:
>   description:  Quarters needed to get full retirement by generation
>   index:
>     generation:
>       description: date of birth
>       type: date
>       selection: asof # could be exact by default or any other  
>     # period is the implicit 0-axis
>     # We can have more than one more axis.
>   values:
>     null:
>       1983-01-01:
>         value: 150.0
>     1934-01-01:
>       1994-01-01:
>         value: 151.0
>       1983-01-01:
>         value: null
> ```
> 
> cc @eraviart @guillett @maukoquiroga

I think that scheme would work great for us, @benjello - I like that it's more streamlined. I'm actually just re-implementing a part of the UK system that I think this feature would probably simplify a lot: the Child Tax Credit has a limit on the number of children eligible (2), but when testing eligibility for each child, we need to check against the child limit at the date of their birth, rather than at the current date. Am I right in thinking that with this PR, we'd just need to do something like the below parameter and formula?
```yaml
description: Child limit
values: 
  null:
    null: .inf
  2017-04-06:
    2017-04-06: 2
```

```python
def formula(person, period, parameters):
  birth_date = person("birth_date", period)
  child_index = person("child_index", period)
  meets_limit = child_index < parameters(period).child_limit[birth_date]
  return meets_limit
```

openfisca_core/parameters/vectorial_asof_date_parameter_node_at_instant.py
```diff
@@ -0,0 +1,65 @@
+import numpy
+
+from openfisca_core import parameters
+
+
+class VectorialAsofDateParameterNodeAtInstant(parameters.VectorialParameterNodeAtInstant):
```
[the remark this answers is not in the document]
benjello: Why not but I don't think this is needed since we do not have a 'VectorialParameter' class but only a 'VectorialParameterNodeAtInstant' ?

benjello:
Implemented in #1212

Issue #966, opened 2020-12-22
Add support for WEEK period


nikhilwoodruff:
Hi there!

Firstly, thanks to the OpenFisca community for enabling us to use such a neat framework. In developing OpenFisca-UK, many variables including the majority of state benefits are specified in weekly amounts, which would be ideal to use as a standard period. I'm aware this has been discussed previously in #670 and #763.

As I understand it the main problems are identified in [this comment](https://github.com/openfisca/openfisca-core/issues/670#issuecomment-391100516):

> About weeks, this is going to bring tricky questions about _relationship between periods_.
> 
> Right now, OpenFisca period model relies on the assumption that period fit well together. A `YEAR` is composed a 12 `MONTH`s . We could relatively easily add `DAY`, as a month is composed of 28 to 31 `DAY`s. We actually used to support them, but deprecated it as no one was using it.
> 
> However, `WEEK` is more challenging: they don't fit well in a month, nor in a year. And question will arise:
> 
> * If `A`'s definition period is the week, can I calculate `A` for a year?
> * If so, what do I do with the weeks that are only partially included in the year?
> * Can we have a _generic_ answer to the previous question, or is it going to be _country-specific_?
> 

OpenFisca-UK has been relying off an implementation of the WEEK period in [this fork](https://github.com/nikhilwoodruff/openfisca-core/tree/week-period), but it'd be good to see if this is something that can benefit other country packages. In our implementation, we answered the above questions with:

> * If `A`'s definition period is the week, can I calculate `A` for a year?

Yes.

> * If so, what do I do with the weeks that are only partially included in the year?

Include the week of the first day of the year, and add weeks until we have reached (```DAYS_IN_YEAR // 7```) weeks. This also applies to months, respectively.

> * Can we have a _generic_ answer to the previous question, or is it going to be _country-specific_?

Use Monday as the first day of the week (this [might be different](https://en.wikipedia.org/wiki/Week#:~:text=While%2C%20for%20example%2C%20the%20United,first%20day%20of%20the%20week.) in other countries)

[...]

PR #979, opened 2021-03-11
Add support for WEEK/DAY period

maukoquiroga:
#### New features

This is a test-driven attempt to capitalise on #966 from @nikhilwoodruff, to add support for weeks in OpenFisca.

The main difference is that this draft makes a distinction between `days` and `weekdays`.

In doing so, we can treat both ISO dates and ISO week dates interchangeably.

Scope is yet more limited than your examples @nikhilwoodruff.

> Right now, OpenFisca period model relies on the assumption that period fit well together. A `YEAR` is composed a 12 `MONTH`s .

My idea would be to rely on `Instant` for values, and let Period handle both date types, `2014-01-03` and `2014-W3-1`.

We could either let the parser do that with the actual `type` string, or separate the two periods in different classes and let `Period` orchestrate/dispatch between the two as needed.

> * If `A`'s definition period is the week, can I calculate `A` for a year?

Yes, as long as we identify common years of 52W or leap years of 53W, or that we use the average of ~52.2 weeks per year.

> * If so, what do I do with the weeks that are only partially included in the year?

TBD

> * Can we have a _generic_ answer to the previous question, or is it going to be _country-specific_?

TBD

Comments? @sandcha @guillett @Morendil @Br3nda @sarafalamaki @benjello et al.


maukoquiroga:
Superseded by #1146

PR #981, opened 2021-03-16
Fix test_variable_documentation failing test

maukoquiroga:
#### Technical changes

- Fixes failing tests introduced by openfisca-core/country-template#97
- Fixes #980

Issue #980, opened 2021-03-16
Failing Web API tests


maukoquiroga:
Hi there!

I really enjoy OpenFisca, but I recently encountered an issue.

### Here is what I did:

```
make test
```

### Here is what I expected to happen:

```
508 passed, 1 xfailed, 32 warnings in 30.34s
```

### Here is what actually happened:

```
========================================================================================= short test summary info ==========================================================================================
FAILED tests/web_api/test_variables.py::test_variable_formula_content - assert 'def formula(...me_tax_rate\n' == 'def formula(...me_tax_rate\n'
FAILED tests/web_api/test_variables.py::test_variable_documentation - AssertionError: assert '\nHousing al... necessary.\n' == '\nTo compute... necessary.\n'
========================================================================== 2 failed, 506 passed, 1 xfailed, 32 warnings in 49.60s ==========================================================================
make: *** [test] Error 1
```

## Context

Some Web API tests rely on specifics of openfisca/country-template implementation, specifically literal strings - the objective is to be sure we can obtain de string literals of variables and their documentation via the Web API.

Or, as we might expect, that's fragile, as any change to the string representation of those elements will make the tests fail.

The present failing tests were introduced by openfisca/country-template#97

I identify more as a:

- Developer _(I create tools that use the existing OpenFisca code)_.

PR #982, opened 2021-03-17
Allow array parameters

MattiSG:
#### Technical changes

- Support arrays as parameters.
- Improves wrong parameter types error message.

- - -

This was added to move more constants into parameters. For example, in OpenFisca-France: https://github.com/openfisca/openfisca-france/blob/master/openfisca_france/model/caracteristiques_socio_demographiques/demographie.py#L3-L35

- - -

TODO :

- [x] Add Web API tests.
- [x] Update [documentation](https://openfisca.org/doc/key-concepts/parameters.html) which currently states that parameters are “numeric parameters”.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-### 35.2.0 [#982](https://github.com/openfisca/openfisca-core/pull/982)
+## 35.2.0 [#982](https://github.com/openfisca/openfisca-core/pull/982)
 
 #### Technical changes
 
@@ -12,7 +12,7 @@
 
 - Fix false negative web API test following an update in the country template used for testing.
 
-### 35.1.0 [#973](https://github.com/openfisca/openfisca-core/pull/973)
+## 35.1.0 [#973](https://github.com/openfisca/openfisca-core/pull/973)
 
 #### Technical changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+### 35.2.0 [#982](https://github.com/openfisca/openfisca-core/pull/982)
+
+#### Technical changes
+
+- Allow parameters to be arrays.
+
 ### 35.1.1 [#981](https://github.com/openfisca/openfisca-core/pull/981)
 
 #### Technical changes
--- a/openfisca_core/parameters.py
+++ b/openfisca_core/parameters.py
@@ -10,7 +10,6 @@ import logging
 import os
 import sys
 import traceback
-import typing
 
 import yaml
 import numpy as np
@@ -37,7 +36,7 @@ FILE_EXTENSIONS = {'.yaml', '.yml'}
 # 'unit' and 'reference' are only listed here for backward compatibility.
 #  It is now recommended to include them in metadata, until a common consensus emerges.
 COMMON_KEYS = {'description', 'metadata', 'unit', 'reference', 'documentation'}
-ALLOWED_PARAM_TYPES = (float, int, bool, type(None), typing.List)
+ALLOWED_PARAM_TYPES = (float, int, bool, type(None), List)
 
 
 def date_constructor(loader, node):
--- a/openfisca_core/parameters.py
+++ b/openfisca_core/parameters.py
@@ -10,6 +10,7 @@ import logging
 import os
 import sys
 import traceback
+import typing
 
 import yaml
 import numpy as np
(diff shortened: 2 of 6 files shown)


MattiSG:
CI should be fixed by #981.

benjello (approved):

MattiSG:
Output from web-api should be checked before merging.

MattiSG:
@benjello can you please confirm this should not damage vectorial computation performance?

MattiSG:
I wrote the associated test in a manner decoupled from the country template, in order to avoid a repeat from #980 and in prevision of #984. However, that means discoverability from this feature will be limited in the country template.

MattiSG:
Documentation update can be reviewed at https://github.com/openfisca/openfisca-doc/pull/234.

benjello:
> @benjello can you please confirm this should not damage vectorial computation performance?

I do not see why this could damage it. LGTM.

MattiSG:
API output tested manually on a fork of the country template, got the following result:

```json
{
  "description": "Array types should be allowed.",
  "id": "general.array_type",
  "metadata": {},
  "source": "https://github.com/openfisca/country-template/blob/3.12.4/openfisca_country_template/parameters/general/array_type.yaml",
  "values": {
    "2018-06-01": ["ES", "FR", "IT", "NZ"],
    "2013-01-01": ["FR"]
  }
}
```

Not sure about how the legislation explorer will format these results, though.

openfisca_core/parameters.py
```diff
@@ -36,7 +37,7 @@
 # 'unit' and 'reference' are only listed here for backward compatibility.
 #  It is now recommended to include them in metadata, until a common consensus emerges.
 COMMON_KEYS = {'description', 'metadata', 'unit', 'reference', 'documentation'}
-ALLOWED_PARAM_TYPES = (float, int, bool, type(None))
+ALLOWED_PARAM_TYPES = (float, int, bool, type(None), typing.List)
```
maukoquiroga, line 40: Any preference between ```typing.List``` and just ```List```? Asking just for consistency (current usage is ```from typing import XXX```, but that is not consistent with other libraries usage, for example ```numpy``` and so on).

openfisca_core/parameters.py
```diff
@@ -851,7 +852,7 @@ def load_parameter_file(file_path, name = ''):
     :returns: An instance of :any:`ParameterNode` or :any:`Scale` or :any:`Parameter`.
     """
     if not os.path.exists(file_path):
-        raise ValueError("{} doest not exist".format(file_path))
+        raise ValueError("{} does not exist".format(file_path))
```
maukoquiroga, line 855: Good catch!

openfisca_core/parameters.py
MattiSG: No preference.

maukoquiroga: Could you use ```from typing import List``` just for consistency? Otherwise GTM :)

tests/core/parameter_validation/test_parameter_validation.py
```diff
@@ -37,7 +37,12 @@ def check(file_name, keywords):
     ])
 def test_parsing_errors(test):
     with pytest.raises(ParameterParsingError):
-        check(*test)
+        check_fails_with_message(*test)
+
+
+def test_array_type():
+    path = os.path.join(BASE_DIR, 'array_type.yaml')
+    load_parameter_file(path, 'array_type')
```
maukoquiroga, line 45:
```suggestion
    assert load_parameter_file(path, 'array_type')
```

openfisca_core/parameters.py
```diff
@@ -10,6 +10,7 @@
 import os
 import sys
 import traceback
+import typing
```
maukoquiroga, line 13:
```suggestion
from typing import List
```

openfisca_core/parameters.py
maukoquiroga, line 40:
```suggestion
ALLOWED_PARAM_TYPES = (float, int, bool, type(None), List)
```

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+### 35.2.0 [#982](https://github.com/openfisca/openfisca-core/pull/982)
```
maukoquiroga, line 3:
```suggestion
## 35.2.0 [#982](https://github.com/openfisca/openfisca-core/pull/982)
```

maukoquiroga (approved):
Thanks for this contribution! Looks good to me ✨

tests/core/parameter_validation/test_parameter_validation.py
maukoquiroga: It will anyway pass as it is a truthy value -otherwise it would fail. Just a proposition for reader's clarity.

MattiSG:
@maukoquiroga @benjello you approved this PR —thanks again! 🙂 Would you mind spending 3 minutes reviewing the associated doc update? It's really short and would avoid an outdated documentation to create confusion: https://github.com/openfisca/openfisca-doc/pull/234
Thanks! 🙏

maukoquiroga:
Thanks @MattiSG and @benjello for the review :)

PR #983, opened 2021-03-18
Add variable metadata

RamParameswaran:
#### New features

- Introduce `metadata` attribute to Variable class
  - `metadata` is a python dict
  - it accepts all basic types and can be nested to arbitrary depth
- Adds `metadata` key to the web-api route "/variable/&lt;id&gt;"
  - converts dict to JSON and includes in Response 

#### Use Case

- Allows the addition of arbitrary structured metadata to OpenFisca variables
- Allows retrieval of that structured metadata in the web-api
  - Can allow for presenting additional context-specific details in frontend applications, such as structured regulation references [like these](https://github.com/Openfisca-NSW/openfisca_nsw_community_gaming/blob/master/openfisca_nsw_community_gaming/variables/regulation_reference.py)
  - Can allow for filtering/search in frontend applications ...

This feature is essentially an extension of the `reference` attribute, but allows for arbitrary dictionaries instead of just strings and lists of strings...

#### Example

Example of structured metadata:
```
class ABCD(Variable):
    ...
    metadata = {"regulation_reference": {"part": 5,
                                         "clause": "A",
                                         "schedule": "34b",
                                         "notes": ["foo", "bar"],
                                         },
                "is_output": True                        
                }
```

Example web-api Response for `GET /variables/ABCD/`:
```
{
  id: "ABCD",
    ...
  metadata:{"regulation_reference": {"part": 5,
                                     "clause": "A",
                                     "schedule": "34b",
                                     "notes": ["foo", "bar"],
                                      },
            "is_output": True                        
            }
}

```

#### Notes

- Tests for the `web_api` Response are tightly coupled to the "openfisca_country_template" repo, so I've (for the moment) forgone testing for the `web_api` Response.
- Failing tests are fixed by branch 'fix-doc-test-api'


RamParameswaran:
If #984 is accepted, I can add web_api tests as appropriate.

PR #986, opened 2021-03-22
Fix failing web API test

### LINKED ISSUE #980: Failing Web API tests

MattiSG:
### Technical changes

- Fixes failing tests introduced by openfisca-core/country-template#97
- Fixes #980
- Alternative implementation of #981.

### DIFF
--- a/tests/web_api/test_variables.py
+++ b/tests/web_api/test_variables.py
@@ -94,8 +94,9 @@ def test_variable_formula_github_link():
 
 
 def test_variable_formula_content():
-    formula_code = "def formula(person, period, parameters):\n    return person('salary', period) * parameters(period).taxes.income_tax_rate\n"
-    assert variable['formulas']['0001-01-01']['content'] == formula_code
+    content = variable['formulas']['0001-01-01']['content']
+    assert "def formula(person, period, parameters):" in content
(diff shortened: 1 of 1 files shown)


augusto-herrmann (commented):
I like this approach to testing responses.

PR #954, opened 2020-05-09
Upgrade numpy dependency to 1.18

sandcha:
Fixes #924

⚠️ As this is possibly a breaking change that affects every formula, this PR should be checked with at least one country model updated to numpy 1.18 as well (and the country-template is too simple to be a good example of the effect of this update 🙂).

---

`select` deprecations were described in [this comment](https://github.com/openfisca/openfisca-core/issues/924#issuecomment-615512006)

Add checks at `select` routine usage when a call to the function using `select` might lead to an empty condition.

No update on conditions types as we already used booleans.

---

#### Breaking changes

- Bump `numpy` dependency to `1.18`

PR #758, opened 2018-11-06
Update numpy requirement from <1.15,>=1.11 to >=1.11,<1.16

Br3nda:
#### Technical changes

- Updates the requirements on [numpy](http://www.numpy.org) to permit the latest version.

PR #990, opened 2021-03-23
Upgrade numpy to add support for M1 processors

MattiSG:
#### Technical changes

- Upgrade numpy dependency.

`numpy` can be compiled natively on darwin-arm64 from v1.20.1 on. OpenFisca thus needs to allow this version.
Some additional work is needed for installing on Apple Silicon M1 processors, which is documented [here](https://stackoverflow.com/questions/66768259/how-to-install-and-import-scipy-numpy-numexpr-and-others-on-m1-apple-silicon/66768260#66768260). I considered updating the README but concluded that the install procedure should be normalised in the coming months, and that it was not worth documenting transitory instructions in the README, and that this PR discussion would be sufficient.

I'm open to requests to include instructions in the README 🙂

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+### 35.3.7 [#990](https://github.com/openfisca/openfisca-core/pull/990)
+
+#### Technical changes
+
+- Update dependencies.
+  - Extend NumPy compatibility to v1.20 to support M1 processors.
+
 ### 35.3.6 [#984](https://github.com/openfisca/openfisca-core/pull/984)
 
 #### Technical changes
--- a/openfisca_core/indexed_enums/enum.py
+++ b/openfisca_core/indexed_enums/enum.py
@@ -7,23 +7,28 @@ import numpy
 
 from openfisca_core.indexed_enums import config, EnumArray
 
+if typing.TYPE_CHECKING:
+    IndexedEnumArray = numpy.object_
+
 
 class Enum(enum.Enum):
     """
-    Enum based on `enum34 <https://pypi.python.org/pypi/enum34/>`_, whose items have an
-    index.
+    Enum based on `enum34 <https://pypi.python.org/pypi/enum34/>`_, whose items
+    have an index.
     """
 
     # Tweak enums to add an index attribute to each enum item
     def __init__(self, name: str) -> None:
-        # When the enum item is initialized, self._member_names_ contains the names of
-        # the previously initialized items, so its length is the index of this item.
+        # When the enum item is initialized, self._member_names_ contains the
+        # names of the previously initialized items, so its length is the index
+        # of this item.
         self.index = len(self._member_names_)
 
     # Bypass the slow Enum.__eq__
     __eq__ = object.__eq__
 
-    # In Python 3, __hash__ must be defined if __eq__ is defined to stay hashable.
+    # In Python 3, __hash__ must be defined if __eq__ is defined to stay
+    # hashable.
     __hash__ = object.__hash__
 
     @classmethod
@@ -31,16 +36,18 @@ class Enum(enum.Enum):
             cls,
             array: typing.Union[
                 EnumArray,
-                numpy.ndarray[int],
-                numpy.ndarray[str],
-                numpy.ndarray[Enum],
+                numpy.int_,
+                numpy.float_,
+                IndexedEnumArray,
                 ],
             ) -> EnumArray:
         """
-        Encode a string numpy array, an enum item numpy array, or an int numpy array
-        into an :any:`EnumArray`. See :any:`EnumArray.decode` for decoding.
+        Encode a string numpy array, an enum item numpy array, or an int numpy
+        array into an :any:`EnumArray`. See :any:`EnumArray.decode` for
+        decoding.
 
-        :param ndarray array: Array of string identifiers, or of enum items, to encode.
+        :param ndarray array: Array of string identifiers, or of enum items, to
+                              encode.
 
         :returns: An :any:`EnumArray` encoding the input array values.
         :rtype: :any:`EnumArray`
@@ -59,24 +66,31 @@ class Enum(enum.Enum):
         >>> encoded_array[0]
         2  # Encoded value
         """
-        if type(array) is EnumArray:
+        if isinstance(array, EnumArray):
             return array
 
-        if array.dtype.kind in {'U', 'S'}:  # String array
+        # String array
+        if isinstance(array, numpy.ndarray) and \
+                array.dtype.kind in {'U', 'S'}:
             array = numpy.select(
                 [array == item.name for item in cls],
                 [item.index for item in cls],
                 ).astype(config.ENUM_ARRAY_DTYPE)
 
-        elif array.dtype.kind == 'O':  # Enum items arrays
+        # Enum items arrays
+        elif isinstance(array, numpy.ndarray) and \
+                array.dtype.kind == 'O':
             # Ensure we are comparing the comparable. The problem this fixes:
             # On entering this method "cls" will generally come from
-            # variable.possible_values, while the array values may come from directly
-            # importing a module containing an Enum class. However, variables (and
-            # hence their possible_values) are loaded by a call to load_module, which
-            # gives them a different identity from the ones imported in the usual way.
-            # So, instead of relying on the "cls" passed in, we use only its name to
-            # check that the values in the array, if non-empty, are of the right type.
+            # variable.possible_values, while the array values may come from
+            # directly importing a module containing an Enum class. However,
+            # variables (and hence their possible_values) are loaded by a call
+            # to load_module, which gives them a different identity from the
+            # ones imported in the usual way.
+            #
+            # So, instead of relying on the "cls" passed in, we use only its
+            # name to check that the values in the array, if non-empty, are of
+            # the right type.
             if len(array) > 0 and cls.__name__ is array[0].__class__.__name__:
                 cls = array[0].__class__
 
--- a/openfisca_core/indexed_enums/enum_array.py
+++ b/openfisca_core/indexed_enums/enum_array.py
@@ -7,6 +7,8 @@ import numpy
 if typing.TYPE_CHECKING:
     from openfisca_core.indexed_enums import Enum
 
+    IndexedEnumArray = numpy.object_
+
 
 class EnumArray(numpy.ndarray):
     """
@@ -20,7 +22,7 @@ class EnumArray(numpy.ndarray):
     # https://docs.scipy.org/doc/numpy-1.13.0/user/basics.subclassing.html#slightly-more-realistic-example-attribute-added-to-existing-array.
     def __new__(
             cls,
-            input_array: numpy.ndarray[int],
+            input_array: numpy.int_,
             possible_values: typing.Optional[typing.Type[Enum]] = None,
             ) -> EnumArray:
         obj = numpy.asarray(input_array).view(cls)
@@ -28,15 +30,15 @@ class EnumArray(numpy.ndarray):
         return obj
 
     # See previous comment
-    def __array_finalize__(self, obj: typing.Optional[numpy.ndarray[int]]) -> None:
+    def __array_finalize__(self, obj: typing.Optional[numpy.int_]) -> None:
         if obj is None:
             return
 
         self.possible_values = getattr(obj, "possible_values", None)
 
     def __eq__(self, other: typing.Any) -> bool:
-        # When comparing to an item of self.possible_values, use the item index to
-        # speed up the comparison.
+        # When comparing to an item of self.possible_values, use the item index
+        # to speed up the comparison.
         if other.__class__.__name__ is self.possible_values.__name__:
             # Use view(ndarray) so that the result is a classic ndarray, not an
             # EnumArray.
@@ -49,8 +51,8 @@ class EnumArray(numpy.ndarray):
 
     def _forbidden_operation(self, other: typing.Any) -> typing.NoReturn:
         raise TypeError(
-            "Forbidden operation. The only operations allowed on EnumArrays are "
-            "'==' and '!='.",
+            "Forbidden operation. The only operations allowed on EnumArrays "
+            "are '==' and '!='.",
             )
 
     __add__ = _forbidden_operation
@@ -62,7 +64,7 @@ class EnumArray(numpy.ndarray):
     __and__ = _forbidden_operation
     __or__ = _forbidden_operation
 
-    def decode(self) -> numpy.ndarray[Enum]:
+    def decode(self) -> IndexedEnumArray:
         """
         Return the array of enum items corresponding to self.
 
@@ -72,14 +74,16 @@ class EnumArray(numpy.ndarray):
         >>> enum_array[0]
         >>> 2  # Encoded value
         >>> enum_array.decode()[0]
-        <HousingOccupancyStatus.free_lodger: 'Free lodger'>  # Decoded value : enum item
+        <HousingOccupancyStatus.free_lodger: 'Free lodger'>
+
+        Decoded value: enum item
         """
         return numpy.select(
             [self == item.index for item in self.possible_values],
             list(self.possible_values),
             )
 
-    def decode_to_str(self) -> numpy.ndarray[str]:
+    def decode_to_str(self) -> numpy.str_:
         """
         Return the array of string identifiers corresponding to self.
 
--- a/openfisca_core/taxscales/abstract_rate_tax_scale.py
+++ b/openfisca_core/taxscales/abstract_rate_tax_scale.py
@@ -3,28 +3,36 @@ from __future__ import annotations
 import typing
 import warnings
 
-import numpy
-
 from openfisca_core.taxscales import RateTaxScaleLike
+
+if typing.TYPE_CHECKING:
+    import numpy
+
+    NumericalArray = typing.Union[numpy.int_, numpy.float_]
 
 
 class AbstractRateTaxScale(RateTaxScaleLike):
     """
-    Base class for various types of rate-based tax scales: marginal rate, linear
-    average rate...
+    Base class for various types of rate-based tax scales: marginal rate,
+    linear average rate...
     """
 
-    def __init__(self, name: typing.Optional[str] = None, option = None, unit = None) -> None:
+    def __init__(
+            self, name: typing.Optional[str] = None,
+            option: typing.Any = None,
+            unit: typing.Any = None,
+            ) -> None:
         message = [
-            "The 'AbstractRateTaxScale' class has been deprecated since version",
-            "34.7.0, and will be removed in the future.",
+            "The 'AbstractRateTaxScale' class has been deprecated since",
+            "version 34.7.0, and will be removed in the future.",
             ]
+
         warnings.warn(" ".join(message), DeprecationWarning)
-        super(AbstractRateTaxScale, self).__init__(name, option, unit)
+        super().__init__(name, option, unit)
 
     def calc(
             self,
-            tax_base: typing.Union[numpy.ndarray[int], numpy.ndarray[float]],
+            tax_base: NumericalArray,
             right: bool,
             ) -> typing.NoReturn:
         raise NotImplementedError(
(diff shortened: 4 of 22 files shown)


MattiSG:
I need help with the `error: "ndarray" expects no type arguments, but 1 given` errors given in CI. I tried to look up [Numpy types](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.html), but the expected usage in OpenFisca is not clear to me.

maukoquiroga:
Hi @MattiSG ! Thanks for this PR.

> I need help with the error: "ndarray" expects no type arguments, but 1 given errors given in CI. I tried to look up Numpy types, but the expected usage in OpenFisca is not clear to me.

The purpose of those type annotations in general is to enforce the contracts of OpenFisca internals, and in particular with Numpy to provide an idea of the type of data we're expecting —an array of ```int```, ```float```, and so on.

As Numpy provided no type annotations prior 1.20.x, those annotations were not useful programatically: MyPy just defaults them to ```Any```.

With 1.20.x, taking a look at the documentation, Numpy now provides the generic [```ArrayLike```](https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike) via the ```numpy.typing``` module.

The current type annotations being invalid, for this update we would have to either remove them or use the new valid one —```ArrayLike```.

(Ideally if we remove them we could improve the documentation of the expected parameters of those functions and methods, so to keep their informative value).

There's just one caveat regarding the latter, from [Numpy's documentation](https://numpy.org/doc/stable/release/1.20.0-notes.html):

> Some of the types in this module rely on features only present in the standard library in Python 3.8 and greater. If you want to use these types in earlier versions of Python, you should install the typing-extensions package.

Given those type annotations are not actually programmatically linted, I'd say the former would be better as we would have to eventually deprecate Python 3.7 support without justifiable reason.

Besides, because of the amount of [deprecations listed in the release note](https://numpy.org/doc/stable/release/1.20.0-notes.html) I think we should take this update carefully.

In fact, last Numpy update took several pull requests spread among different OpenFisca packages in order to deal with the deprecations and make them compatible:

- https://github.com/openfisca/openfisca-core/issues/924
- https://github.com/openfisca/openfisca-france/pull/1431

I think we should definitely do not lag with this update, and a quick assessment of the impact of the deprecations would be much welcome.

Have you tried it already for example with OpenFisca France?

MattiSG:
Thank you very much @maukoquiroga for this information! 😊

> Have you tried it already for example with OpenFisca France?

Yes, this changeset is the minimal one that bring M1 compatibility, which I backported from a local branch I experimented with over the last two weeks, and matches the setup I currently use and that enabled me to deliver https://github.com/openfisca/openfisca-france/pull/1489 among others.

maukoquiroga:
Thanks @MattiSG, looking at the StackOverflow link you shared, I still have one question: NumPy prior to 1.20 does not work _at all_ on M1 processors?

BTW, I got the tests passing [here](https://github.com/openfisca/openfisca-core/compare/bump-numpy?expand=1).

MattiSG:
> NumPy prior to 1.20 does not work _at all_ on M1 processors?

Nope, can't get it to compile before 1.20.0-rc2, even from source and with options such as no-PEP-517. You can find detailed information on https://github.com/numpy/numpy/issues/17807, and in a more readable format on https://stackoverflow.com/a/65581354.

MattiSG:
> BTW, I got the tests passing [here](https://github.com/openfisca/openfisca-core/compare/bump-numpy?expand=1).

😍 just push on this branch!

README.md
```diff
@@ -26,7 +26,7 @@ If you want to contribute to OpenFisca-Core itself, welcome! To install it local
 ```bash
 git clone https://github.com/openfisca/openfisca-core.git
 cd openfisca-core
-pip install --editable .[dev] --use-deprecated=legacy-resolver
+pip install --editable .[dev]
```
maukoquiroga, line 29: I invite you to propose this change in a separate pull request, in order to have a specific place to discuss about the new dependency resolver, and the best strategy to use it —or not.

Spoiler alert: it can be [way too slow](https://github.com/pypa/pip/issues/9187#issuecomment-806905059), which has proven to be painful in time-sensitive contexts as CI and [test deployments](https://app.netlify.com/sites/openfisca-doc/deploys/6060eb0268d09100073b4b5b).

maukoquiroga: I've just opened this pull request to fix this very same issue on the doc https://github.com/openfisca/openfisca-doc/pull/235

maukoquiroga:
Done! New type hints look very promising on Numpy 1.20 and 1.21 to help refactor OpenFisca Core —as a first step as unit tests are really hard to write today.

@cesco-fran in phase with your draft https://github.com/openfisca/openfisca-core/pull/976.

@MattiSG regarding the bump itself, as this version comprises some [expired deprecations](https://numpy.org/devdocs/release/1.20.0-notes.html#expired-deprecations), I would mark it as a major one.

README.md
MattiSG: This very branch [takes 2 seconds](https://app.circleci.com/pipelines/github/openfisca/openfisca-core/231/workflows/b03f2033-6319-49ba-b822-67bdeb6fb5d5/jobs/6780) to install dependencies in CI. Based on the issue you linked to (https://github.com/pypa/pip/issues/9187), I understand that long builds could happen from a specific set of requirements, not randomly. If I understood correctly, is there anything that makes us suspect our set of requirements could generate such long builds? 🙂

maukoquiroga: You're right, pip tries to download all the intervals, and when it comes to NumPy and the like that adds a [lot of overhead](https://app.circleci.com/pipelines/github/openfisca/openfisca-doc/424/workflows/f4145b16-44e2-4d61-8e76-b5adb9f40615/jobs/1193).

The problem is with the first run, then libraries should normally be saved to cache in CircleCI, per branch —one possible workaround being a stricter version interval (like `==`) but aren't we then defeating the purpose of the tool? It looks like next version of pip is going to adresses this backtracking problem.

I know @sarafalamaki and @sandcha run into problems as well with the new resolver —I don't know if the same problem or others.

maukoquiroga: Just in case my point is not to not use the new resolver but to do so in a new pull request so to document and to address the problems users may have encountered that lead to the actual use of the legacy resolver.

maukoquiroga: BTW found this previous discussion, very interesting for reference https://github.com/openfisca/extension-template/issues/10.

maukoquiroga: > This very branch takes 2 seconds to install dependencies in CI.

That build [is using](https://circleci.com/api/v1.1/project/github/openfisca/openfisca-core/6780/output/104/0?file=true&allocation-id=606105d7a1285115cf0d3ccc-0-build%2F5E874C36) the legacy resolver.

---

I've also found the strategies proposed by the Pip folks https://pip.pypa.io/en/latest/user_guide/#dependency-resolution-backtracking, which leds me to think that, before PyPi provides package's metadata as an API, we will have to find a workaround for this problem.

Good news is there are some dependencies we know as a fact can be pinned, like `numpy`, `numexpr`,  `scipy`, and so on, which could be a solution.

I run the following test —I had to cancel at some point:

##### First attempt

As it is now.

```python
general_requirements = [
    'dpath >= 1.5.0, < 2.0.0',
    'pytest >= 4.4.1, < 6.0.0',  # For openfisca test
    'numpy >= 1.11, < 1.19',
    'psutil >= 5.4.7, < 6.0.0',
    'PyYAML >= 3.10',
    'sortedcontainers == 2.2.2',
    'numexpr >= 2.7.0, <= 3.0',
    ]

api_requirements = [
    'werkzeug >= 1.0.0, < 2.0.0',
    'flask == 1.1.2',
    'flask-cors == 3.0.10',
    'gunicorn >= 20.0.0, < 21.0.0',
    ]

dev_requirements = [
    'autopep8 >= 1.4.0, < 1.6.0',
    'flake8 >= 3.9.0, < 4.0.0',
    'flake8-bugbear >= 19.3.0, < 20.0.0',
    'flake8-print >= 3.1.0, < 4.0.0',
    'pytest-cov >= 2.6.1, < 3.0.0',
    'mypy >= 0.701, < 0.800',
    'openfisca-country-template >= 3.10.0, < 4.0.0',
    'openfisca-extension-template >= 1.2.0rc0, < 2.0.0'
    ] + api_requirements
```

```
pip freeze | grep -v "^-e" | xargs pip uninstall -y
pip cache purge
pip install --upgrade pip
time pip install --editable .[dev] --upgrade

...

ERROR: Operation cancelled by user

real	17m37.289s
user	2m36.219s
sys	0m17.993s
```

##### Second attempt

Tightening dependency versions.

```python
general_requirements = [
    'dpath ~= 1.5.0',
    'pytest ~= 5.4.0',  # For openfisca test
    'numpy == 1.18.5',
    'psutil ~= 5.8.0',
    'PyYAML ~= 5.4.0',
    'sortedcontainers == 2.2.2',
    'numexpr ~= 2.7.0',
    ]

api_requirements = [
    'werkzeug ~= 1.0.0',
    'flask == 1.1.2',
    'flask-cors == 3.0.10',
    'gunicorn ~= 20.1.0',
    ]

dev_requirements = [
    'autopep8 ~= 1.5.0',
    'flake8 ~= 3.9.0',
    'flake8-bugbear ~= 19.8.0',
    'flake8-print ~= 3.1.0',
    'pytest-cov ~= 2.11.0',
    'mypy == 0.790',
    'openfisca-country-template >= 3.10.0, < 4.0.0',
    'openfisca-extension-template >= 1.2.0rc0, < 2.0.0',
    ] + api_requirements
```

```
pip freeze | grep -v "^-e" | xargs pip uninstall -y
pip cache purge
pip install --upgrade pip
time pip install --editable .[dev] --upgrade

...

ERROR: Operation cancelled by user

real	7m22.893s
user	6m25.452s
sys	0m2.349s
```

### Third attempt

As it is but pre-calculating some dependencies with `pip-tools`.

It works! I mean, it fails, but the resolver works: we have an incompatible circular dependency.

Even if we reduce the time it take pip to resolve to a couple of minutes, we have core depending on country-template, and country-template depending on core, so it'll inevitably fail.

```
pip freeze | grep -v "^-e" | xargs pip uninstall -y
pip cache purge
pip install --upgrade pip
pip install pip-tools
time ( echo "--editable .[dev]" | pip-compile --output-file=- > /tmp/requirements.txt - -qo- && pip install -r /tmp/requirements.txt --upgrade )

...

Requirement already satisfied: OpenFisca-Core[web-api]<36.0.0,>=35.0.0 in /Users/hyperion/Sites/openfisca/openfisca-core (from openfisca-country-template==3.12.5->-r /tmp/requirements.txt (line 65)) (35.3.0)
Collecting OpenFisca-Core[web-api]<36.0.0,>=35.0.0
  Downloading OpenFisca_Core-35.2.0-py3-none-any.whl (162 kB)
     |████████████████████████████████| 162 kB 1.3 MB/s 
  Downloading OpenFisca_Core-35.1.1-py3-none-any.whl (162 kB)
     |████████████████████████████████| 162 kB 1.0 MB/s 
  Downloading OpenFisca_Core-35.1.0-py3-none-any.whl (162 kB)
     |████████████████████████████████| 162 kB 265 kB/s 
  Downloading OpenFisca_Core-35.0.5-py3-none-any.whl (162 kB)
     |████████████████████████████████| 162 kB 2.1 MB/s 
  Downloading OpenFisca_Core-35.0.4-py3-none-any.whl (162 kB)
     |████████████████████████████████| 162 kB 1.2 MB/s 
  Downloading OpenFisca_Core-35.0.3-py3-none-any.whl (162 kB)
     |████████████████████████████████| 162 kB 2.6 MB/s 
  Downloading OpenFisca_Core-35.0.1-py3-none-any.whl (162 kB)
     |████████████████████████████████| 162 kB 1.9 MB/s 
  Downloading OpenFisca_Core-35.0.0-py3-none-any.whl (162 kB)
     |████████████████████████████████| 162 kB 2.0 MB/s 

...

ERROR: Cannot install openfisca-core 35.3.0 (from /Users/hyperion/Sites/openfisca/openfisca-core), openfisca-core[web-api]==35.0.0, openfisca-core[web-api]==35.0.1, openfisca-core[web-api]==35.0.3, openfisca-core[web-api]==35.0.4, openfisca-core[web-api]==35.0.5, openfisca-core[web-api]==35.1.0, openfisca-core[web-api]==35.1.1, openfisca-core[web-api]==35.2.0 and openfisca-core[web-api]==35.3.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested openfisca-core 35.3.0 (from /Users/hyperion/Sites/openfisca/openfisca-core)
    openfisca-core[web-api] 35.3.0 depends on openfisca-core 35.3.0 (Installed)


real	0m49.232s
user	0m23.754s
sys	0m1.487s
...

---

Considering the results of those tests, I'm afraid this change break things for users, not just because of the time overhead but as we've just discovered also because of the circular dependencies.

maukoquiroga: > If I understood correctly, is there anything that makes us suspect our set of requirements could generate such long builds? 🙂
> > You're right, pip tries to download all the intervals, and when it comes to NumPy and the like that adds a lot of overhead.

Well... when I pinned NumPy's version I discovered the resolver backtracks following dependencies dependencies, which presents two problems:

1. Backtracking is exponential.
2. We don't have a have any control over the way our dependencies manage their own dependencies.

The following are the [solutions proposed](https://pip.pypa.io/en/latest/user_guide/#possible-ways-to-reduce-backtracking-occurring) by the Pip folks:

> 1. ALLOW PIP TO COMPLETE ITS BACKTRACKING

Already tried, it doesn't work unless we solve our two major problems (wild dependencies' dependencies and circular dependencies).

> 2. REDUCE THE NUMBER OF VERSIONS PIP WILL TRY TO BACKTRACK THROUGH

Also already tried, cf. supra.

> 3. USE CONSTRAINT FILES OR LOCKFILES

This could be easily done with `pip-tools`, `pipgrip`, `Pipenv`, `Poetry`, etc.

We'd still have to solve the circular dependency problem, though, unless we ignore them —given `pip` allows us to do so, or using `pipgrip` as an alternative.

> 4. BE MORE STRICT ON PACKAGE DEPENDENCIES DURING DEVELOPMENT

Same as 1. and 2.

😃

MattiSG: Thanks a lot for this thorough and documented exploration of the problem space @maukoquiroga! 🙏😃

MattiSG:
> as this version comprises some [expired deprecations](https://numpy.org/devdocs/release/1.20.0-notes.html#expired-deprecations), I would mark it as a major one.

Indeed, these elements are removed from the public NumPy API: `Complex64`, `Complex32`, `sctypeNA`, `typeNA` and `ctypeslib.ctypes_load_library`. However, they are not used in the codebase. Did Core expose any of them? 🤔

maukoquiroga:
Hello! 

I was thinking more on the libraries depending on core: as some use NumPy directly and extensively (I'm particularly thinking on OpenFisca France, OpenFisca France Data, LexImpact Server...) an upgrade of NumPy in core will force them to upgrade NumPy as well.

However, I haven't checked whether those libraries rely in their codebase on the expired deprecations you mention.

I know that we do have no contract with the depending libraries in terms of the public NumPy API, which could indeed make this just be a minor release.

On the other hand, I feel inclined at the same time to warn depending libraries' maintainers of possible but unchecked impacts on their codebase, hence a breaking-change for them (but related to NumPy not to core).

What do you think?

MattiSG:
Mmh I'm not familiar enough with Python packaging rules to understand how much each dependent sharing a dependency is impacted by conflicting versions 🤔 Spontaneously, I would hope that dependents can have their own version of Numpy without relying on ours! Otherwise, I understand that we should have a breaking change for every major library upgrade, including stuff like `flake8`.

MattiSG:
I tried to rebase this branch on top of `master` but after #992 every modified file has disappeared 😭
@maukoquiroga do you feel like re-applying your changes? I can try to do it but it'd be quite literally monkey-patching as in “I have no idea what I'm doing” 😅

maukoquiroga:
> I tried to rebase this branch on top of `master` but after #992 every modified file has disappeared 😭
> @maukoquiroga do you feel like re-applying your changes? I can try to do it but it'd be quite literally monkey-patching as in “I have no idea what I'm doing” 😅

Of course, most of them are commits of my own.

maukoquiroga:
Done! it think this is no longer a draft 😃

maukoquiroga (approved):
LGTM at least for me! If you agree we keep aside the resolver commit the time we find a solution for that (just tried again one last time to no avail...).

MattiSG:
Resolver legacy option removal removed.

I'm going with a minor version bump as I expect that the API surface forwarded from NumPy to country packages by OF-Core has not changed with this version change.

MattiSG:
I made a mistake 😞
The version bump that I made in the code was not a _minor_ version bump but a _patch_ version bump.
 
- **Expected impact**: minimal. Dependents will have to use a specific patch version as minimum if they expect to support M1 processors, instead of a minimum minor version number, which would be more semantic.
- **Worst case impact**: minor. If the API surface is impacted by breaking changes in NumPy, updating the library could break some use cases. However, the intention was already to publish a minor version and not a major one, so the main foreseen impact would be that dependents would have to specify an upper version number lock that would prevent patches from being applied in their older, locked-down version. However, backporting patches is very rare in OpenFisca, which has a very linear deployment process.
- **Avoiding similar mistakes in the future**: (1) re-request a review after version bump; or (2) do not approve PRs that don't include a version bump commit. Either seem hard to achieve in country packages considering the amount of small-scale PRs and associated administrative costs. However, it could be made mandatory in Core, for example by enabling automatic review dismissal after new commits are pushed.

maukoquiroga:
Thanks @MattiSG, I made a mistake as well, as you said, I should have waited for your answer and for the version bump before approving it, or have had a way to chime in and to check your last commits. I think automatic review dismissal is a good idea.

maukoquiroga:
#1004

guillett:
This PR has more negative impact then anticipated. #1012

sandcha:
> Mmh I'm not familiar enough with Python packaging rules to understand how much each dependent sharing a dependency is impacted by conflicting versions 🤔 Spontaneously, I would hope that dependents can have their own version of Numpy without relying on ours! Otherwise, I understand that we should have a breaking change for every major library upgrade, including stuff like `flake8`.

@MattiSG I think that we have a specific situation for `numpy` updates as you can have code written in openfisca syntax that is not in the code base of a country model as it is, for example, the case for the reforms. And we frequently have user scripts in python that create an openfisca simulation and also use numpy to manage the numpy arrays of the simulation results. In those two examples, I think that we need to inform the users that getting an openfisca with a bump on numpy might break their code. And to do so, the only way that I know to communicate is by a major bump on openfisca-core. 
We can also discuss this on #1010 that currently needs a version bump 🙂

maukoquiroga:
To give some historical context, there have been three NumPy bumps recently:

- #758 was a patch, tests were run against openfisca/openfisca-france only
- #954 was a major, tests were run against openfisca/openfisca-france, openfisca/country-template and openfisca/extension-template
- #990 was a patch intended minor, test were run against openfisca/openfisca-france, openfisca/country-template and openfisca/extension-template

And my 2c regarding the situation:

1. The actual problem we have in master is not at all related to NumPy, but MyPy. This problem could have been avoided, if I didn't accept my own code as I now avoided in #1010, if we tested against the versions we declare as being compatible, which we do not do, or not (we cannot just foresee and avoid everything and every single bug). In any case, that fact alone does not justify by itself a major version IMHO.

2. When we consider supporting a *new NumPy version*, or in general a new version of any "general" dependency (for everyone, not just a "dev" one, like `flake8`), we can rightly say, as it was done here, that it is a minor change, given we ensure compatibility of new changes to the core with not just the new added version but at least the older one supported (in this case 1.11.0 and 1.20.0).

@MattiSG said justly:

> I would hope that dependents can have their own version of Numpy without relying on ours!

They do rely on ours, but on all the range of them, not just the latest. Upgrading NumPy does not force anyone to use that version, unless we force them to do it —which we could— like in #1012.

However, the fact that we're not currently testing that they can actually continue to use the version they're already using, makes it impossible to ourselves to know whether we're introducing a breaking-change or not.

In that specific case, I think we should not defer the responsibility of knowing so to the users, knowing how to avoid it —we're literally saying "we mark it as a breaking change because, as we can know if it is, but we haven't tested so, you will have to do it by yourself".

3. When we consider *dropping support of an older NumPy version**, it's the exact opposite, and I 100% with @sandcha's comment supra: we do absolutely not know how that will impact users, and contrary to a version bump I can just ignore, here I'm being forced to upgrade. In that case I'd systematically assume a breaking change.

PR #992, opened 2021-04-01
Render the library more modular for contributors' goodness ✨

maukoquiroga:
@openfisca/international-contrib what do you think?

#### Technical improvements

- Render all OpenFisca Core components modular and sandboxed
  - Allows for simpler contribution
  - Allows for better unit testing
  - Allows for dealing with circular dependencies
  - Allows for more explicit contracts between components

#### Deprecations

- Reorganise `formula_helpers`, `memory_config`, `rates`, `simulation_builder`.
  - Their functionalities are still there, just moved around
  - Transitional `__init__.py` files added to make transition smooth


augusto-herrmann:
I think this proposal seems to make sense, especially considering, as you mentioned, it would avoid cyclic dependency problems.

Unfortunately I can't say anything from a practical point of view, as I am not really using OpenFIsca yet.

CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
 # Changelog
 
+## 35.4.0 [#992](https://github.com/openfisca/openfisca-core/pull/992)
+
+#### New features
```
fpagnoux, line 5: > I'm a little openfisca rusty, but are these really "new features"? Looks more like internal tech debt removal without consequences for users?

fpagnoux (approved):
900 lines of code per file... phew 😅 

Assuming this PR only moves things around without changing any functionality (i didn't look in details), I think leaning towards smaller and more modular files is a great idea. Thanks for the busy work!

CHANGELOG.md
maukoquiroga: You're right, there are no technically new features.

maukoquiroga:
> 900 lines of code per file... phew 😅
> 
> Assuming this PR only moves things around without changing any functionality (i didn't look in details), I think leaning towards smaller and more modular files is a great idea. Thanks for the busy work!

Yes! I took care of keeping exposed all the same functions, classes, and modules, with the original names. I didn't expose protected functions however (the one of the ones ```_like(this):```) but if people are relying on them we can just make them public.

That's basically why I didn't touch the tests at all.

MattiSG:
Wow, that's impressive! Fingers crossed that such a large changeset passes everywhere 🤞 If it all goes smoothly, I expect this to improve contributors’ navigation in the codebase 👏

maukoquiroga:
Thanks @MattiSG ! I plan to broadcast this and reach as many users as possible for help and feedback on possible issues. And yes I think this is a big step on making it easier for contributors to navigate and to contribute :)

PR #993, opened 2021-04-01
Gracefully handle unexpected errors

### LINKED ISSUE #991: Bug // web_api does not handle `ParameterNotFound` error gracefully. Instead it returns no response.
### Here is what I did:

- I have discovered a bug while calling the `/calculate` endpoint. It occurs when you provide an invalid "period" value, but the web_api does not return a meaningful error response. 
- I've created a failing test [in this branch](https://github.com/openfisca/openfisca-core/tree/bug-webapi-parameter-indexerror)

- I expect the web_api to return an error response (400 error, or even 500 error)
- I expect a meaningful error message indicating the "period" provided is invalid based on the parameter values specified.

- The web_api returns no response. Instead it halts due to the IndexError.
- Upon receiving no response, Chrome assumes it is a CORS error. The screenshot below shows the frontend error.
image

- Here is the backend stack trace:
image

## Context

I identify more as a:

- Developer - I create frontend applications which ingest OpenFisca rulesets 

Thanks in advance!

maukoquiroga:
Fixes #991

#### Bug fix

- [Web API] Gracefully handle unexpected errors 
  - The exception signature expected by the internal server error handler was not the good one
  - Henceforth no response was being given to the user, when a 500 with an explanation was expected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+### 35.3.1 [#993](https://github.com/openfisca/openfisca-core/pull/993)
+
+#### Bug fix
+
+- [Web API] Gracefully handle unexpected errors 
+  - The exception signature expected by the internal server error handler was not the good one
+  - Henceforth no response was being given to the user, when a 500 with an explanation was expected
+
 ## 35.3.0 [#985](https://github.com/openfisca/openfisca-core/pull/985)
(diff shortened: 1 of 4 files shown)


RamParameswaran (approved):
All looks good to me. Approved :+1: 
Thanks @maukoquiroga

PR #994, opened 2021-04-08
In SimulationBuilder, repair expansion of axes on a variable given as input

eraviart:
#### Bug fix

- Repair expansion of axes on a variable given as input
  - When expanding axes, the expected behavour is to override any input value for the requested variable and period
  - As longs as we passed some input for a variable on a period, it was not being overrode, creating a NumPy's error (boradcasting)
  - By additionally checking that an input was given, now we make that the array has the correct shape by constructing it with NumPy's tile with a shape equal to the number of the axis expansion count requested.

---

Updated by @maukoquiroga the 10/04/2021.

### DIFF
--- a/tests/core/test_axes.py
+++ b/tests/core/test_axes.py
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
+import pytest
 from pytest import fixture, approx
 
 from openfisca_core.simulation_builder import SimulationBuilder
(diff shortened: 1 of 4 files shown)


tests/core/test_axes.py
```diff
@@ -131,6 +131,17 @@ def test_add_perpendicular_axes(simulation_builder, persons):
     assert simulation_builder.get_input('salary', '2018-11') == approx([0, 1500, 3000, 0, 1500, 3000])
     assert simulation_builder.get_input('pension', '2018-11') == approx([0, 0, 0, 2000, 2000, 2000])
 
+
+def test_axis_on_existing_variable(simulation_builder, persons):
```
maukoquiroga, line 135: Is it just on an existing variable or in a variable with an already defined input for that period?

maukoquiroga (approved):
✨ Thanks! I've just added an additional test and corrected some minor typos.

tests/core/test_axes.py
maukoquiroga: Found my answer tinkering with the code: it fails when there is already and existing input for a variable of the requested period.

The reason it was failing is sadly not the good one: if we were to give as an expansion count `1`, it wont expand at all —which makes sense, but maybe it is not the least surprising behaviour.

PR #996, opened 2021-04-11
Extract axes from SimulationBuilder

bonjourmauko:
This is a work in progress based on my willingness to give a proper answer to #933 and #934.

I hoped to have a more advanced proposal for a request for comments, but axes are a hard one to tackle properly.

What do you think? Any feedback at this point @benjello @guillett?

Details below.

#### New features

- Introduce `Axis`, `AxisArray`, and `AxisExpander`.
  - It clearly separates the data logic from the domain logic (setting up a collection of axes coherently vs actually expanding a simulation over a given collection of axes).

#### Improvements

- Introduce proper separation of concerns between an axis, a multi-dimensional axes collection, and its expansion.
  -  Allows for a better readability and testability, and future improvement of the expand axes feature.
  - Adds more unit test coverage, including some untested invariants (like the number of counts for a parallel set of axes).

#### Deprecations

- Deprecate `add_parallel_axis`, `add_perpendicular_axis`, and `expand_axes`.
  - Those functionalities are now provided by `AxisArray` and `AxisExpander`.

### DIFF
--- a/openfisca_core/simulations/__init__.py
+++ b/openfisca_core/simulations/__init__.py
@@ -24,5 +24,6 @@
 from openfisca_core.errors import CycleError, NaNCreationError, SpiralError  # noqa: F401
 
 from .helpers import calculate_output_add, calculate_output_divide, check_type, transform_to_strict_syntax  # noqa: F401
+from .axis import Axis  # noqa: F401
 from .simulation import Simulation  # noqa: F401
 from .simulation_builder import SimulationBuilder  # noqa: F401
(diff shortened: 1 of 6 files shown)


benjello (approved):
I am not an expert but it looks cleaner than before therefore I have no objection.

maukoquiroga:
Thanks @benjello, to be more clear what I would like is to know if the assumptions I'm doing are the right one in terms of field expertise. For example:

- Does it make sense that all axes perpendicular to themselves have the same amount of counts, or steps?
- Does it make sense to have more than 2 dimensions?
- Etc.

Most of the intentent is not yet implemented but outlined in the documentation.

benjello:
> * Does it make sense that all axes perpendicular to themselves have the same amount of counts, or steps?

- No

> * Does it make sense to have more than 2 dimensions? (a 4-dimension axes collection is unplottable, but maybe it makes sense for a use case of yours).

- Yes

maukoquiroga:
Thanks @benjello, this is not yet ready to be merged, I wanted to be sure I was shooting at the right direction.



> > * Does it make sense that all axes perpendicular to themselves have the same amount of counts, or steps?
> 
> * No

So we could have, for example, something like:

```
axes[[2,2], [3], [4]]
```

But no:

```
axes[[5,2], [3], [4]]
```

?

PR #997, opened 2021-04-11
Make CountryTaxBenefitSystem reusable as a fixture

maukoquiroga:
Related to #969 (it is almost a subset of it, it only extracts CountryTaxBenefitSystem to a fixture).
Related to #984 (it adds more modularity but does not evince country template's scenarios).

#### Technical changes

- In tests, extract `CountryTaxBenefitSystem` to a fixture reusable by all the tests in the test suite.
  - It allows for a better reusability of test scenarios available to new contributors.
  - To mitigate possible performance issues, by default the fixture is initialised once per test module.

##### Performance

Impact on performance of the test suite is neglectable. 

Before:

```
openfisca-core[master] > time for i in {1..10}; do pytest; done


real	1m59.024s
user	1m42.614s
sys	0m27.081s
```

After:

```
openfisca-core[extract-tbs-fixture] >  time for i in {1..10}; do pytest; done

real	2m5.148s
user	1m46.425s
sys	0m28.280s
```

### DIFF
--- a/tests/core/test_countries.py
+++ b/tests/core/test_countries.py
@@ -91,34 +91,26 @@ def test_input_with_wrong_period(simulation_builder, tax_benefit_system):
         simulation_builder.build_from_variables(tax_benefit_system, variables)
 
 
-def test_variable_with_reference(simulation_builder, isolated_tax_benefit_system):
+def test_variable_with_reference(make_simulation, isolated_tax_benefit_system):
     variables = {"salary": 4000}
-    simulation_builder.set_default_period(PERIOD)
+    simulation = make_simulation(isolated_tax_benefit_system, variables, PERIOD)
 
-    simulation = \
-        simulation_builder \
-        .build_from_variables(isolated_tax_benefit_system, variables)
+    result = simulation.calculate("disposable_income", PERIOD)
 
-    disposable_income = simulation.calculate("disposable_income", PERIOD)
-
-    assert disposable_income > 0
+    assert result > 0
 
     class disposable_income(Variable):
-
         definition_period = periods.MONTH
 
         def formula(household, period):
             return household.empty_array()
 
     isolated_tax_benefit_system.update_variable(disposable_income)
+    simulation = make_simulation(isolated_tax_benefit_system, variables, PERIOD)
 
-    simulation = \
-        simulation_builder \
-        .build_from_variables(isolated_tax_benefit_system, variables)
+    result = simulation.calculate("disposable_income", PERIOD)
 
-    disposable_income = simulation.calculate("disposable_income", PERIOD)
-
-    assert disposable_income == 0
+    assert result == 0
 
 
 def test_variable_name_conflict(tax_benefit_system):
--- a/tests/core/test_countries.py
+++ b/tests/core/test_countries.py
@@ -1,137 +1,145 @@
-# -*- coding: utf-8 -*-
+import pytest
 
-from pytest import fixture, raises
-
+from openfisca_core import periods, populations, tools
+from openfisca_core.errors import VariableNameConflictError, VariableNotFoundError
 from openfisca_core.variables import Variable
-from openfisca_core.periods import MONTH
-from openfisca_core.simulation_builder import SimulationBuilder
-from openfisca_core.taxbenefitsystems import VariableNameConflict, VariableNotFound
-from openfisca_core import periods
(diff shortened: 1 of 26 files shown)


tests/fixtures/simulations.py
```diff
@@ -0,0 +1,30 @@
+import functools
+
+import pytest
+
+from openfisca_core.simulation_builder import SimulationBuilder
+
+
+@pytest.fixture
+def simulation_builder():
```
RamParameswaran, line 9: I don't have much experience with Simulations, so this might not be a relevant question - but...

Why do we need to create a fixture for `simulation_builder`? It seems to me that `simulation_builder` just returns an newly instantiated "SimulationBuilder" object. Why not instantiate the "SimulationBuilder" object directly in the test, and avoid the confusion of having a fixture?

tests/core/test_entities.py
```diff
@@ -32,22 +28,21 @@
 MONTH = "2016-01"
 
 
-def new_simulation(test_case, period = MONTH):
-    builder = SimulationBuilder()
-    builder.set_default_period(period)
-    return builder.build_from_entities(tax_benefit_system, test_case)
+def new_simulation(simulation_builder, tax_benefit_system, test_case, period = MONTH):
```
RamParameswaran, line 31: Related to my comment in `tests/fixtures/simulations.py` ... 
Why use the 'simulation_builder' fixture here instead of instantiating SimulationBuilder() directly?

The existing code is more readable IMHO.

tests/core/test_countries.py
```diff
@@ -1,137 +1,126 @@
...
-
-    simulation_base = make_isolated_simulation(tax_benefit_system, {'salary': 4000})
-
-    revenu_disponible_avant_reforme = simulation_base.calculate('disposable_income', "2016-01")
-    assert(revenu_disponible_avant_reforme > 0)
+    assert result > 0
 
     class disposable_income(Variable):
-        definition_period = MONTH
+        definition_period = periods.MONTH
```
RamParameswaran, line 103: The import style for periods should (ie. `MONTH` vs `periods.MONTH`) should be consistent across the code base. It seems that most other tests in this PR ([this for example](https://github.com/openfisca/openfisca-core/blob/96a543320a26eaf66f971d563165e431bc727ba6/tests/core/test_formulas.py#L3)) import MONTH directly.

RamParameswaran (commented):
Thanks @maukoquiroga. This is a good forward step towards modularity!

I just have one general question regarding the 'simulation_builder' fixture (see inline comments).

Otherwise, generally looks good - happy to approve :smile:

tests/fixtures/simulations.py
maukoquiroga: I've just copied it over, but it makes sense, yes.

tests/core/test_countries.py
maukoquiroga: Yes, I've tried to correct that here #992 :

- All classes are imported in an explicit way when it is possible to do so (cyclic dependencies)
- All helper functions and variables are imported with their parent module (as here for example)

I didn't want to clutter this pull request with that, so this is probably just an automatism and we could ignore it, but I'd expect all the code base to follow the `periods.MONTH` pattern.

maukoquiroga:
> I just have one general question regarding the 'simulation_builder' fixture (see inline comments).

You're right! I've accepted your suggestion.

tests/core/test_countries.py
maukoquiroga: I will profit from this opportunity to add a first version of a style guideline. That will help, I hope, in subsequent pull requests to be building it progressively. I think this to add a lot of value as for OpenFisca "unwritten" style choices being largely undocumented.

tests/fixtures/simulations.py
maukoquiroga: Changed.

tests/core/test_countries.py
maukoquiroga: Changed.

tests/core/test_entities.py
maukoquiroga: Changed.

RamParameswaran (approved):
LGTM. 
I have no additional concerns since my last review. Thanks for incorporating my comment about the simulation_builder fixture.
And the styleguide is fantastic :boom: 

Nice one @maukoquiroga! :clap:

PR #998, opened 2021-04-13
Feature  single amount tax scale interpolation

RamParameswaran:
#### New features

- Introduce ability to interpolate between values in a SingleAmountTaxScale
- Currently we can return the nearest value (floor or ceiling) between two thresholds, based on an 'input_value' using [SingleAmountTaxScale.calc(input_value)](https://github.com/openfisca/openfisca-core/blob/c1acafa93b03eea6eaaaacc1aa4473df921f6515/openfisca_core/taxscales/single_amount_tax_scale.py#L12)
- This PR allows calculating the linear-interpolated value between two thresholds.

  - Behind the scenes it uses [the numpy.interp() function](https://numpy.org/doc/stable/reference/generated/numpy.interp.html)
  - I've passed through the `right=` and `left=` keyword arguments for user flexibility

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+### 35.4.0 [#994](https://github.com/openfisca/openfisca-core/pull/994)
+
+#### New features
+
+- Introduce ability to interpolate between values in a SingleAmountTaxScale
+
+#### Usage notes
+
+- Interpolation is performed within the `SingleAmountTaxScale.calc()` function if the keyword argument `interpolate=True` is provided
+
 ### 35.3.3 [#994](https://github.com/openfisca/openfisca-core/pull/994)
 
 #### Bug fix
--- a/openfisca_core/taxscales/single_amount_tax_scale.py
+++ b/openfisca_core/taxscales/single_amount_tax_scale.py
@@ -13,12 +13,25 @@ class SingleAmountTaxScale(AmountTaxScaleLike):
             self,
             tax_base: typing.Union[numpy.ndarray[int], numpy.ndarray[float]],
             right: bool = False,
+            interpolate: bool = False,
             ) -> numpy.ndarray[float]:
         """
         Matches the input amount to a set of brackets and returns the single cell value
         that fits within that bracket.
+
+        :arguments:
+        - `interpolate` (bool): if True, returns the linear interpolated value
+                                between the nearest two thresholds.
+
         """
-        guarded_thresholds = numpy.array([-numpy.inf] + self.thresholds + [numpy.inf])
-        bracket_indices = numpy.digitize(tax_base, guarded_thresholds, right = right)
-        guarded_amounts = numpy.array([0] + self.amounts + [0])
-        return guarded_amounts[bracket_indices - 1]
+        if interpolate:
+            thresholds = self.thresholds
+            amounts = self.amounts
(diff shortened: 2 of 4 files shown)


benjello:
@RamParameswaran : it looks like the interpolation procedure you are applying is converting a `SingleAmountTaxScale` to a `LinearAverageRateTaxScale`.  Could you confirm ? If it is the cas why about using a converter.

RamParameswaran:
> @RamParameswaran : it looks like the interpolation procedure you are applying is converting a `SingleAmountTaxScale` to a `LinearAverageRateTaxScale`. Could you confirm ? If it is the cas why about using a converter.

@benjello I think the interpolation in this PR is different from a LinearAverageRateTaxScale because it specifically deals with `amounts`, not `rates`. 

In my use case, we want to specify the parameters yaml as `[threshold, amount]` pairs. We probably _could_ use a LinearAverageRateTaxScale, but we'd have to transform the parameter data so much, that it would not be easily traceable back to the legislation. I think this PR is a much cleaner approach.

```
# An example: 
thresholds = [0,2,4,6,8,10]    # x-values
values = [0,10,30,60,100,50]   # y-values

# We want to find y values for x=3, x=7, and x=9, using piecewise linear interpolation
y(3) ==> 20
y(7) ==> 80
y(9) ==> 75
```

benjello:
@RamParameswaran : yes you will have to compute the rates but I think that my suggestion is cleaner since in the end what you get _is_ actually a `LinearAverageRateTaxScale`.

maukoquiroga:
Thanks @benjello ! And @RamParameswaran thanks for this contribution. One question: why not, instead of adding another responsibility to the `SingleAmountTaxScale`, create another amount scale responsible for just this use case?

maukoquiroga:
So for example we would have:

```
class LinearAverageAmountTaxScale(AmountTaxScaleLike):
   pass
```

Then we will be able to rely on polymorphism —multiple dispatch— to decide which scale to instantiate depending on the requested simulation parameters.

benjello:
@maukoquiroga : you will add another object that behaves like an existing one ... 
But I won't die on this hill and I can live with @RamParameswaran solution but I would definitely like to have a converter
to enrich the `TaxScale` ecosystem.

RamParameswaran:
> you will add another object that behaves like an existing one ...

Agreed - for this reason, I do not think creating another TaxScale is appropriate.

@benjello I see your point about it effectively becoming a LinearAverageRateTaxScale. I'll try my hand at creating a converter tomorrow, and follow up :+1:

RamParameswaran:
Hi @benjello and @maukoquiroga 

I've written a converter function ([see branch: 'add-converter-SingAmountTaxScale')](https://github.com/openfisca/openfisca-core/compare/add-converter-SingAmountTaxScale). I'm not sure how useful it is, or indeed what it even means to convert a SingleAmountTaxScale to a LinearAverageRateTaxScale. What are the use-cases? :confused: Would you mind doing a sanity-check @benjello? 

Nonetheless - I am still in favour of the approach of this current PR (998). I think that conceptualising what it means to convert a SingleAmountTaxScale into a LinearAverageRateTaxScale requires quite a lot of mental abstraction. I think it is unreasonable to expect users to do this, just to perform simple arithmetic like linear interpolation between two thresholds. But I appreciate that this comment is rather philosophical :sweat_smile: - keen to hear your thoughts! Thanks! :smile:

PR #999, opened 2021-04-13
Change log.warning to warnings.warn to allow users of the package to hide them.

### LINKED ISSUE #995: Use Python warnings package instead of logging.warning
Hi there!

I'm working at the French Parliament (Assemblée Nationale) in the LexImpact team who help members of the parliament to evaluate the impact of laws projects. We use OpenFisca to compute the income tax.

We compute  income taxes for the whole population. Doing so, we see many of these warnings:

```
WARNING:openfisca_core.simulations:Intermediate results will be stored on disk in /tmp/openfisca_c1rkeofp in case of memory overflow. You should remove this directory once you're done with your simulation.

WARNING:openfisca_core.memory_config:Memory configuration is a feature that is still currently under experimentation. You are very welcome to use it and send us precious feedback, but keep in mind that the way it is used might change without any major version bump.
```

I'd like to be able to hide these warnings.

I could make a PR to use warnings package like this:

```
class OpenFiscaMemoryWarning(UserWarning):
...     pass
... 
... warnings.warn('Memory configuration is a feature that is still currently under experimentation.', OpenFiscaMemoryWarning)
```
So I could hide the warnings like this:
`warnings.simplefilter(action="ignore", category=OpenFiscaMemoryWarning)`

What do you think of this proposition ?

benoit-cty:
Fixes #995 

#### Technical changes

- Change logging.warning to warnings.warn to allow users of the package to hide them. See https://github.com/openfisca/openfisca-core/issues/995 for more info.

So the users could hide the warnings like this:

```python
warnings.simplefilter(action="ignore", category=TempfileWarning)
```

### DIFF
--- a/openfisca_core/experimental/memory_config.py
+++ b/openfisca_core/experimental/memory_config.py
@@ -1,5 +1,6 @@
 import warnings
-from openfisca_core.warnings import OpenFiscaMemoryWarning
+
+from openfisca_core.warnings import MemoryConfigWarning
 
 
 class MemoryConfig:
@@ -13,7 +14,7 @@ class MemoryConfig:
             "You are very welcome to use it and send us precious feedback,",
             "but keep in mind that the way it is used might change without any major version bump."
             ]
-        warnings.warn(" ".join(message), OpenFiscaMemoryWarning)
+        warnings.warn(" ".join(message), MemoryConfigWarning)
 
         self.max_memory_occupation = float(max_memory_occupation)
         if self.max_memory_occupation > 1:
--- a/openfisca_core/experimental/memory_config.py
+++ b/openfisca_core/experimental/memory_config.py
@@ -1,6 +1,12 @@
 import logging
+import warnings
 
 log = logging.getLogger(__name__)
+
+
+class OpenFiscaMemoryWarning(UserWarning):
+    # Custom Warning for MemoryConfig
+    pass
 
 
 class MemoryConfig:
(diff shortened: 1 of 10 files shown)


benoit-cty:
To view the warnings with _make test_, remove _--disable-pytest-warnings_ from _setup.cfg_

openfisca_core/experimental/memory_config.py
```diff
@@ -1,15 +1,26 @@
 import logging
+import warnings
 
 log = logging.getLogger(__name__)
```
maukoquiroga, line 4: Do we still need this one?

openfisca_core/experimental/memory_config.py
```diff
@@ -1,15 +1,26 @@
...
 
 
+class OpenFiscaMemoryWarning(UserWarning):
+    # Custom Warning for MemoryConfig
```
maukoquiroga, line 8:
```suggestion
    """
    Custom warning for MemoryConfig.
    """
```

openfisca_core/parameters/config.py
```diff
@@ -1,15 +1,27 @@
 import logging
```
maukoquiroga, line 1: Same: do we still need one here?

openfisca_core/parameters/config.py
```diff
@@ -1,15 +1,27 @@
...
+import warnings
 import os
 import yaml
 import typing
 
+
+class OpenFiscaLibyamlWarning(UserWarning):
+    # Custom Warning for libyaml not installed
```
maukoquiroga, line 9:
```suggestion
    """
    Custom warning for libyaml not installed.
    """
```

maukoquiroga (commented):
It looks great! Thanks @benoit-cty ✨ 

I've left you a couple of questions otherwise LGTM for me.

One additional thing: wouldn't it make sense to move the warnings to a specific warning folder in the root/openfisca_core, as is the case with errors? (not blocking).

benoit-cty:
Thanks, I just push a modification to put all warnings in a new package _openfisca_core.warnings_

openfisca_core/experimental/memory_config.py
```diff
@@ -1,6 +1,5 @@
-import logging
-
-log = logging.getLogger(__name__)
+import warnings
+from openfisca_core.warnings import OpenFiscaMemoryWarning
```
maukoquiroga, line 2: I just forgot this one: I guess the prefixing here "OpenFisca" is redundant.

maukoquiroga (approved):
Thanks!

PR #1000, opened 2021-04-13
Scale from brackets

benjello:
#### New features

- ParameterScale can now be created from brackets


openfisca_core/parameters/parameter_scale.py
```diff
@@ -21,20 +21,27 @@ class ParameterScale(AtInstantLike):
...
+
+        if brackets:
+            self.brackets = brackets
+
+        helpers._validate_parameter(self, data, data_type = dict, allowed_keys = self._allowed_keys)
         helpers._set_backward_compatibility_metadata(self, data)
         self.metadata.update(data.get('metadata', {}))
 
+        if brackets:
+            return
```
maukoquiroga, line 43: @benjello How does this new behaviour relates to the one just below?

```python
        if not isinstance(data.get('brackets', []), list):
            raise ParameterParsingError(
                "Property 'brackets' of scale '{}' must be of type array."
                .format(self.name),
                self.file_path
                )

        brackets = []
        for i, bracket_data in enumerate(data.get('brackets', [])):
            bracket_name = helpers._compose_name(name, item_name = i)
            bracket = parameters.ParameterScaleBracket(name = bracket_name, data = bracket_data, file_path = file_path)
            brackets.append(bracket)
        self.brackets: typing.List[parameters.ParameterScaleBracket] = brackets
```

I;m just wondering whether the feature is already implemented but not used...

maukoquiroga (commented):
Thanks @benjello for this proposal. I've left you a comment as I'm wondering whether this is actually already implemented but not used. The test is very clear about the intentions 😃

benjello:
@maukoquiroga : I created a new branch https://github.com/openfisca/openfisca-core/pull/1001 implementing a faling test.

benjello:
@maukoquiroga : could you give a look at this PR again. I would like to have it in core ASAP. Thanks !

maukoquiroga:
Hi @benjello ! I've left you a question, I have the impression that part of this feature is already implemented. Could you please take a look quickly to see if it is the case?

benjello:
@maukoquiroga I just responded to your question by writing a failing test using the approach you suggested in #1000.

maukoquiroga:
Ah yes in #1001! I'll take a look thanks 🙏

maukoquiroga:
Thanks for the @benjello it is I guess clear for me know. Please correct me.

So, basically, what works today is creating a `ParameterScale` from bracket's **input data** (a `dict`):

```python
@pytest.fixture
def bracket():
    bracket = """
        rate:
            2017-01-01: 0.02
            2015-01-01: 0.04
            2013-01-01: 0.03
        threshold:
            2013-01-01: 0.0
        """

    return config.yaml.load(bracket, Loader = config.Loader)


def test_build_scale_from_input_scale_brackets(bracket):
    my_scale = ParameterScale(
        name = "MyScale",
        data = {"brackets": [bracket]},
        file_path = None,
        )

    my_bracket = ParameterScaleBracket(
        name = "MyBracket",
        data = bracket,
        )

    assert my_bracket not in my_scale.brackets
    assert my_bracket != my_scale.brackets[0]
    assert str(my_bracket) == str(my_scale.brackets[0])
```

The feature you propose is to create a `ParameterScale` from a bracket **instance** (a `ParameterScaleBracket`)!

Which won't work as you already discovered because of this line:

```python
helpers._validate_parameter(self, data, data_type = dict, allowed_keys = self._allowed_keys)
```

Because we're passing already a `ParameterScaleBracket` instead of a `dict`.

I profit to mention that the error thrown is misleading:

```
openfisca_core.errors.parameter_parsing_error.ParameterParsingError: 'MyScale[0]' must be of type object.
```

It _is_ an object.

benjello:
@maukoquiroga : yes you got it !

maukoquiroga:
So, as I suspected, the feature already exists 😃 . The problem relies within the validator, who fails to realise that for the purposes of creating a scale, a bracket should be considered as a dict, whether it is a dict or not.

Please see the code I've just pushed to make my point (it is not production ready, just a demonstration).

What I would propose you is to help you fix the validator problem, that'll allow you to do what you intend.

benjello:
@maukoquiroga : I got your point. I tried your solution but it didn't work. I added a test to pinpoint where is it dysfunctional.

maukoquiroga:
Thanks @benjello ! I think we're making some progress here.

Taking a look at the error, it fails here:

```
>       assert scale.brackets[0].rate('2017-01-01') == 0.2
E       AttributeError: 'ParameterScaleBracket' object has no attribute 'rate'
```

I'll investigate that further.

benjello:
Yes @maukoquiroga . I think that the problem is [here](https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/parameters/parameter_scale.py#L46-L50) and  I do not know how to solve that.

benjello:
The initialisation of the scale uses dict data to recreate brackets.

maukoquiroga:
So I pushed my point to the verge: treat a bracket as a dict-like object, which seemed pythonic to me. But, unexpected behaviour arose that makes me less confident of that route.

(Not quite so, I still think it is the _correct_ thing to do, but looks more complicated than I though, and I'm not confident of being able to integrate a solution in that sense because of the high risk of possible undetected and hard to debug bugs) 😕

Being pragmatic, the feature you request _should_ already work, but it doesn't, and the implementation you proposed pollutes an already complex signature method.

In that light, a correct compromise to me would be to extract, for now, this to a class method (so we'll have building from a yaml file, the default, and from a bracket object).

Would that work for you @benjello? Tell me what do you think.

benjello:
Yep

benjello:
I am closing this since I do not need it anymore (changing the brackets upfront).

Issue #1004, opened 2021-04-15
Have a clear process for maintenance/contribution


bonjourmauko:
Hi there!

I really enjoy OpenFisca, but I recently encountered an issue.

### Here is what I did:

I proposed a pull request.

### Here is what I expected to happen:

To have an answer.

### Here is what actually happened:

Nothing.

### Here is data (or links to it) that can help you reproduce this issue:

https://github.com/openfisca/openfisca-core/pulls

Something like https://numpy.org/doc/stable/dev/governance/governance.html

## Context

I identify more as a:

- Business expert _(I create tests and model legislation)_.
- Developer _(I create tools that use the existing OpenFisca code)_.
- Lobbyist _(I model reforms to make them exist)_.

MattiSG:
I'm not sure I really understand the content of this issue, which seems to be only the default issue template…? 🤔

maukoquiroga:
Thanks @MattiSG 😱 ! Clearly description was not saved... I'll fix it soon.

bonjourmauko:
I'm closing in favour of other emergent RFCs since then.

PR #1014, opened 2021-04-29
Drop latest NumPy supported version to 1.18.x

### LINKED ISSUE #1009: OpenFisca-Core seems to depend on numpy 1.20

maukoquiroga:
Fixes #1009
Supersedes  #1012
Depended on by #1010

#### Bug fix

- Drop latest NumPy supported version to 1.18.x
  - OpenFisca relies on MyPy for optional duck & static type checking
  - When libraries do not implement their own types, MyPy provides stubs, or type sheds
  - Thanks to `__future__.annotations`, those stubs or type sheds are casted to `typing.Any`
  - Since 1.20.x, NumPy now provides their own type definitions
  - The introduction of NumPy 1.20.x in #990 caused one major problem: 
    - It is general practice to do not import at runtime modules only used for typing purposes, thanks to the `typing.TYPE_CHEKING` variable
    - The new `numpy.typing` module was being imported at runtime, rendering OpenFisca unusable to all users depending on previous versions of NumPy (1.20.x-)
  - These changes revert #990 and solve #1009 and #1012

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,5 @@
 # Changelog
 
-### 35.3.7 [#990](https://github.com/openfisca/openfisca-core/pull/990)
-
-#### Technical changes
-
-- Update dependencies.
-  - Extend NumPy compatibility to v1.20 to support M1 processors.
-
 ### 35.3.6 [#984](https://github.com/openfisca/openfisca-core/pull/984)
 
 #### Technical changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,28 @@
 # Changelog
 
+### 35.3.8 [#1014](https://github.com/openfisca/openfisca-core/pull/1014)
+
+#### Bug fix
+
+- Drop latest NumPy supported version to 1.18.x
+  - OpenFisca relies on MyPy for optional duck & static type checking
+  - When libraries do not implement their own types, MyPy provides stubs, or type sheds
+  - Thanks to `__future__.annotations`, those stubs or type sheds are casted to `typing.Any`
+  - Since 1.20.x, NumPy now provides their own type definitions
+  - The introduction of NumPy 1.20.x in #990 caused one major problem: 
+    - It is general practice to do not import at runtime modules only used for typing purposes, thanks to the `typing.TYPE_CHEKING` variable
+    - The new `numpy.typing` module was being imported at runtime, rendering OpenFisca unusable to all users depending on previous versions of NumPy (1.20.x-)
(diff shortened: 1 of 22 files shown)


sandcha (approved):
Compared to #990 content and all seems to be reverted.

> I see that this PR isn't the method that we chose with @benjello and @guillett (!= method 2️⃣ on slack) but it was before you analyzed was we did in the past for a similar situation and I agree that this way is cleaner (and doesn't introduce a new way of managing unpublished revisions with its risks).

maukoquiroga:
Thanks a lot @sandcha !

> I see that this PR isn't the method that we chose with @benjello and @guillett (!= method 2️⃣ on slack) but it was before you analysed was we did in the past for a similar situation and I agree that this way is cleaner (and doesn't introduce a new way of managing unpublished revisions with its risks).

Again thanks a lot for your proactiveness to try to find a way to unblock users 😃 

There are a couple of factors that convinced me to propose this way of dealing with the issue, and here instead of Slack:

1. As you mention, it seems that we already have a way of dealing with this situation, but its documentation is largely absent. In general, I think that the whole corpus of previous decisions is an asset we have to take into account —referring to the "jurisprudence" more often and calling "referendums" only in case of dissent as a subsidiary decision-making strategy.

2. In parallel, I feel the boundary between Github's and Slack's role is rather blurry. This proposal is thus motivated by the following hypothetical rule-of-thumb: 

- OpenFisca's Slack is an _éphémère_ space mingling,  exchange, cross-pollination, etc. Being _éphémère_ renders it IMHO unsuitable for documenting decisions.

- OpenFisca's Slack is also _underrepresented_ when it comes to the core, most of the public discussions (23% of total) being held in the context of the country models. Thus I'm not even sure it is a suitable place for product decision-making.

- Regarding that previous point: what works for example with France, where we can today with confidence have in a same call a decent amount of contributor's representation, the same does not hold _yet_ true for the core.

3. Finally I think the fact that we already lack a clear process for maintenance/contribution (#1004) makes it the much more undesirable to further increase information and participation asymmetry: what's the weight of my opinion, yours, @guillett's, @benjello's, etc., regarding the whole of the core's users' and contributors'?

This issue is rather easy because of 1. supra, but others won't be so.

PR #1015, opened 2021-05-01
build: add pyproject.toml, poetry, and tox

bonjourmauko:
#### Technical changes

- Add `pyproject.toml` to the repository.
- Add `poetry` to manage dependencies.
- Add `tox` to manage isolated tests.

#### Deprecations

- `openfisca_core.commons.Dummy` => `openfisca_core.commons.empty_clone`
- `openfisca_core.errors.ParameterNotFound` => `openfisca_core.errors.ParameterNotFoundError`
- `openfisca_core.errors.VariableNameConflict` => `openfisca_core.errors.VariableNameConflictError`
- `openfisca_core.errors.VariableNotFound` => `openfisca_core.errors.VariableNotFoundError`
- `openfisca_core.formula_helpers.py` => `openfisca_core.commons`
- `openfisca_core.memory_config.py` => `openfisca_core.experimental`
- `openfisca_core.parameters.Bracket` => `openfisca_core.errors.ParameterScaleBracket`
- `openfisca_core.parameters.ParameterNotFound` => `openfisca_core.errors.ParameterNotFoundError`
- `openfisca_core.parameters.ParameterParsingError` => `openfisca_core.errors.ParameterParsingError`
- `openfisca_core.parameters.Scale` => `openfisca_core.errors.ParameterScale`
- `openfisca_core.rates` => `openfisca_core.commons`
- `openfisca_core.simulation_builder` => `openfisca_core.simulations`
- `openfisca_core.simulations.CycleError` => `openfisca_core.errors.CycleError`
- `openfisca_core.simulations.NaNCreationError` => `openfisca_core.errors.NaNCreationError`
- `openfisca_core.simulations.SpiralError` => `openfisca_core.errors.SpiralError`
- `openfisca_core.taxbenefitsystems.VariableNameConflict` => `openfisca_core.errors.VariableNameConflictError`
- `openfisca_core.taxbenefitsystems.VariableNotFound` => `openfisca_core.errors.VariableNotFoundError`
- `openfisca_core.taxscales.EmptyArgumentError` => `openfisca_core.errors.EmptyArgumentError`


README.md
```diff
@@ -17,7 +17,7 @@ This package contains the core features of OpenFisca, which are meant to be used
 
 OpenFisca runs on Python 3.7. More recent versions should work, but are not tested.
 
-OpenFisca also relies strongly on NumPy. Last four minor versions should work, but only latest/stable is tested.
+OpenFisca also relies strongly on NumPy. Last four minor versions should work, but only upper and lower bound versions are tested.
```
MattiSG, line 20: Where do these “last four minor” come from? 🤔

```suggestion
OpenFisca also relies strongly on NumPy. Only upper and lower bound versions are tested.
```

requirements/debug
```diff
@@ -0,0 +1,5 @@
+# These are dependencies to help with debug and profiling, so we always want
+# the latest versions.
+
+# Interactive console on steroids (even makes your coffee)!
+ipython
```
MattiSG, line 5: What is this useful for? Is it a suggestion to contributors, or is it actually used by OpenFisca-Core anywhere?

requirements/debug
```diff
@@ -0,0 +1,5 @@
+# These are dependencies to help with debug and profiling, so we always want
+# the latest versions.
+
+# Interactive console on steroids (even makes your coffee)!
```
MattiSG, line 4: I'm not at ease with this tongue-in-cheek comment. If this is added as a dependency, I'd prefer that an actual justification along with a link to the documentation is added as a comment 🙂

requirements/dev
```diff
@@ -0,0 +1,25 @@
+# Please make sure to pin all dependency versions, in order to avoid unwanted
+# functional and integration breaks caused by external code updates.
+#
+# These are dev dependencies, so pin them.
```
MattiSG, line 4: > Unnecessary considering the above comments.

requirements/dev
```diff
@@ -0,0 +1,25 @@
...
+
+# For PyTest test coverage integration.
+pytest-cov == 2.11.1
+
+# For optional duck & static type checking.
+mypy == 0.812
+
+# For testing: parameters, variables, etc.
+openfisca-country-template == 3.12.5
+openfisca-extension-template == 1.3.6
```
MattiSG, line 22: Pinning those seems risky to me, as we are likely to lose the benefits from #1020.

requirements/install
```diff
@@ -0,0 +1,68 @@
+# Please make sure to cap all dependency versions, in order to avoid unwanted
+# functional and integration breaks caused by external code updates.
+#
+# These are general dependencies, so:
+#
+# * the ranges should be broad, and
+# * the number of them kept discreet.
```
MattiSG, line 7:
```suggestion
# * the number of them kept low.
```

requirements/install
```diff
@@ -0,0 +1,68 @@
+# Please make sure to cap all dependency versions, in order to avoid unwanted
+# functional and integration breaks caused by external code updates.
+#
+# These are general dependencies, so:
```
MattiSG, line 4:
```suggestion
# These dependencies are always installed, so:
```

requirements/install
```diff
@@ -0,0 +1,68 @@
...
+#
+# * the ranges should be broad, and
+# * the number of them kept discreet.
+#
+# If a dependency seems trivial, please propose a pull request to remove it.
```
MattiSG, line 9:
```suggestion
# If a dependency seems redundant, please propose a pull request to remove it.
```

requirements/install
```diff
@@ -0,0 +1,68 @@
...
+#
+# General Rule
+# ========
+#
+# * Cap to latest major.
+#
+# NumPy Rule
+# ========
+#
+# * Cap to latest four minors (document why).
```
MattiSG, line 19: This seems to be documented below already, appropriately at the `numpy` declaration line. Remove this part?

requirements/install
```diff
@@ -0,0 +1,68 @@
...
+#
+# Exceptions
+# ========
+#
+# * Avoiding a version (document why).
+# * Pinning a version (document why).
+
+# For globing over dictionaries as if they were filesystems.
```
MattiSG, line 27:
```suggestion
# For globbing over dictionaries as if they were filesystems.
```

requirements/install
```diff
@@ -0,0 +1,68 @@
...
+#
+# We support from 1.3.2 on because it is the first version published
+# following the semantic versioning specification.
+#
+# We do not support 2 yet because of a bug introduced to period parsing in
+# OpenFisca's Web API.
```
MattiSG, line 33: > Link to an issue?

requirements/install
```diff
@@ -0,0 +1,68 @@
...
+# For openfisca test
+#
+# We support pytest >= 5.4.2 because `openfisca test` relies on the signature of
+# `pytest.Item.from_module()` introduced since this 5.4.2.
+#
+# We do not support 6 yet because it requires fixing some tests before.
+# See: https://docs.pytest.org/en/stable/deprecations.html#node-construction-changed-to-node-from-parent
+pytest >= 5.4.2, < 6
+
+# For parameters, tests.
```
MattiSG, line 64:
```suggestion
# For parameters and tests parsing.
```

requirements/tracker
```diff
@@ -0,0 +1,7 @@
+# Please make sure to pin all dependency versions, in order to avoid unwanted
+# functional and integration breaks caused by external code updates.
+#
+# These are web-api tracking dependencies, so pin them.
```
MattiSG, line 4: I'm not sure I understand why this is pinned. I'd prefer that we can deploy an upgrade to this code without requiring a Core upgrade.

requirements/web-api
```diff
@@ -0,0 +1,12 @@
+# Please make sure to pin all dependency versions, in order to avoid unwanted
+# functional and integration breaks caused by external code updates.
+#
+# These are web-api dependencies, so pin them.
```
MattiSG, line 4: > This comment does not seem to provide a lot of additional clarity, at least to me 😅

setup.py
```diff
@@ -1,57 +1,46 @@
...
-general_requirements = [
-    'dpath >= 1.5.0, < 2.0.0',
-    'pytest >= 4.4.1, < 6.0.0',  # For openfisca test
-    'numpy >= 1.11, < 1.21',
-    'psutil >= 5.4.7, < 6.0.0',
-    'PyYAML >= 3.10',
-    'sortedcontainers == 2.2.2',
-    'numexpr >= 2.7.0, <= 3.0',
-    ]
+def require(filename: str) -> List[str]:
```
MattiSG, line 8: Using `require` as a local function name in a requirements file makes for unclear declarations below.
```suggestion
def load_requirements_from_file(filename: str) -> List[str]:
```

setup.py
```diff
@@ -64,14 +53,16 @@
             'openfisca-run-test=openfisca_core.scripts.openfisca_command:main',
             ],
         },
+    python_requires = ">= 3.7",
+    install_requires = open("requirements/install").readlines(),
```
MattiSG, line 57: Why not use the same loading mechanism (`require` / `load_requirements_from_file`) as the `extras_require`?

setup.py
```diff
@@ -1,57 +1,46 @@
...
     author_email = 'contact@openfisca.org',
     classifiers = [
         'Development Status :: 5 - Production/Stable',
         'License :: OSI Approved :: GNU Affero General Public License v3',
         'Operating System :: POSIX',
         'Programming Language :: Python',
-        'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
+        'Programming Language :: Python :: 3.8',
+        'Programming Language :: Python :: 3.9',
```
MattiSG, line 37: Which parts of this changeset enable support for Python 3.8 and 3.9? 🤔

MattiSG (changes requested):
Exciting! 😃

I understand that this PR brings a few different sets of changes:

1. Refactor of the way requirements are declared: from arrays in the `setup.py` to separate files in the `requirements` folder.
2. Pinning down all non-default requirements versions.
3. Supporting constraining the `numpy` version.
4. Change the supported Python version from 3.6 and 3.7 to 3.7, 3.8 and 3.9.
5. Refactor all CI-specific dependencies to the standard requirements declaration.

If this is correct, then:

- I am very much in support of (5).
- While in favour of it, I don't understand how this changeset enables (4). I believe we should add matrix testing in CI if we want to support multiple Python versions.
- I don't understand the expected usage of (3): if we advertise support for multiple `numpy` versions, why do we test only against the oldest version?
- I don't understand the added value of (2), while I see the cost of needing to deploy a new Core version whenever a patch is published in one of the dev requirements 😰
- I am not sure I understand the added value of (1). I like how each set of requirements has its own file, but I also like that all requirements are in `setup.py`. What is the most standard in Python dependency management?

maukoquiroga:
Thanks for the review!

> * I am very much in support of (5).

🙌

> * While in favour of it, I don't understand how this changeset enables (4). 

You're right, I forgot to add [pyupgrade](https://github.com/openfisca/country-template/pull/111/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R26).

> I believe we should add matrix testing in CI if we want to support multiple Python versions.

We could. I'm assuming 3.7 compatible code will also be 3.7+, albeit naively maybe.

> * I don't understand the expected usage of (3): if we advertise support for multiple `numpy` versions, why do we test only against the oldest version?

Again, I'm naively assuming that if 1.17 works, and 1.21 works, all in between should work.

> * I don't understand the added value of (2)

The idea is to:

1. reduce entropy for contributors —without impact for reusers
2. being able to use pip without the legacy option —currently it is too slow as dep ranges are too broad

> while I see the cost of needing to deploy a new Core version whenever a patch is published in one of the dev requirements 😰

It's definitely a trade-off. 

My assumption is that we don't need to:

- core dep ranges needs to be broad —we can't know reusers' constraints upfront
- numpy specifically needs also to be retro-compatible, for the same reason
- the rest do not, as dev is done locally, so we can live with outdated dev deps and update them once every 3 months for ex.

Again it is just an assumption.

> * I am not sure I understand the added value of (1). I like how each set of requirements has its own file, but I also like that all requirements are in `setup.py`. What is the most standard in Python dependency management?

Seems to be to use different files (both for requirements and for constraints), similar to what's proposed here.

I did actually borrowed the idea to both https://github.com/opendatateam/udata and https://github.com/jazzband/pip-tools.

README.md
maukoquiroga: To assure compatibility for @openfisca/france-contrib-ipp and @openfisca/france-contrib-leximpact.

But it is still arbitrary: whatever number or road we choose, I think it should be properly documented as it is a contract we assume with our users.

requirements/dev
maukoquiroga: There's a bit of a tricky decision to be made here, the context:

- Core & the templates are circular dependencies
- That prevents us of using the latest pip version, without the legacy resolver
- In order to fix that, several scenarios, notably two:
  - Relax versions somewhere, and pin them somewhere else —as proposed here
  - Extract the dependencies, with the trade-off of losing integration tests, as discussed already in another issue.

I think this requires a proper issue, exclusively to fix the `pip` problem.

As proposed [here](https://github.com/openfisca/openfisca-core/pull/1020#issue-629112079):

> Force to phase breaking changes so to adapt the packages depending on OpenFisca-Core

requirements/install
maukoquiroga: I was meaning _unnecessary_. Maybe that's a better wording?

requirements/tracker
maukoquiroga: Agreed, I guess pinning only makes sense for dev deps.

requirements/web-api
maukoquiroga: Agreed, I guess pinning only makes sense for dev deps.

setup.py
maukoquiroga: Leftover! Thanks :)

requirements/web-api
MattiSG: I was really focusing on the comment itself more than on arguing the validity of pinning those dependencies :)

README.md
MattiSG: Based on the discussion in #1012, I understand that version upgrades are costly based more on time periods when they happen than on the version number difference, so I am not sure a “4 last versions” policy makes any specific sense. But this PR might not be the place to discuss it. It's probably worth having a discussion and opening up an RFC on that topic.

MattiSG: My suggestion for rephrasing still stands in the current context 🙂

requirements/install
MattiSG: Yep, good for me.

MattiSG:
> I'm assuming 3.7 compatible code will also be 3.7+

If that assumption held, then why wouldn't 3.6-compatible code be 3.7-compatible? 🤔 

> I'm naively assuming that if 1.17 works, and 1.21 works

This sounds reasonable but I wouldn't advertise compatibility explicitly if there is no guaranteed testing.
Also, I'm not sure I understand how the system implemented in this PR checks for both upper and lower bounds. I understand it checks explicitly for the lower bound, and checks for some version matching version constraints, on which we have no control, and we assume it should be the latest but we are not certain (maybe the cache is old and the package manager is satisfied with it).

> 1. reduce entropy for contributors —without impact for reusers

Do we have documented cases of debugging / contribution being hindered and that would have been solved by pinning?

> we can't know reusers' constraints upfront

Exactly, that's why I'm a bit worried with pinning dependencies.

I'd love to have @openfisca/france-contrib-ipp, @guillett, @MaxGhenis… chime in on that. What would be the expected impact of pinning down dependencies for your teams?

maukoquiroga:
Thanks @MattiSG, I'll split this PR so to move incrementally while reaching consensus on the controversial bits 👍

requirements/web-api
maukoquiroga: Added a better comment, let me know if it's ok.

requirements/dev
maukoquiroga: I improved the comment, please let me know if that's OK.

requirements/dev
maukoquiroga: I proposed changes unpinning them, please let me know what do you think.

maukoquiroga:
Well I did a couple of changes, it looks better now :)

maukoquiroga:
> Thank you very much for your work!
> 
> This gets pretty close to shipping. I believe it might be appropriate to use this very impactful change to ship as many breaking deprecations as possible at once, as reusers are in for some adaptation work. I think of removing `Dummy`, for example. Do you see others?

I wanted to deprecate half of the codebase ...

maukoquiroga:
Where I am at so far https://github.com/maukoquiroga/pysemver (will retex in a bit)
________________________________
De : Matti Schneider ***@***.***>
Envoyé : Monday, October 11, 2021 9:07:55 AM
À : openfisca/openfisca-core ***@***.***>
Cc : Mauko Quiroga ***@***.***>; Author ***@***.***>
Objet : Re: [openfisca/openfisca-core] Improve dependency management (#1015)


> Are we planning on doing matrix testing with #1030 <https://github.com/openfisca/openfisca-core/issues/1030>? If so, we could just simply drop this
>
Yes, matrix testing is definitely on the way! 😃
If you see a smarter way to define the specific target versions than by declaring them in the CI config file, let’s discuss this!

maukoquiroga:
@MattiSG ?

MattiSG:
When I see the impact on country-template (https://github.com/openfisca/country-template/pull/121), I am clearly in favour of getting rid of the system of deprecation and actually using major version breaks. I fail to understand the added value of this delayed way of updating the API, and now we have PRs that are not backwards-compatible yet have no version bump 🙃

requirements/common
benoit-cty, line 1: Can I suggest to name the files in `requirements` in the form `requirements-common.txt` instead of just `common` to let understand that it can be used with `pip install -r` ?

bonjourmauko: This is an old PR, I'll follow the extact same pattern than in `extension-explorer`.

sandcha: Hello @bonjourmauko, 

Did you mean [extension-template](https://github.com/openfisca/extension-template)? ^^
Do you want to close this PR in order to replace it with another one?

bonjourmauko: Sure.

bonjourmauko:
As things have converged toward the use of a `lockfile`, I'm closing this PR in favour of opening a new one.

PR #1020, opened 2021-05-03
Run country & extension template tests systematically

maukoquiroga:
#### Technical changes

- Run country & extension template tests systematically
  - This will help prevent maintainers integrating incompatible breaking changes
  - Also this will force to phase breaking changes so to adapt the packages depending on OpenFisca-Core


Makefile
```diff
@@ -28,7 +34,9 @@ format-style:
...
 test: clean check-syntax-errors check-style check-types
-	env PYTEST_ADDOPTS="$$PYTEST_ADDOPTS --cov=openfisca_core" pytest
+	PYTEST_ADDOPTS="$$PYTEST_ADDOPTS ${pytest_addopts}" pytest
+	openfisca test ${COUNTRY_TEMPLATE_TESTS} -c ${COUNTRY_TEMPLATE} ${optional_arguments}
+	openfisca test ${EXTENSION_TEMPLATE_TESTS} -c ${COUNTRY_TEMPLATE} -e ${EXTENSION_TEMPLATE} ${optional_arguments}
 
-api:
-	openfisca serve --country-package openfisca_country_template --extensions openfisca_extension_template
+serve:
+	openfisca serve -c ${COUNTRY_TEMPLATE} -e ${EXTENSION_TEMPLATE} ${optional_arguments}
```
MattiSG, line 42: Always prefer long options in scripts to ensure readability. Short options are only syntactic sugar for interactive usage.

```suggestion
	openfisca serve --country-package ${COUNTRY_TEMPLATE} --extensions ${EXTENSION_TEMPLATE} ${optional_arguments}
```

Makefile
```diff
@@ -28,7 +34,9 @@ format-style:
 	autopep8 `git ls-files | grep "\.py$$"`
 
 test: clean check-syntax-errors check-style check-types
-	env PYTEST_ADDOPTS="$$PYTEST_ADDOPTS --cov=openfisca_core" pytest
```
MattiSG, line 31: I don't know what it was used for locally, but are we sure we want to drop the `--cov=openfisca_core` option and now only run it in CI?

Makefile
```diff
@@ -28,7 +34,9 @@ format-style:
...
+	PYTEST_ADDOPTS="$$PYTEST_ADDOPTS ${pytest_addopts}" pytest
+	openfisca test ${COUNTRY_TEMPLATE_TESTS} -c ${COUNTRY_TEMPLATE} ${optional_arguments}
+	openfisca test ${EXTENSION_TEMPLATE_TESTS} -c ${COUNTRY_TEMPLATE} -e ${EXTENSION_TEMPLATE} ${optional_arguments}
 
-api:
-	openfisca serve --country-package openfisca_country_template --extensions openfisca_extension_template
+serve:
```
MattiSG, line 41: Are we sure this should not be a breaking change if we change the `make` target name? What if some user automates with `make api`?

> I'm still in favour of this rename, but I believe we should either add an alias or make this a breaking change (or drop this target altogether because I don't understand what's the point of `make test` vs `openfisca test`).

Makefile
```diff
@@ -28,7 +34,9 @@ format-style:
 	autopep8 `git ls-files | grep "\.py$$"`
 
 test: clean check-syntax-errors check-style check-types
-	env PYTEST_ADDOPTS="$$PYTEST_ADDOPTS --cov=openfisca_core" pytest
+	PYTEST_ADDOPTS="$$PYTEST_ADDOPTS ${pytest_addopts}" pytest
+	openfisca test ${COUNTRY_TEMPLATE_TESTS} -c ${COUNTRY_TEMPLATE} ${optional_arguments}
+	openfisca test ${EXTENSION_TEMPLATE_TESTS} -c ${COUNTRY_TEMPLATE} -e ${EXTENSION_TEMPLATE} ${optional_arguments}
```
MattiSG, line 39: > This seems acceptable considering Country and Extension templates [are already](https://github.com/openfisca/openfisca-core/blob/master/setup.py#L32-L33) dev dependencies.

MattiSG (dismissed):
Thank you @maukoquiroga! 👏 

This works for me with `pip install openfisca-core[dev] && make test`, I confirm that all of Core, Country Template and Extension Template tests are run.

This adds a risk: if Country or Extension template tests are broken, Core tests will break as well even though this would not be related to Core changes. I believe it is an acceptable risk considering that these repository have CI and should never fail anyway, against the added value of checking that Core changes don't break the templates.

Makefile
maukoquiroga: Good catch! I think that change has little to do with this PR.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+### 35.4.3 [#1020](https://github.com/openfisca/openfisca-core/pull/1020)
+
+#### Technical changes
+
+- Run country & extension template tests systematically
+  - This will help prevent maintainers integrating incompatible breaking changes
+  - Also this will force to phase breaking changes so to adapt the packages depending on OpenFisca-Core
```
MattiSG, line 9: > I'm not sure these justifications are useful in the Changelog 🙂

Makefile
```diff
@@ -1,34 +1,165 @@
+OK := "$$(tput setaf 2) [✓] $$(tput sgr0)"
+KO := "$$(tput setaf 1) [!] $$(tput sgr0)"
+PASS := "$$(tput setaf 3) [~] $$(tput sgr0)"
+WORKING := "$$(tput setaf 5)[⚙]$$(tput sgr0)"
+OPTIONS := "$$(tput setaf 6)[ℹ]$$(tput sgr0)"
+WARNINGS = "$$(tput setaf 3)[ℹ]$1$$(tput sgr0)"
```
MattiSG, line 6: Why not `:=` on this line?

Makefile
```diff
@@ -1,34 +1,165 @@
+OK := "$$(tput setaf 2) [✓] $$(tput sgr0)"
```
MattiSG, line 1: (repeat below)
```suggestion
OK_SYMBOL := "$$(tput setaf 2) [✓] $$(tput sgr0)"
```

MattiSG (changes requested):
Thank you for having taken my suggestions into account 🙂

However, the changeset that was originally reviewed has considerably increased, and this PR seems to have suffered from some feature creep… What was initially a way to “run country & extension template tests systematically” now seems to offer an integral rewrite of the makefile and its logs.

This is exciting, but unfortunately means many more checks have to be done, notably on the cross-platform aspect since some shell-specific syntax is used. It would have to be checked against Windows console, CI, zsh-enabled macOS… Could we please refocus this PR on its original scope? I understand this might be frustrating 😟 but I would really love to see the original proposal land in Core, while minimising risks and enabling a proper conversation around build tools evolution 🙂

maukoquiroga:
Ok with the feature creep 😄 , I'll split things up.

MattiSG:
Thanks @maukoquiroga!

Makefile
```diff
@@ -27,8 +32,32 @@ format-style:
...
 	autopep8 `git ls-files | grep "\.py$$"`
 
-test: clean check-syntax-errors check-style check-types
-	env PYTEST_ADDOPTS="$$PYTEST_ADDOPTS --cov=openfisca_core" pytest
+## Run openfisca-core & country/extension template tests.
+test:: clean check-syntax-errors check-style check-types
+test:: \
+	$(shell git ls-files "tests/*.py") \
+	${python_packages}/${country_template}/tests \
+	${python_packages}/${extension_template}/tests
```
MattiSG, line 40: After running the tests locally, I believe it would add a lot to readability if the Core tests were run separately from the Country and Extension templates, both here and in CI.
Could we create dedicated targets for each of those, with `test` running all three? 🙂

MattiSG (changes requested):
Nice! I'm still a bit appalled at the complexity of the Makefile syntax, I must say, but thank you so much for mastering it and improving it!

Makefile
```diff
@@ -27,8 +32,32 @@ format-style:
 	@# `make` needs `$$` to output `$`. Ref: http://stackoverflow.com/questions/2382764.
 	autopep8 `git ls-files | grep "\.py$$"`
 
-test: clean check-syntax-errors check-style check-types
-	env PYTEST_ADDOPTS="$$PYTEST_ADDOPTS --cov=openfisca_core" pytest
+## Run openfisca-core & country/extension template tests.
+test:: clean check-syntax-errors check-style check-types
+test:: \
```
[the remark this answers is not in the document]
maukoquiroga: Yes.

Makefile
```diff
@@ -1,3 +1,8 @@
+working_symbol := "$$(tput setaf 5)[⚙]$$(tput sgr0)"
+python_packages := $(shell python -c "import sysconfig; print(sysconfig.get_paths()[\"purelib\"])")
+country_template := openfisca_country_template
+extension_template := openfisca_extension_template
```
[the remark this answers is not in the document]
maukoquiroga: Yeah I've removed them.

openfisca_core/tools/test_runner.py
```diff
@@ -64,7 +64,7 @@ def run_tests(tax_benefit_system, paths, options = None):
 
     """
 
-    argv = ["--capture", "no"]
+    argv = []
```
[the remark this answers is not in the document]
maukoquiroga: Refactored and documented it.

benjello:
Hello @MattiSG @maukoquiroga : this PR stands in the way of some other PRs.
Shouldn't we merge it ASAP or do you plan an overhaul of what is fixed here ?

maukoquiroga:
I think it is safe to merge this one.

benjello:
@MattiSG does it LGTM for you ?

Makefile
```diff
@@ -51,10 +52,38 @@ check-types: openfisca_core openfisca_web_api
...
+	@##
+	@##		make test
+	@##		make test pytest_args="--exitfirst"
+	@##		make test openfisca_args="--performance"
+	@##		make test pytest_args="--exitfirst" openfisca_args="--performance"
+	@##
+	@$(call help,$@:)
+	@${MAKE} test-core
+	@${MAKE} test-country
+	@${MAKE} test-extension
```
MattiSG, line 71: Why is this not simply added as a dependent target to `test`, like `clean` etc? 🤔

Makefile
```diff
@@ -51,10 +52,38 @@ check-types: openfisca_core openfisca_web_api
...
+	@$(call help,$@:)
+	@PYTEST_ADDOPTS="${PYTEST_ADDOPTS} --cov=openfisca_core ${pytest_args}" openfisca test $? ${openfisca_args}
+
+# Run country-template tests.
+test-country: ${python_packages}/openfisca_country_template/tests
+	@$(call help,$@:)
+	@PYTEST_ADDOPTS="${PYTEST_ADDOPTS} ${pytest_args}" openfisca test $? --country-package openfisca_country_template ${openfisca_args}
+
+# Run extension-template tests.
+test-extension: ${python_packages}/openfisca_extension_template/tests
```
MattiSG, line 84: > I am not sure I understand the added value of passing the folder as a parameter when it is used only in one place in the target 😉

.circleci/config.yml
```diff
@@ -31,7 +31,7 @@ jobs:
 
       - run:
           name: Run Core tests
-          command: env PYTEST_ADDOPTS="--exitfirst" make test
+          command: make test pytest_args="--exitfirst"
```
MattiSG, line 34: I would love to have different steps for Core, Country Template and Extension Template visible in CI 🙂

MattiSG (approved):
👍

maukoquiroga:
@MattiSG I think we just changed settings to dismiss stale reviews after #990. I've just deactivated it for now.

PR #1023, opened 2021-05-31
Add script and methods to save YAML parameter tree on disk

benjello:

#### New features

- Introduce a script to unfold and normalize parameters

### DIFF
--- a/openfisca_core/parameters/__init__.py
+++ b/openfisca_core/parameters/__init__.py
@@ -33,7 +33,7 @@ from .config import (  # noqa: F401
     )
 
 from .at_instant_like import AtInstantLike  # noqa: F401
-from .helpers import contains_nan, load_parameter_file  # noqa: F401
+from .helpers import contains_nan, load_parameter_file, save_parameters_to_dir  # noqa: F401
 from .parameter_at_instant import ParameterAtInstant  # noqa: F401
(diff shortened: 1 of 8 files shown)


benjello:
@Sasha-Laniece @sandcha @eraviart : here @cbenz script to unfold an normalize parameters. 
It would be nice to be able to use it for the convergence openfisca-france <-> barèmes IPP but it may be useful for larger purposes. But please review it to see if it is sufficient or can be improved. 

It would be nice to have a complementary tool to fold parameters at a fixed directory depth/path length too.

This was used for exemple in https://github.com/openfisca/openfisca-france/pull/1414.

MattiSG:
Thanks @benjello @cbenz for sharing this script which, if I understand correctly, generates a dump of all parameters in their full name in a YAML file!

In order to better assess how this could be useful to the wider community, could you please:

1. Confirm what this does (I also see mentions of [maximum name length](https://github.com/openfisca/openfisca-core/pull/1023/files#diff-5a56534ae14d08c5e7908cb8b42c55a92d3d0a8e09943c63ccf018a5b68c640fR21-R26), so maybe it does more than just dumping a YAML list?).
2. Clarify what are the use cases (how does it help in converging two separate sets of parameters?).
3. Provide an example output of this script (if possible on the Country Template, so that we have a clear and concise reference).

Thank you! 🙂

Issue #1030, opened 2021-08-19
Switch from CircleCI to GitHub Actions


MattiSG:
# RFC

This RFC suggests switching the default CI provider for OpenFisca codebases from CircleCI to GitHub Actions, enabling a 5x performance gain.

## Context

Currently, OpenFisca tests in continuous integration run with [CircleCI](https://circleci.com)’s free plan. This is the case for OpenFisca Core, but also for the Country Template, and for most known country packages (since they derive from the Country Template).

With the main aim to decrease the duration of France's longer test runs (between 10 and 12 minutes per build), an investigation was led to identify ways to speed up test that could benefit the entire community in #1027.

Some minor ways for improvement were identified and implemented (https://github.com/openfisca/openfisca-france/pull/1649, https://github.com/openfisca/openfisca-france/pull/1650) and relevant ones are or soon will be shared back in the Country Template (https://github.com/openfisca/country-template/pull/113).

However, by far the most effective way to speed up the tests that was identified is to parallelise them.

There is some exploratory work to do this for local tests in #1025. In CI, however, parallelisation has been effective for 4 years (since #516).

## Current situation

With CircleCI's free plan for open-source, we can run builds with up to 4x parallelism.

## Expected situation

With GitHub Action's free plan for opens-source, we would run builds with up to 20x parallelism.

## Expected gains

- 5 times faster test runs, as measured and evidenced in https://github.com/openfisca/openfisca-core/issues/1027#issuecomment-895878588.
- Deeper integration within GitHub.

## Expected costs

- Writing a new CI setup for both Core and the Country Template. This has already been covered a lot in #1027’s exploration, and has been volunteered by @openfisca/france-contrib-dinsic (in particular @HAEKADI).
- Setting up the new CI pipeline. This has already been covered a lot in #1027’s exploration, and has been volunteered by @openfisca/international-maintainers (in particular @sandcha).

## Expected losses

- Detailed logs from test runs won't be retained after 90 days.

## Workarounds

- Rely on worflow-level metadata (which is kept indefinitely) for performance decisions made on test runs, and learn to copy-paste in the issue body the subset of the logs when we make a decision based on them.

## More information

- A detailed comparison table is available in https://github.com/openfisca/openfisca-core/issues/1027#issuecomment-901924153.
- The CircleCI account and pipeline will stay active for the foreseeable future, letting each country package migrate on its own schedule.

# Decision-making process

[...]

Issue #1027, opened 2021-07-27
Improve tests performance


MattiSG:
It would be useful for all country packages to improve the speed of tests, and to improve the speed of computations in general. Currently, running all tests in the most complex tax and benefits system (`OpenFisca-France`) takes [10 to 20 minutes](https://app.circleci.com/pipelines/github/openfisca/openfisca-france?branch=master) in CI, and users report the same duration locally (#1025).

This issue aims at consolidating potential venues for performance improvements, associated costs, risks and expected benefits, in order to define the best implementation strategy for a team with limited resources that cannot tackle all at once.

| Option | Implementation | Gain | Risk |
|--------|----------------|-----|-----|
| Tests parallelisation | #1025 | Only locally, depends on cores count | Decrease in test output readability |
| Parameters caching | https://github.com/openfisca/openfisca-france/pull/1311 | For every execution; not clear if each test run is its own execution | Cache mysteries |
| Native YAML parser | _to be done_ | For every execution; not clear if each test run is its own execution | Platform-specific incompatibilities |
| Install `libyaml` in CI | _to be done_ | For every execution; not clear if each test run is its own execution | —  |

### Native YAML parser

The “Native YAML parser” option consists of assessing whether `libyaml`, currently suggested as an optional dependency as it speeds up parameter and tests loading by using a native compiled extension instead of the native Python YAML reader, could not be added as a systematic dependency. The last assessment dates back from a few years and revealed platform-specific incompatibilities. Maybe this has been fixed in upstream. There might be also be an alternative option that fits the same purpose with less incompatibilities.

## Strategy & contributions

- @HAEKADI will consolidate here results of exploring the “Native YAML parser” and “Install `libyaml` in CI” options.
- All other suggestions to explore are welcome!
- Benchmarks for current tests durations in different country packages are welcome!
- Declarations of interest for benchmarking your own country package against potential options are welcome!
- Analyses assessing which option(s) might be most impactful are welcome.

maukoquiroga:
Great initiative! 🙌

I share with you a couple of learnings in the hope it helps.

I think what's good about the options proposed is that they not mutually exclusive.

##### Tests parallelisation

Seems to me one of the the most impactful in terms of raw performance, and scales horizontally.

However, I'm still blocked in trying to make it work, as there's some black magic: 

- `xdist` launches independent processes to collect and run the tests, which are then invisible to the one that launched `pytest` in the first place, and
- it replaces `openfisca test` tests collectors with its own.

[...]

Issue #1025, opened 2021-06-25
Introduce a way to run test in parallel


benoit-cty:
Hi there!

I really enjoy OpenFisca, but I recently encountered an issue.

### Here is what I did:

`openfisca test  --country-package openfisca_france tests`

### Here is what I expected to happen:

Use of all the core of my CPU to run test so that it took less than 3 minutes.

### Here is what actually happened:

Only one core is used, so test took 17 minutes.

## Context

I identify more as a:

- Developer _(I create tools that use the existing OpenFisca code at https://LexImpact.an.fr)_.

# Experimentation I have made

## Use GNU parallel

`find tests/**/*.{yaml,yml} | parallel -j 32 --progress openfisca test  --country-package openfisca_france | grep FAILED`
Pros:
- it is really fast, less than 3 minutes.
Cons:
- All the test output is printed to the console, so you need to process them to know error, with `grep FAILED` for example.
- Need an install at OS level `sudo apt-get install parallel`.

## Use of Xdist
pytest-xdist is a Pytest plugins that run test in parallel.
To use it:
`pip install pytest-xdist[psutil]`
Add in setup.cfg:
```
[tool:pytest]
addopts      = -n auto` 
```
Run the test
`openfisca test  --country-package openfisca_france tests`

Pros: 
- Use all CPU core
- Nice pytest output
Cons:
- Need to upgrade pytest version dependencies 
- Don't work as expected : it run the test of the project like a normal pytest, not the OpenFisca YAML testfiles.

It seems that the OpenFisca pytest plugin https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/tools/test_runner.py need to override some xdist method to give it the right test.

## Use pytest-parallel
https://github.com/browsertron/pytest-parallel
Not tested, it seems to work like Xdist, so may have the same problem.

maukoquiroga:
Hi @benoit-cty and thanks for this issue.

I see there is an option overlap regarding `-n` https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/scripts/openfisca_command.py#L34-L43.

So I've tried this:

`PYTEST_ADDOPTS="--numprocesses=auto" openfisca test --country-package openfisca_france tests`

And got:

`
platform darwin -- Python 3.7.0, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /Users/hyperion/Sites/openfisca/openfisca-france, configfile: setup.cfg
[...]

PR #1031, opened 2021-08-20
Run tests in a parallel matrix

maukoquiroga:
Hello!

(Working) prototype for solving #1025 and #1035, poke #1027, please chime in! 😄 

⚠ This is **not** a proposal but a demonstration of the different tools, so we can discuss and choose what make more sense in terms both of _performance_ and _readability_ to the community.

**Trying it out**

```
$ make test.matrix
...

real	2m13.152s
user	10m12.086s
sys	1m48.311s
```

#### New features

- Run `openfisca test` tests in parallel with `make`, `nox`, and `invoke`
- Run tests against Python 3.7+
- Run tests against NumPy 1.18+

#### Notes

- We don't have control over the way subprocesses are launched and run, so reporting could be a bit confusing
- Performance gains are relative to the number of cores in your machine
- OpenFisca is not `threadsafe`, so each subprocess will run a single thread
- If one day we solve concurrency problems, ... that could go faster.

#### TODO

- [x] Run each test suite with `invoke`
- [x] Run the matrix of tests with `nox`
- [x] Parallelize with `make`
- [x] Parallelize with `pytest-parallel`
- [ ] Parallelize with `pytest-split`
- [x] Run .py tests in parallel
- [ ] Run .yaml tests in parallel
- [ ] Provide decent reporting
- [ ] What to do in CI #1030 
- [x] Benchmark


MattiSG:
Thanks @maukoquiroga! The promise of such a speed improvement is exciting, and I would be very appreciative of a refactor / overhaul of the current test runner 😃

However, I must say that the changeset is pretty daunting, and seems to have impacts way beyond the test runner, with some core functions having their visibility changed… Could we maybe convene of a video discussion with other interested @openfisca/international-contrib contributors, in order to understand the goals, side effects and side benefits? 🙂

In terms of pure performance and parallelisation, it might also be that #1030 allows us to use https://github.com/nektos/act as suggested by @benoit-cty, and we could have local parallelisation with the exact same setup as CI, which I would be very excited about 😃

maukoquiroga:
> However, I must say that the changeset is pretty daunting, and seems to have impacts way beyond the test runner, with some core functions having their visibility changed…

Some of them were absolutely required, in which case I've tried to avoid a breaking change by exposing changes through `__init__.py` files, some maybe just because I just was rushing to make it work (hence the draft status).

Indeed a partial refactoring of the test runner was imposed in order to make this work, which could definitely go in a PR on its own, with proper deprecations, and agreed upon breaking changes.

maukoquiroga:
Closing for now.

PR #1033, opened 2021-08-23
Improve openfisca-core's commons module doctests

maukoquiroga:
### Motivation

Python provides `doctests` to help keep a testable documentation.

They provide some benefits:

- It helps maintainers and contributors to spot bugs and improvement opportunities
- It helps reusers to understand the code quickly
- It helps separating tests (unit, functional, ...)
- It produces the [official doc](https://openfisca.org/doc)

### Complications

Today the code coverage by doctests is mild, they are not run, and most of them are broken. 

Some observed consequences:

- The code is in overall hard to grasp without a consequent time investment, thus hard to contribute to
- The code is hard to test unitarily due to its current design
- There are then very few unit tests
- The documentation is [broken](https://github.com/openfisca/openfisca-doc/issues/244)

### Proposal

Fix, complete, and test doctests systematically!

#### Technical changes

- Improve openfisca-core's commons module doctests
  - Fix existing doctests
  - Add missing doctests
  - Remove duplication between doctests and pytest


openfisca_core/commons/tests/test_rates.py
```diff
@@ -0,0 +1,28 @@
+"""Tests for the rates module."""
+
+import numpy
+from numpy.testing import assert_array_equal
+
+from openfisca_core import commons
+
+
+def test_average_rate_when_varying_is_zero():
+    """When the varying gross income crosses zero it yields infinity."""
```
HAEKADI, line 10: > ```suggestion
>     """Yields infinity when the varying gross income crosses zero."""
> ```

openfisca_core/commons/tests/test_rates.py
```diff
@@ -0,0 +1,28 @@
...
+    target = numpy.array([1, 2, 3])
+    varying = 0
+
+    result = commons.average_rate(target, varying)
+
+    assert_array_equal(result, [- numpy.inf, - numpy.inf, - numpy.inf])
+
+
+def test_marginal_rate_when_varying_is_zero():
+    """When the varying gross income crosses zero it yields infinity."""
```
HAEKADI, line 21: Maybe add more details to show a distinction from the the previous test? 
> ```suggestion
>     """Yields infinity when the varying gross income crosses zero."""
> ```

HAEKADI (commented):
Thank you @maukoquiroga! I ran the tests and everything seems to be in order :)

MattiSG (changes requested):
Thank you @maukoquiroga, this is impressive documentation work! 😲 👏

How are these tests run in CI?

I would personally need the following to validate that PR:

1. A review by at least one person who has extensive knowledge of the API and can double-check both the given information and that this is isofunctional in terms of tests. I'll ask for a review accordingly.
2. As mentioned in a comment, that all style convention changes are extracted from this PR.
3. That an explanation on how to run these tests and produce the documentation is added, at least to this PR, possibly to the doc itself, so that I can validate the changeset on my machine 🙂

MattiSG:
How can I confirm this would fix https://github.com/openfisca/openfisca-doc/issues/244? 🙂

setup.cfg
```diff
@@ -1,22 +1,50 @@
...
+    D202,
+    D412,
+    E128,
+    E251,
+    F403,
+    F405,
+    E501,
+    RST301,
+    W503,
+    W504
```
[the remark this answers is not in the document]
maukoquiroga: > I understand this changes some of the style conventions.

Hello!

In terms of code style, everything remains the same, no change of style.

In terms of documentation, more than changing styles it narrows them down:

There are 4 coexisting doc styles today in the code:
  - Sphinx style
  - Numpy style
  - Google style
  - Invalid documentation 

And none of them are linted nor tested.

My intention is to be able to test both doctests & docs in CI programatically.

maukoquiroga:
Thanks for your comments @HAEKADI & @MattiSG 



> How can I confirm this would fix [openfisca/openfisca-doc#244](https://github.com/openfisca/openfisca-doc/issues/244)? 🙂

Good point, I'll work on it.

setup.cfg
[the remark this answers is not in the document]
maukoquiroga: I get what you mean by (documentation) style now , roger ;)

maukoquiroga:
That was some work! So I've:

- Removed all style changes to code
- Added pytest to automatically check that commons doctests run
- Added linters to automatically check that docstrings are valid and up-to-date :) 🙌

@benjello regarding the style of _docstrings_ specifically, do you have any preference? To resume my take on this:

- There are four main styles: Google, NumPy, Sphinx, and PEP
- We currently have all four in the code 😄 and no official guideline as to prefer one over the others
- I like Google because is easier to write and read, and numpydoc could make sense because we use numpy a lot.

In any scenario, this changeset has little to no impact on that because we're adding/completing doc and doctests rather than rewriting it to a style or the other, but as we move on documenting the whole code base (are we?) it could make sense to have a collective community discussion to choose one style and adhere to it, as it'll make contributions and maintenance easier.

benjello:
@maukoquiroga : I like Google style too.

openfisca_core/commons/rates.py
```diff
@@ -1,14 +1,37 @@
...
+
+    Returns:
+        The average rate for each target.
+
+        When ``trim`` is provided, values that are out of the provided bounds
+        are replaced by :obj:`numpy.nan`.
+
+    Examples:
+        >>> target = numpy.array([1, 2, 3])
+        >>> varying = 2
```
benjello, line 28: Usually target and varying are array of the same size.

openfisca_core/commons/tests/test_rates.py
```diff
@@ -0,0 +1,26 @@
...
+from numpy.testing import assert_array_equal
+
+from openfisca_core import commons
+
+
+def test_average_rate_when_varying_is_zero():
+    """Yields infinity when the varying gross income crosses zero."""
+
+    target = numpy.array([1, 2, 3])
+    varying = 0
```
benjello, line 11: Usually target and varying are the same size.

openfisca_core/commons/tests/test_rates.py
```diff
@@ -0,0 +1,26 @@
...
+    result = commons.average_rate(target, varying)
+
+    assert_array_equal(result, [- numpy.inf, - numpy.inf, - numpy.inf])
+
+
+def test_marginal_rate_when_varying_is_zero():
+    """Yields infinity when the varying gross income crosses zero."""
+
+    target = numpy.array([1, 2, 3])
+    varying = numpy.array([0, 0])
```
benjello, line 22: Usually target and varying are the same size.

benjello (changes requested):

openfisca_core/commons/typing.py
```diff
@@ -1,11 +1,12 @@
...
 
-import numpy
+from nptyping import NDArray
+
 
 _T = TypeVar("_T", float, str)
 """:obj:`.TypeVar`: A wildcard type used by other types."""
 
-ArrayLike = Union[numpy.ndarray, Sequence[_T]]
+ArrayLike = Union[Union[NDArray[float], NDArray[str]], Sequence[_T]]
```
benjello, line 9: @maukoquiroga : shouldn't we add NDArrya[bool] and NDArray[int] ?

benjello (commented):
See my comment. 
My fear is the the current typing may force bool and int array conversion to float array. 
Otherwise good to go.
Thanks !

openfisca_core/commons/typing.py
```diff
@@ -2,11 +2,9 @@
 
 from nptyping import NDArray as Array
 
-
-_T = TypeVar("_T", float, str)
-""":obj:`.TypeVar`: A wildcard type used by other types."""
-
-ArrayLike = Union[Union[Array[float], Array[str]], Sequence[_T]]
+_Types = TypeVar("_Types", float, int, str, bool)
+_Array = Union[Array[float], Array[int], Array[str], Array[bool]]
```
benjello, line 6: Shoudn't we use alphabetical order ?
Or does the order is defined for testing purposes, the more likely being in front of the list ?

maukoquiroga: Alphabetical makes sense :)

maukoquiroga:
I followed @MattiSG advice in changing the base of this to #1038.

benjello (dismissed):

maukoquiroga:
Thanks @benjello ! This still requires #1038 for the tests to pass 😃 if could take a look.

MattiSG:
Couldn't this be merged into #1038 already though? 🙂

maukoquiroga:
@MattiSG It could, I just see a major drawback: #1038 standalone fixes the doc rendering, which is not perfect but already an acceptable deliverable step IMHO.

benjello (dismissed):

benjello (dismissed):

maukoquiroga:
Hello @MattiSG, would you be LGTM with the latest changes introduced?

> Thank you @maukoquiroga, this is impressive documentation work! 😲 👏
> 
> How are these tests run in CI?

All is run in CI now.

> 
> I would personally need the following to validate that PR:
> 
> 1. A review by at least one person who has extensive knowledge of the API and can double-check both the given information

I see @benjello you accepted this PR, do you confirm that the descriptions are correct?

> and that this is isofunctional in terms of tests. I'll ask for a review accordingly.

I have automated what can be: a coverage threshold limit, based on master (78 %).

Whenever a changeset falls below, the test suite will fail.

> 2. As mentioned in a comment, that all style convention changes are extracted from this PR.

Done, I've only kept (1) further syntax linting and (2) syntax/style linting of docs only, not code.

All very minimal realy.

> 3. That an explanation on how to run these tests and produce the documentation is added, at least to this PR, possibly to the doc itself, so that I can validate the changeset on my machine 🙂

Done! :)

benjello:
@maukoquiroga : the descriptions are ok.

HAEKADI (commented):
Thank you @maukoquiroga 👏 
I tested `make test-doc` from #1038 and it works.

maukoquiroga:
> Thank you @maukoquiroga 👏
> I tested `make test-doc` from #1038 and I got the following : 🤔
> 
> ```
> error: pathspec 'doctests/commons' did not match any file(s) known to git
> [!] The branch 'doctests/commons' doesn't exist, checking out 'master' instead...
> ```

And it passed :) It just means that all works as expected, maybe the message could be improved? Or the icon used.

HAEKADI:
> > Thank you @maukoquiroga 👏
> > I tested `make test-doc` from #1038 and I got the following : 🤔
> > ```
> > error: pathspec 'doctests/commons' did not match any file(s) known to git
> > [!] The branch 'doctests/commons' doesn't exist, checking out 'master' instead...
> > ```
> 
> And it passed :) It just means that all works as expected, maybe the message could be improved? Or the icon used.

Yes, it passed :) But does that mean it used the doc based on `master`or the current branch ? 🤔

maukoquiroga:
> Yes, it passed :) But does that mean it used the doc based on `master`or the current branch ? 🤔

On `openfisca_doc/master`, as it didn't found a `openfisca_doc/<current-core-branch>` to check-out.

maukoquiroga:
@HAEKADI So this is how it looks now:

Capture d’écran 2021-09-13 à 12 14 05

benjello:
Great @maukoquiroga !

@MattiSG @sandcha we need your feedback !

maukoquiroga:
Hi @MattiSG ! Is that changes introduced since your review make it LGTM? Please let me know what do you think so we can either integrate, or otherwise improve/reject this proposal 😃

Makefile
```diff
@@ -1,9 +1,18 @@
-help = sed -n "/^$1/ { x ; p ; } ; s/\#\#/[⚙]/ ; s/\./.../ ; x" ${MAKEFILE_LIST}
+info = $$(tput setaf 6)[i]$$(tput sgr0)
+warn = $$(tput setaf 3)[!]$$(tput sgr0)
+work = $$(tput setaf 5)[⚙]$$(tput sgr0)
+pass = echo $$(tput setaf 2)[✓]$$(tput sgr0) Good work! =\> $$(tput setaf 8)$1$$(tput sgr0)$$(tput setaf 2)passed$$(tput sgr0) $$(tput setaf 1)❤$$(tput sgr0)
+help = sed -n "/^$1/ { x ; p ; } ; s/\#\#/$(work)/ ; s/\./.../ ; x" ${MAKEFILE_LIST}
```
MattiSG, line 5: We at least need a comment on this incantation.

Makefile
```diff
@@ -1,9 +1,18 @@
-help = sed -n "/^$1/ { x ; p ; } ; s/\#\#/[⚙]/ ; s/\./.../ ; x" ${MAKEFILE_LIST}
+info = $$(tput setaf 6)[i]$$(tput sgr0)
```
MattiSG, line 1: > Suffix all those with `_message`, or prefix with `print`?
```suggestion
print_info = $$(tput setaf 6)[i]$$(tput sgr0)
```

Makefile
```diff
@@ -35,26 +44,48 @@ clean: \
 check-syntax-errors: .
 	@$(call help,$@:)
 	@python -m compileall -q $?
+	@$(call pass,$@:)
+
+## Run linters to check for syntax and style errors.
+check-style: \
+	check-style-all \
```
MattiSG, line 51: The difference between these two targets is not clear enough from their names, and is actually counterintuitive: `check-style-all` does less than `check-style` 🙃
I don't understand enough what each does to suggest a better name.

Makefile
```diff
@@ -35,26 +44,48 @@ clean: \
...
+	@flake8 `git ls-files | grep "\.py$$"`
 
-## Run code formatters to correct style errors.
-format-style: $(shell git ls-files "*.py")
+## Run linters to check for syntax and style errors.
+check-style-doc-%:
 	@$(call help,$@:)
-	@autopep8 $?
+	@flake8 --select=D101,D102,D103,DAR openfisca_core/$*
+	@pylint --enable=classes,exceptions,imports,miscellaneous,refactoring --disable=W0201,W0231 openfisca_core/$*
```
MattiSG, line 65: Isn't the configuration for `flake8` and `pylint` already managed in a config file? If not, this is acceptable for me as RFC #1040 says source of truth should be the task manager.

Makefile
```diff
@@ -35,26 +44,48 @@ clean: \
...
+
+## Run static type checkers for type errors.
+check-types: \
+	check-types-all \
+	check-types-strict-commons \
+	check-types-strict-types
```
MattiSG, line 71: The difference between these targets is not clear, and the `-all` is misleading, as explained above for `check-style` :wink:

Makefile
```diff
@@ -85,16 +117,29 @@ test-doc-checkout:
 	@cd doc && { \
 		git reset --hard ; \
 		git fetch --all ; \
-		[ $$(git branch --show-current) != master ] && git checkout master || : ; \
-		[ ${branch} != "master" ] \
+		[ "$$(git branch --show-current)" != "master" ] && git checkout master || : ; \
+		[ "${branch}" != "master" ] \
```
MattiSG, line 121: > I am not sure I understand the added value of quotes when all these values can by definition contain no space.

openfisca_core/commons/__init__.py
```diff
@@ -1,28 +1,71 @@
...
+
+Note:
+    How imports are being used today::
+
+        from openfisca_core.commons import *  # Bad
+        from openfisca_core.commons.formulas import switch  # Bad
+        from openfisca_core.commons.decorators import deprecated  # Bad
+
+
+    The previous examples provoke cyclic dependency problems, that prevents us
```
MattiSG, line 33:
```suggestion
    The previous examples provoke cyclic dependency problems, that prevent us
```

openfisca_core/commons/__init__.py
```diff
@@ -1,28 +1,71 @@
...
+    from modularizing the different components of the library, so as to make
```
MattiSG, line 34:
```suggestion
    from modularizing the different components of the library, which would make
```

openfisca_core/commons/__init__.py
```diff
@@ -1,28 +1,71 @@
...
+    them easier to test and to maintain.
+
+    How could them be used after the next major release::
```
MattiSG, line 37:
```suggestion
    How they could be used in a future release:
```

.circleci/config.yml
```diff
@@ -3,6 +3,8 @@ jobs:
   run_tests:
     docker:
       - image: python:3.7
+    environment:
+      TERM: dumb
```
MattiSG, line 7: Where is this used?

maukoquiroga:
Thanks @MattiSG 

> Thanks again for this PR. It adds an impressive level of details in the documentation, and seems to offer a host of features and improvements. However, this wide array of concerns makes it very hard to review and validate.
> 
> Indeed, while advertised as “improving doctests”, I understand that it:
> 
> 1. Improves the `make` UI.

That could go to another PR.

> 2. Massively improves contents of the documentation itself.

Do you think it is beyond scope? I had to write it in order to understand more or less what I was doing with the doctests themselves. 

> 3. Introduces explicit and programmatic deprecation.

Not in this one, the module `Dummy` was already deprecated, I've just extracted the warning to a decorator for reuse.

> 4. Adds typing checks.

You mean adding the annotations or checking them? For the latter, I can safely remove them, no worries.

> 5. Refines the linting and formatting configuration, and adds a new layer of linting with `pylint`.

Just for the documentation, as you suggested, there is no impact on code formatting.

For `pylint` I agree, it can be extracted from this PR, again it is useful for TDD.

> 6. Extracts some pieces of Core into a `commons` module, to be explicitly imported.

This PR does not modify the visibility of the already existing `commons` module.

> 7. Adds several dev and prod dependencies.

True.

> 
> Which raises the following questions:
> 
> * I am surprised that [6] doesn't warrant a major version change: as a reuser, will I not be impacted by the import syntax?

No, I took great care of not impacting the actual way of things. What's added is a rephrase of a comment that's already there, but the impact to the reuser, programatically, is zero.

I think it is a minor, or at least I worked to be sure it is.

> * Linting setup is controversial, as seen in #977 and others. Changing it in this PR looks like a knight that could easily backfire.

There is no code linting in this PR, just syntax checks and documentation linting.

> * Is adding type checks really necessary to the “doctests” and the production of automated documentation?

Yes for the automated documentation, no for the doctests.

For the former :

* They can be programatically tested
* Linters use them to see if the documentation is up-to-date
* The documentation builder uses them to produce de documentation

If you omit the actual running the type-checker, they're part of the doc.

> * How is the deprecation process expected to be handled? The proposed system in this changeset enables (and seems to push for) explicit version commitments for removal after deprecation. How would we uphold these promises?

I haven't deprecated anything in this PR, but in the following ones yes.

That is a good point, I'm ok extracting deprecations somewhere else to discuss that further —again there are none in this specific PR.

> 
> I would love to see this land in `master`, but I think we need to be very careful about keeping in check the many different areas that seem to be impacted. The consequences of merging this are, at the moment, rather large and I am concerned about the amount of work they can trigger.

If they have even a minimum amount of impact on the end user, considering we remove the lint/type config, I wouldn't merge it myself either.

> 
> Maybe the easiest would be to have a video call over this and some other similarly exciting yet large PRs that have been opened recently 🙂

Absolutely, I have been very careful at basically two things:

* Making the actual public contracts explicit
* Rearranging things so to make doctests actually possible
* Making it easier for documentation and doctests to be checked automatically —at some point
* Keeping all existing contracts there, adding if need be or correcting bugs, but avoiding any breaking-change

The whole `[X/10]` series should go without any major.

And that's the whole point of the modularisation work we did before: you can move things inside while having zero impact on the end user, and the other modules that depend on this one.

Makefile
maukoquiroga: Well, this are transitional for me, indeed I can move them to a specific task just for the maintainers, following consensus of #1040. At least, it is not —yet— my intention to impose them, even if those rules are a subset of what's already on country-template.

Makefile
maukoquiroga: It is a bug actually, somehow some shells like mine will complain if no quotes are added —just one more reason to move this to Python or something more cross-platform.

I can extract this to another PR, it is standalone.

.circleci/config.yml
maukoquiroga: This is for `tput` (make colors).

Makefile
maukoquiroga: Yes, I agree, what I propose is a specific task for maintainers, as discussed. I'll do that.

Makefile
maukoquiroga: I tried to avoid changing names, but I agree.

The idea behind is, for all checks (python, types, lint):

* Run the baseline
* Run a subset of more strict rules while I move module to module
* Be sure there is no inconsistency (the all should always be more relaxed that the stricts, but always a superset).

Makefile
maukoquiroga: I'm not very proud myself, but it works! Sure.

maukoquiroga:
I forgot to mention the « heuristics » of this work and those who follow:

1. The objective is and rests making the doc and the doctests a guarantee of what we promise to the user
2. Or, big or, it is an inductive work, because what's written there 50 % of the time is not the actual signature of things
3. It doesn't help either that the majority of the existing tests are integration ones

So how do I try to minimise the possible unintended breaking changes of these change sets ?

1. First and foremost: typing.

I use tools that will evaluate the code during a run tests —I use France— and will propose you type annotations based on how code is actually used, at least in the tests. 

It is not perfect, but without that I couldn't do anything else. And types are used also for several other things, like generating the documentation.

2. When possible, correcting the existing doctests.

Recycling any unit tests if there are. And adding more, to cover some edge cases. All that would be very hard to do without type-linting, as it warns you each time you're changing the input/output of anything.

3. Structural sub-typing.

You'll see more of that in the next PR, and I'd be happy to host a video conf to tell you why it is important, but TL ; DR it allows to have "static" relationships between the the different modules, without actually having to mix them if not needed.

(That is what I usually refer to as cyclic imports).

That's only possible with types, check #1034 

4. Contract / property-based testing.

For the more critical components, and more complex, types won't suffice. Periods for example, all doctests are commented out since 2014, so they do not correspond to the reality anymore.

For that cases, additional to typing, I've had to have resort to property-based testing, as you can see in #1043 

5. Finally, subpackaging, or reinforcement of public interfaces between modules.

What I propose for now is just to build them, while keeping backwards-compatibility in the meantime. 

This is finally the most important step, and required for doctests: as long as modules can't be tested in isolation, it is very hard to write doctests, which are actually just unit tests.

---

So 1., 2. and 3. are sort of unavoidable, 4. and 5. could eventually be deferred. :)

Makefile
MattiSG: Nah don't worry about it, the administrative costs of extraction are probably not worth it 😉

MattiSG:
Thank you very much for these clarifications @maukoquiroga! 😃

- For the `make` UI change, I would be in favour of extracting it to another PR indeed.
- For `pylint`, I believe it is necessary to extract it and have a wider discussion, as I fear backlash in case some contributors disagree with the added constraints.
- For the deprecation, what I would like to discuss is the process of handling deprecations if we now have a way to make explicit in which version some feature will be deprecated 🙂 
- In the same fashion, I would like to discuss type annotations.

Let's find a moment to have these discussions.

Thank you once more for this impressive documentation work, the overall improvements, and the time you take explaining these changes.

maukoquiroga:
The first three points can be handled separately and asynchronously.

On the other hand, type annotations are a deal-breaker, as I wouldn't be able to feel reassured of the quality and the promise of what I'm delivering to the user.

For the latter, could you please just clarify if it is: 

1. the type annotation itself, as it is pure doc, and already adopted largely in the code, or 
2. the fact of checking them programatically?

Again I can live without 2. for a while, but I would reject this PR myself without 1.

In fact, as a maintainer, I do not remember having accepted many contributions without type annotations, so if there is a controversy about them I would rather address that promptly.

maukoquiroga:
Type-hinting is an inexpensive way of :

1. Documenting the code
2. Building the documentation
2. Finding an correcting bugs —I've spotted quite some already
3. Helping contributors/maintainers enforce contracts, with zero runtime impact to reusers
4. Finally, they're a just part of the standard core of Python, like using a list or concatenating a string

maukoquiroga:
#### TODO after last review

- [x] Extracted deprecations #1044 
- [ ] Extracted PyLint
- [ ] Extracted make improvements
- [ ] Talked about type annotations

Makefile
maukoquiroga: Changed, let me know what do you think.

Makefile
maukoquiroga: Added a comment on why these are here, TL; DR: doing otherwise would force the whole codebase to be linted, diminishing the chances of these improvements to ever be integrated.

Makefile
maukoquiroga: Fixed.

PR #1036, opened 2021-08-25
Document Makefile

maukoquiroga:
#### Non-technical changes

- Document Makefile
  - Add a comment to explain what each task does
  - Print a human friendly message when running a task

```
$ make test

[⚙] Delete builds and compiled python files...
[⚙] Compile python files to check for syntax errors...
[⚙] Run linters to check for syntax and style errors...
[⚙] Run static type checkers for type errors...
Success: no issues found in 131 source files
[⚙] Run openfisca-core tests...
...
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,34 +1,60 @@
+## Same as `make test`.
 all: test
 
-uninstall:
-	pip freeze | grep -v "^-e" | xargs pip uninstall -y
-
+## Install project dependencies.
 install:
-	pip install --upgrade pip twine wheel
-	pip install --editable .[dev] --upgrade --use-deprecated=legacy-resolver
+	@echo [⚙] Installing dependencies...
+	@pip install --upgrade pip twine wheel
+	@pip install --editable .[dev] --upgrade --use-deprecated=legacy-resolver
 
-clean:
-	rm -rf build dist
-	find . -name '*.pyc' -exec rm \{\} \;
+## Install openfisca-core for deployment and publishing.
+build: setup.py
+	@## This allows us to be sure tests are run against the packaged version
+	@## of openfisca-core, the same we put in the hands of users and reusers.
+	@echo [⚙] Building and installing locally built openfisca-core...
+	@python $? bdist_wheel
+	@find dist -name "*.whl" -exec pip install --force-reinstall {}[dev] \;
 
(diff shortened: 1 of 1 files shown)


HAEKADI (commented):
Thank you for the addition @maukoquiroga 👍 The file is much clearer now :)

I tested the commands locally and they're all functional!

HAEKADI (approved):

HAEKADI:
@maukoquiroga Should we do the same for `OpenFisca-France` and `Country-Template`? 
It might also be a good opportunity to address these [issues](https://github.com/openfisca/openfisca-france/pull/1650#issuecomment-901219387) 🤔  
cc @MattiSG

MattiSG:
Should we not deprecate the Makefile entirely and replace it with a task runner that's less obsolete? 😅

maukoquiroga:
Hello @MattiSG, I'm personally agnostic as I like make a lot.

However, I've been looking for pure Python alternatives, for these reasons:

- It's Python, so easier to maintain and test
- It's Python, so we can avoid the bash, zch, Windows limitations

There are two alternatives that look promising to me, [invoke](https://www.pyinvoke.org/) and [nox](https://nox.thea.codes/en/stable/).

Please do not hesitate to open an issue if you think it would be worth the effort 😃

MattiSG:
Thanks @maukoquiroga! “`nox` is a command-line tool that automates testing in multiple Python environments” while “Invoke is a Python task execution tool & library”, so it seems only Invoke would be appropriate, no? 🙂

maukoquiroga:
Yes, I mixed two different use cases. Take a look at it, it looks pretty awesome as a make replacement to me at least.

maukoquiroga:
> @maukoquiroga Should we do the same for `OpenFisca-France` and `Country-Template`?
> It might also be a good opportunity to address these [issues](https://github.com/openfisca/openfisca-france/pull/1650#issuecomment-901219387) 🤔
> cc @MattiSG

Sorry @HAEKADI I completed missed this comment. Taking a look at the link now :)

maukoquiroga:
I think `make` is not the best tool for dependency management. I'm not fully convinced by `poetry` but in #1015 I try to simplify things already just with `setuptools`, I'd appreciate your feedback :)

PR #1037, opened 2021-08-25
Test against packaged version

maukoquiroga:
See openfisca/country-template#59

#### Technical changes

 - Test openfisca-core against its packaged version
   - Doing so allows to prevent publishing corrupted versions
   - What we test is what the user gets when installing openfisca-core


Makefile
```diff
@@ -1,12 +1,22 @@
...
+	@## of openfisca-core, the same we put in the hands of users and reusers.
+	pip install --upgrade pip wheel
+	python $< bdist_wheel
+	find dist -name "*.whl" -exec pip install --force-reinstall {}[dev] \;
+
+install: install.tools
 	pip install --editable .[dev] --upgrade --use-deprecated=legacy-resolver
 
+uninstall:
+	pip freeze | grep -v "^-e" | sed "s/@.*//" | xargs pip uninstall -y
```
MattiSG, line 18: Why is this changed when it is not used anywhere? 🤔

Makefile
```diff
@@ -1,12 +1,22 @@
 all: test
 
-uninstall:
-	pip freeze | grep -v "^-e" | xargs pip uninstall -y
-
-install:
+install.tools:
```
MattiSG, line 3: The naming convention for Make targets [seems to be](https://stackoverflow.com/questions/32130664/is-there-a-naming-convention-for-makefile-targets-and-variables) using hyphens instead of dots.

Makefile
```diff
@@ -1,12 +1,22 @@
...
 	pip install --upgrade pip twine wheel
+
+## Install openfisca-core for deployment and publishing.
+install.build: setup.py clean install.tools
+	@## This allows us to be sure tests are run against the packaged version
+	@## of openfisca-core, the same we put in the hands of users and reusers.
+	pip install --upgrade pip wheel
+	python $< bdist_wheel
```
MattiSG, line 11: Using [`$<`](https://stackoverflow.com/questions/3220277/what-do-the-makefile-symbols-and-mean) costs making this list of instructions ununderstandable to anyone who doesn't know the `make` special shortcuts. What do we gain from using it instead of repeating `setup.py`? 😉

Makefile
```diff
@@ -1,12 +1,22 @@
...
-
-install:
+install.tools:
 	pip install --upgrade pip twine wheel
+
+## Install openfisca-core for deployment and publishing.
+install.build: setup.py clean install.tools
+	@## This allows us to be sure tests are run against the packaged version
+	@## of openfisca-core, the same we put in the hands of users and reusers.
+	pip install --upgrade pip wheel
```
MattiSG, line 10: This seems to repeat `install.tools` while that target is a dependency of this one. What do we gain from splitting `install` in two if in the end `install.tools` is a dependency of both `install.build` and `install` and repeated in `install.build`?

Makefile
```diff
@@ -1,12 +1,22 @@
...
+	python $< bdist_wheel
+	find dist -name "*.whl" -exec pip install --force-reinstall {}[dev] \;
+
+install: install.tools
```
MattiSG, line 14: I had initially understood `install` as being for “usage” only, but as it depends on `install.tools`, I understand that running `make install` will also install [`twine`](https://pypi.org/project/twine/), which is “a utility for publishing Python packages on PyPI”, which seems appropriate for the `install.build` target considering its documentation, not for `install`.

.circleci/config.yml
```diff
@@ -20,15 +20,20 @@ jobs:
...
             # pip install --editable git+https://github.com/openfisca/extension-template.git@BRANCH_NAME#egg=OpenFisca-Extension-Template  # use a specific branch of OpenFisca-Extension-Template
 
       - save_cache:
-          key: v1-py3-{{ .Branch }}-{{ checksum "setup.py" }}
+          key: v1-py3-deps-{{ .Branch }}-{{ checksum "setup.py" }}
           paths:
             - /tmp/venv/openfisca_core
 
+      - save_cache:
+          key: v1-py3-build-{{ .Revision }}
```
MattiSG, line 33: > This is going to be costly in terms of object storage, as we are caching new artifacts for every commit. I am wondering if it would not be less costly in terms of resources to re-run `build` at the deploy step, because that step is triggered on such a small proportion of overall builds that it might outbalance the gain from caching.

Makefile
```diff
@@ -1,12 +1,22 @@
...
+install.tools:
 	pip install --upgrade pip twine wheel
+
+## Install openfisca-core for deployment and publishing.
+install.build: setup.py clean install.tools
+	@## This allows us to be sure tests are run against the packaged version
+	@## of openfisca-core, the same we put in the hands of users and reusers.
+	pip install --upgrade pip wheel
+	python $< bdist_wheel
+	find dist -name "*.whl" -exec pip install --force-reinstall {}[dev] \;
```
MattiSG, line 12: Why do we reinstall with `dev` dependencies if this is supposed to be the “packaged” version? 🤔

MattiSG: Wouldn't the `find dist -name "*.whl"` step more appropriately expressed as `find dist -name "*.whl" -exec rm` in the `clean` target, and `pip install` run after the `clean` when we want to publish?

MattiSG (commented):
Thanks @maukoquiroga! I definitely see the value of testing against the packaged version.

However, I am not 100% of the implementation. I understand it mirrors the one in OpenFisca-France, which already seemed very convoluted to me.

I have many questions, which I added as comments here. Until they are addressed, I am not even certain I can request specific changes.

In terms of implementation, I am a bit concerned that this effort seems to overlap with #1030, #1031 and https://github.com/openfisca/openfisca-france/pull/1663, but I don't see good reason to hold from improving the situation even just for a few days. As discussed yesterday over video, I will try to consolidate a vision for where different steps should be described (CI config vs makefile vs OpenFisca CLI).

Makefile
maukoquiroga: I use it a lot: for example to try this changeset out, you have to repeatedly do a clean slate so as to avoid false-positives. This is one of the problems that a tool like `nox` adresses quite gracefully.

It could be great if you added a bit of doc for contributors on how to "troubleshoot".

Makefile
maukoquiroga: Yes, I agree, #1015 adresses this already.

Makefile
maukoquiroga: Semantically does not make a lot of sense, thanks for the feedback.

.circleci/config.yml
maukoquiroga: Indeed.

Makefile
maukoquiroga: Because we're doing it wrong at the outset: we're using `pip install .[x]` to install dependencies, which is actually to install `openfisca core` **and** the dependencies. Those need to be separate steps.

Makefile
MattiSG: Sorry, I wasn't clear: what I'm wondering about is why there was a change in this line, with the added `sed`, I was not questioning the usefulness of this target 🙂

maukoquiroga: Ah, yes, I added a documentation to the task (it is to also uninstall editable versions of other packages, otherwise the task fails).

Makefile
maukoquiroga: Fixed.

Makefile
maukoquiroga: Fixed.

Makefile
maukoquiroga: Fixed it, replaced with a more proper -yet not perfect- way of doing it, in the absence of requirement files:

```
python setup.py egg_info  # Create a requirements file.
pip install `grep -v "^\[" *.egg-info/requires.txt`  # Install openfisca's dependencies.
```

Note that this is going to install `openfisca-core` via `country-template` and `extension-template`, so we're forced to reinstall `openfisca-core`.

But, we will reinstall **only** `openfisca-core`:

```
python setup.py bdist_wheel
find dist -name "*.whl" -exec pip install --force-reinstall --no-dependencies {} \;
```

.circleci/config.yml
```diff
@@ -17,21 +17,34 @@ jobs:
             python -m venv /tmp/venv/openfisca_core
             echo "source /tmp/venv/openfisca_core/bin/activate" >> $BASH_ENV
 
+      - run:
+          name: Install base utils
+          command: pip install --upgrade pip setuptools wheel
```
MattiSG, line 22: Since #1040, this should be a `make` target.

.circleci/config.yml
```diff
@@ -17,21 +17,34 @@ jobs:
...
           paths:
             - /tmp/venv/openfisca_core
 
       - run:
-          name: Run Core tests
-          command: env PYTEST_ADDOPTS="--exitfirst" make test
+          name: Build & install openfisca-core
+          command: |
+            python setup.py bdist_wheel
+            find dist -name "*.whl" -exec pip install --force-reinstall --no-dependencies {} \;  # https://github.com/openfisca/openfisca-core/pull/1015#discussion_r687051988
```
MattiSG, line 41: Since #1040, this should be a `make` target.

.circleci/config.yml
```diff
@@ -17,21 +17,34 @@ jobs:
...
+
+      - run:
+          name: Run openfisca-core tests
+          command: |
+            make clean check-syntax-errors check-style check-types
+            PYTEST_ADDOPTS="$PYTEST_ADDOPTS --cov=openfisca_core --exitfirst" pytest --pyargs openfisca_core.tests openfisca_web_api.tests  # Run the tests on the built & installed version of openfisca.
```
MattiSG, line 47: Why do we use `pytest` instead of calling the `make` target?
Since #1040, this should be a `make` target.

conftest.py
```diff
@@ -1,6 +0,0 @@
-pytest_plugins = [
-    "tests.fixtures.appclient",
-    "tests.fixtures.entities",
-    "tests.fixtures.simulations",
-    "tests.fixtures.taxbenefitsystems",
```
MattiSG, line 5: I am not sure I understand why this was deleted.

.circleci/config.yml
```diff
@@ -17,21 +17,34 @@ jobs:
...
+      - run:
+          name: Install base utils
+          command: pip install --upgrade pip setuptools wheel
+
       - run:
           name: Install dependencies
           command: |
-            make install
+            python setup.py egg_info  # Create a requirements file.
+            pip install `grep -v "^\[" *.egg-info/requires.txt`  # Install openfisca's dependencies.
```
MattiSG, line 28: Since #1040, this should be a `make` target.

MattiSG (changes requested):
Thank you very much for having answered my questions, and thank you for having refined this PR! It seems to grow, which is good in terms of features but bad in terms of review duration 😅 

The current implementation goes against RFC #1040, as it adds many specific commands to the CI. It seems necessary to now converge to having everything in the task manager (which is currently `make`, for better or for worse). This will enable testing the packaged version and publishing from any machine that can execute the Makefile. The CI should only call it.

I am not sure I understand why this changeset:

1. Moves all test files and splits them between `core` and `web_api`.
2. Changes some type hints.

maukoquiroga:
> Thank you very much for having answered my questions, and thank you for having refined this PR! It seems to grow, which is good in terms of features but bad in terms of review duration 😅
> 
> The current implementation goes against RFC #1040, as it adds many specific commands to the CI. It seems necessary to now converge to having everything in the task manager (which is currently `make`, for better or for worse). This will enable testing the packaged version and publishing from any machine that can execute the Makefile. The CI should only call it.
> 
> I am not sure I understand why this changeset:
> 
> 2. Changes some type hints.

Yes, this predates that conversation, I'll modify the things on that sense.

> 1. Moves all test files and splits them between `core` and `web_api`.

Because, currently, the tests module is not being packaged, hence there is no possible way to run the tests against the built package —in a way the user would do it, which is the whole idea of the thing.

What ensues is indeed a more standard and opinionated way of structuring a package, as country-template already does.

> 2. Changes some type hints.

Me neither: mypy is supposed to ignore the tests module, it can be safely removed.

conftest.py
maukoquiroga: Because otherwise we can't predictably test the built packages in an isolated way.

PyTest decided to forbid users to have more than one conftest per project. I don't know why they did that, but it forces us to remove it if we want isolation as we actually ship two packages independently, core and web-api.

One alternative is to ship all together, which makes a bit more sense.

Today we have:

```
openfisca_core
openfisca_web_api
tests (not shipped)
```

With this PR we will have:

```
openfisca_core
openfisca_core.tests
etc.
```

In order for this file to be kept, we would have to do something like

```
src/openfisca_core
str/openfisca_web_api
```

It is my preferred approach, but it seemed too much of a change.

.circleci/config.yml
maukoquiroga: Sure, this is outdated.

MattiSG:
Thank you for these clarifications. Let's have a synchronous discussion on the only outstanding issue: packaging the tests 🙂

PR #1043, opened 2021-09-16
 [7/17] Improve periods module doctests, documentation, and type hints

maukoquiroga:
Part of #1061


openfisca_core/periods/date_unit.py
```diff
@@ -0,0 +1,259 @@
+from __future__ import annotations
+
```
maukoquiroga, line 2: @benjello My idea to abstract date units from periods, so to facilitate their extension (weeks and so on), based on indexed enums. What do you think?

openfisca_core/periods/date_unit.py
```diff
@@ -0,0 +1,259 @@
...
+        >>> DateUnit.DAY.name
+        'DAY'
+
+        >>> DateUnit.DAY.value
+        'day'
+
+    .. versionadded:: 35.9.0
+
+    """
```
maukoquiroga, line 184: This has two benefits:

1. We consolidate the definition of date units in just one place, instead of using literals everywhere.

2. We can, at some point, use `DateUnit.Day.index` instead of `day` or `DateUnit.value`. So to be able to pass from:

```py
if period.unit == "eternity"
period.unit in ["eternirty", ...]
```

To :

```py
allowed_periods = DateUnit.DAY | DateUnit.MONTH

definition_period = [0, 1, 0, 0]

isperiod = definition_period[allowed_periods]
```

maukoquiroga:
Superseded by #1138 and #1139

Issue #1040, opened 2021-09-03
Consolidate build workflow


MattiSG:
# RFC: Consolidate build workflow

## Context

There are currently four sources of truth for how to build, test and deploy an OpenFisca package:

1. CI configuration (`.circleci/config.yml`, soon `.github/workflows/config.yml`, cf. #1030).
2. Shell scripts run in CI but that can also be run locally (`.circleci/*.sh`, soon `.github/*.sh`).
3. `Makefile`.
4. OpenFisca CLI.

Depending on each case, these complement (installing locally, testing both locally and in CI, packaging in CI), reference (`publish-git-tag.sh` from CI config, `openfisca test` in `test-api.sh`) or overlap (`make check-style` vs `.circleci/lint-changed-python-tests.sh`) each other.

## Problem

This setup raises the following issues:

1. Inconsistent workflows (`make install` differs from the install result in CI, which calls `make build` instead), leading to inconsistent test results (e.g. https://github.com/openfisca/openfisca-core/pull/1037).
2. Inconsistent contributor practices (with maintainers more often using `make test` and contributors `openfisca test`), leading to blocked PRs (e.g. https://github.com/openfisca/openfisca-core/pull/1021#issuecomment-852404640).
3. Duplication of work (e.g. https://github.com/openfisca/openfisca-france/pull/1649).
4. Maintainability: where to update the workflow when it changes? This uncertainty leads to concurrent, potentially incompatible pull requests (e.g. #1030, #1031, #1037…).
5. Portability: shell scripts and Makefile have limited portability, and CI configuration is vendor-dependent and most often executable only on CI.

## Discussion

A synchronous discussion with @benjello @benoit-cty @sandcha @maukoquiroga @HAEKADI @MattiSG took place on 02/09/2021 on this topic.

Conclusions on that topic were that:

1. There is interest in converging the workflow on one single tool.
2. Maintainers present in the call preferred `make` over `openfisca` as it is shorter, especially when multiple country packages are installed in one virtualenv (`make test` vs `openfisca test --country-package openfisca_france tests`).
3. The cost of installing Docker was felt as too high, even for maintainers.
4. The representativity of the sample of participants to the discussion was too low to lead to any conclusion as to which tool to prefer.

### Known workflow steps

The following table aims at summarizing all known declared workflow steps. It was established by @MattiSG on 03/09/2021, based on reading through Core, France and Country Template `Makefile`, `.circleci/config.yml` and `.circleci/*.sh`.

[...]

PR #1046, opened 2021-09-27
Improve Make UI and org

maukoquiroga:
#### Non-technical changes

- Reorganise `Makefile` into context files (install, test, publish…)
- Colorise `make` tasks and improve messages printed to the user

##### Note

This follows #1040 and is intended to help consolidate a unified workflow.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,8 @@ jobs:
   run_tests:
     docker:
       - image: python:3.7
+    environment:
+      TERM: xterm-256color  # To colorize output of make tasks.
 
     steps:
       - checkout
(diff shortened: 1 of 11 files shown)


benjello:
@maukoquiroga : LGTM for me but I am not an expert.
Just wondering: what does `xfailed` mean openfisca-core tests output ?
```
================ 481 passed, 1 xfailed, 151 warnings in 22.07s =================
[✓] test-core:passed ❤
```

Thanks !

benjello (approved):

maukoquiroga:
> @maukoquiroga : LGTM for me but I am not an expert. Just wondering: what does `xfailed` mean openfisca-core tests output ?
> 
> ```
> ================ 481 passed, 1 xfailed, 151 warnings in 22.07s =================
> [✓] test-core:passed ❤
> ```
> 
> Thanks !

It is an "anti-test": it passes when it fails (`xfailed`). 

Inversely, it fails when it passes (`xpass`). 

If the latter were the case, the test suite would fail.

That being said, I think that should be fixed, as xfails are usually just temporary.

PR #1047, opened 2021-09-28
Unify syntax, add style checks, remove obsolete Python 2 leftovers

bonjourmauko:
Superset of openfisca/country-template#97
Follows discussion on #1033 
Supersedes #977 
Supersedes #960  
Depends on #1160 

#### Technical changes

- Document/honestify:
  - Styles as enforced and skipped in `setup.cfg`
  - Actual dependecy versions whith whom OpenFisca runs in py37-39 (note that
    this library does not run with py36, py310, numpy v1.17, etc., which is not
    product of any particular change introduced by this changeset)
- Make style checks stricter and clearer to help developers contribute:
  - Early returns are prefered for readability and performance
    - Before:
      ```py
      if 1 > n:
          var = something
      else:
          var = something_else

      return var
      ```
    - After:
      ```py
      if 1 > n:
          return something

      return something_elses
      ```
  - Normalise the use of:
    - Quotes is normalised to double-quotes
    - `@staticmethod` when `self` is unused
    - `except from` for better exception context
    - `isinstance()` as per the best practice
    - f-strings as per good practice
  - Favour:
    - `...` as a placeholder and `pass` for loops
    - Relative to absolute imports to avoid circularity
    - Import sorting as described in the [STYLEGUIDE](STYLEGUIDE.md)
  - Avoid:
    - Catch-all exceptions in favour of more specific ones
    - Import aliases (so `np` becomes `numpy`)
  - Remove:
    - Code using `exec` because dangerous and hard to maintain
  - Extract:
    - Common imports to `commons.imports`
- Most of these can be:
  - Corrected automatically running `make format-style`
  - Fixed by hand with the output of `make test` and `make style-check`

#### New features

- Rename and makes the following public (note this are not considered as
  breking changes because methods were private, so renaming them to make
  them public counts as actually adding a new publicly exposed method, while
  removing a private method doesn't count as a breaking change):
  - `Holder._to_array` to `Holder.to_array`
  - `Holder._set` to `Holder.put`
  - `Holder._memory_storage` to `Holder.memory_storage`
  - `parameters._compose_name` to `parameters.compose_name`
  - `parameters._validate_parameter` to `parameters.validate_parameter`
  - `ParameterNodeAtInstant._name` to `ParameterNodeAtInstant.name`
  - `ParameterNodeAtInstant._instant_str` to `ParameterNodeAtInstant.instant_str`
  - `ParameterNodeAtInstant._children` to `ParameterNodeAtInstant.children`
  - `Population._holders` to `Population.holders`
  - `simulations._get_person_count` to `simulations.get_person_count`
  - `Simulation._check_for_cycle` to `Simulation.check_for_cycle`
  - `Simulation._check_period_consistency` to `Simulation.check_period_consistency`
  - `FlatTracer._enter_calculation` to `FlatTracer.enter_calculation`
  - `FlatTracer._exit_calculation` to `FlatTracer.exit_calculation`
  - `FlatTracer._start_time` to `FlatTracer.start_time`
  - `FlatTracer._end_time` to `FlatTracer.end_time`
  - `PerformanceLog._json` to `PerformanceLog.json`
- Add:
 - `periods.year_start`
 - `periods.year_end`
- Extend:
  - `test_runner.YamlItem.repr_failure` with a third argument `style` as per
    the definition of the class it inherits from.

#### Breaking changes

- Expire deprecation of:
  - The `Dummy` class, now provided by `commons`
  - `formula_helpers`, now provided by `commons`
  - `memory_config`, now provided by `experimental`
  - `rates`, now provided by `commons`
  - `Scale`, now provided by `ParameterScale`
  - `Bracket`, now provided by `ParameterScaleBracket`
  - `ValuesHistory`, now provided by `ParameterScale`
  - `ParameterNotFound`, now provided by `ParameterNotFoundError`
  - `VariableNameConflict`, now provided by `VariableNameConflictError`
  - `VariableNotFound`, now provided by `VariableNotFoundError`
  - `simulation_builder`, now provided by `simulations`
  - `openfisca-run-test`, now provided by `openfisca test`
  - `TaxBenefitSystem.new_scenario`, now provided by `SimulationBuilder`
  - `AbstractRateTaxScale`, now provided by `RateTaxScaleLike`
  - `AbstractTaxScale`, now provided by `TaxScaleLike`
  - _In fine_, All errors that were not but are not under `errors`
- Delete (no replacement):
  - `scripts.measure_performance`
  - `scripts.migrations.v16_2_to_v17`
- Normalise the use of:
  - `file_path` for file context-managers (implies argument name rename)

#### Publising instructions

Because this library has a circular depedency with `openfisca-country-template`
and `openfisca-extension-template`, publish of this version requires:
1. Publishing of one or many release candidates
2. Publishing next major or `openfisca-country-template`
3. Publishing next major or `openfisca-extension-template`
4. Publishing next major of this library
5. Publishing due diligence or unexpected outcome fixes as a release candidates


benjello:
Hi @maukoquiroga : I am not very qualified to review all this work which BTW seems  very substantial and valuable. 

I am just wondering why you changed may methods `_foo` to `foo` or even `foo_` ?

maukoquiroga:
> Hi @maukoquiroga : I am not very qualified to review all this work which BTW seems very substantial and valuable.

I'd like to have your opinion indeed, the idea would be to remove as many elements of style and keep those that are more tied to syntax and less controversial.

> I am just wondering why you changed may methods `_foo` to `foo` or even `foo_` ?

That's a convention thing, he's the rationale, and why it may or not make sense for us:

Normally when you have `_foo` it means that a method is for internal use, and not to be called on externally. That's why usually documentation builders ignore them by default, and linters are going to warn you if you call them outside the scope.

So what I did was not to rename them, but to _open_ them: because they're already being used as public methods, I made them public, while keeping the `_` for backwards-compatibility.

Where it may not make sense for us is that we reorganised modules in a very modular way, and then we could for example argue that all methods marked as `_foo` inside the `parameters` module can be used inside, but not outside, etc.

Regarding `foo_`: that is just because `_set`, to be public, becomes `set`, but `set` is a builtin function, so it becomes `set_`. 

That's an example of a pure style thing we could choose to ignore, as we do for variables for example, see https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles.

openfisca_core/parameters/helpers.py
```diff
@@ -33,40 +33,47 @@ def _compose_name(path, child_name = None, item_name = None):
...
-        return '{}.{}'.format(path, child_name)
+        return f'{path}.{child_name}'
     if item_name is not None:
-        return '{}[{}]'.format(path, item_name)
+        return f'{path}[{item_name}]'
+
+    return None
+
+
+compose_name = _compose_name
```
MattiSG, line 43: To remove in a major.

openfisca_core/parameters/helpers.py
```diff
@@ -97,3 +104,6 @@ def _validate_parameter(parameter, data, data_type = None, allowed_keys = None):
                     .format(key, parameter.name, list(allowed_keys)),
                     parameter.file_path
                     )
+
+
+validate_parameter = _validate_parameter
```
MattiSG, line 109: To remove in a major.

openfisca_core/parameters/parameter_node_at_instant.py
```diff
@@ -13,6 +13,18 @@ class ParameterNodeAtInstant:
...
+    def children(self):
+        return self._children
+
+    @property
+    def instant_str(self):
+        return self._instant_str
+
+    @property
+    def name(self):
+        return self._name
```
MattiSG, line 26: New properties should be at least documented in the CHANGELOG.

openfisca_core/parameters/values_history.py
```diff
@@ -5,5 +5,3 @@ class ValuesHistory(Parameter):
     """
     Only for backward compatibility.
```
MattiSG, line 6: To remove in a major.

openfisca_core/parameters/vectorial_parameter_node_at_instant.py
```diff
@@ -12,15 +12,23 @@ class VectorialParameterNodeAtInstant:
     Vectorized parameters allow requests such as parameters.housing_benefit[zipcode], where zipcode is a vector
     """
 
+    @property
+    def instant_str(self):
+        return self._instant_str
+
+    @property
+    def name(self):
+        return self._name
```
MattiSG, line 21: Document new properties.

openfisca_core/periods/helpers.py
```diff
@@ -201,3 +203,29 @@ def unit_weights():
 
 def unit_weight(unit):
     return unit_weights()[unit]
+
+
+def year_start(year: int, month: int, day: int) -> Tuple[int, ...]:
```
MattiSG, line 208: Document new helper in CHANGELOG.

openfisca_core/periods/helpers.py
```diff
@@ -201,3 +203,29 @@ def unit_weights():
...
+        year += 1
+        month = 1
+
+    day -= last_day
+    last_day = calendar.monthrange(year, month)[1]
+
+    return year, month, day, last_day
+
+
+def year_end(year: int, month: int, day: int) -> Tuple[int, ...]:
```
MattiSG, line 222: Document new helper in CHANGELOG.

openfisca_core/populations/population.py
```diff
@@ -9,17 +9,26 @@
...
 
 class Population:
+
+    @property
+    def holders(self):
+        return self._holders
+
+    @holders.setter
+    def holders(self, value):
+        self._holders = value
```
MattiSG, line 19: Document in CHANGELOG.

openfisca_core/rates.py
```diff
@@ -6,4 +6,4 @@
 # The following are transitional imports to ensure non-breaking changes.
 # Could be deprecated in the next major release.
 
-from openfisca_core.commons import average_rate, marginal_rate  # noqa: F401
+from openfisca_core.commons import average_rate, marginal_rate  # noqa: F401 # pylint: disable=unused-import
```
MattiSG, line 9: To remove in a major.

openfisca_core/scripts/find_placeholders.py
```diff
@@ -1,24 +1,32 @@
...
 import sys
+import warnings
+
+try:
+    from bs4 import BeautifulSoup
 
-from bs4 import BeautifulSoup
+except ImportError:
+    message = [
+        "You tried to use the 'bs4' package, or it is not currently",
```
MattiSG, line 13:
```suggestion
        "You tried to use the 'bs4' package, but it is not currently",
```

openfisca_core/scripts/measure_performances.py
```diff
@@ -1,225 +0,0 @@
-#! /usr/bin/env python
-# -*- coding: utf-8 -*-
-# flake8: noqa T001
-
-
-"""Measure performances of a basic tax-benefit system to compare to other OpenFisca implementations."""
```
MattiSG, line 6: @benjello can you confirm this is not used anymore?

openfisca_core/scripts/measure_performances_fancy_indexing.py
```diff
@@ -1,25 +1,37 @@
...
+import warnings
 
+import numpy
+
+try:
+    from openfisca_france import CountryTaxBenefitSystem
+
+except ImportError:
+    message = [
+        "You tried to use the 'openfisca_france' package, or it is not",
```
MattiSG, line 13:
```suggestion
        "You tried to use the 'openfisca_france' package, but it is not",
```

openfisca_core/scripts/migrations/v24_to_25.py
```diff
@@ -1,15 +1,23 @@
...
+import warnings
+
+try:
+    from ruamel.yaml import YAML
+    from ruamel.yaml.comments import CommentedSeq
 
-from ruamel.yaml.comments import CommentedSeq
+except ImportError:
+    message = [
+        "You tried to use the 'ruamel' package, or it is not",
```
MattiSG, line 14:
```suggestion
        "You tried to use the 'ruamel' package, but it is not",
```

openfisca_core/simulations/helpers.py
```diff
@@ -41,5 +41,8 @@ def _get_person_count(input_dict):
             return 1
 
         return len(first_value)
-    except Exception:
+    except (StopIteration, TypeError):
         return 1
+
+
+get_person_count = _get_person_count
```
MattiSG, line 48: To remove in a major.

openfisca_core/simulations/simulation.py
```diff
@@ -284,13 +286,15 @@ def _check_for_cycle(self, variable: str, period):
...
-            raise CycleError("Circular definition detected on formula {}@{}".format(variable, period))
+            raise CycleError(f"Circular definition detected on formula {variable}@{period}")
         spiral = len(previous_periods) >= self.max_spiral_loops
         if spiral:
             self.invalidate_spiral_variables(variable)
-            message = "Quasicircular definition detected on formula {}@{} involving {}".format(variable, period, self.tracer.stack)
+            message = f"Quasicircular definition detected on formula {variable}@{period} involving {self.tracer.stack}"
             raise SpiralError(message, variable)
 
+    check_for_cycle = _check_for_cycle
```
MattiSG, line 296: To remove in a major.

openfisca_core/tools/test_runner.py
```diff
@@ -75,25 +57,28 @@ def run_tests(tax_benefit_system, paths, options = None):
     if isinstance(paths, str):
         paths = [paths]
 
-    return pytest.main([*argv, *paths] if True else paths, plugins = [OpenFiscaPlugin(tax_benefit_system, options)])
+    if options is None:
+        options = {}
```
MattiSG, line 61: This does not make sense in these lines, it should be moved before the first call to `options.get`.

setup.cfg
```diff
@@ -12,17 +11,40 @@
...
 include-in-doctest  = openfisca_core/commons openfisca_core/types
 rst-directives      = attribute, deprecated, seealso, versionadded, versionchanged
 rst-roles           = any, attr, class, exc, func, meth, obj
 strictness          = short
 
+; Enabled checks are almost exclusively about syntax rather than style. These
+; checks are a subset of what is currently agreed upon by the community, and
+; are meant to improve the quality of the codebase progressively.
+;
+; C0103:        	We snake case variables and reforms.
```
MattiSG, line 25: > What about moving each of those comments on a line next to the value they document? 🙂

setup.py
```diff
@@ -44,11 +44,13 @@
     'openfisca-extension-template >= 1.2.0rc0, < 2.0.0',
     'pycodestyle >= 2.8.0, < 2.9.0',
     'pylint == 2.10.2',
+    'pylint-pytest == 1.1.2',
+    'pyupgrade == 2.28.0',
```
MattiSG, line 48: Do not pin versions.

tests/core/test_yaml.py
```diff
@@ -1,7 +1,9 @@
 import os
 import subprocess
 
+import pkg_resources
```
MattiSG, line 4: Why?

tests/core/tools/test_runner/test_yaml_runner.py
```diff
@@ -14,71 +14,79 @@
...
 
     def __init__(self):
         self.config = None
+        self.nodeid = 'testname'
         self.session = None
-        self._nodeid = 'testname'
 
 
-class TestItem(YamlItem):
+class MyItem(YamlItem):
```
MattiSG, line 74: I understand we have a naming conflict with the `pytest` API. However, these new names are not satisfactory.
What about `Stub` or `Fixture`? 🙂

tests/fixtures/entities.py
```diff
@@ -2,22 +2,22 @@
...
 from openfisca_core.entities import Entity, GroupEntity
 
-from .variables import TestVariable
+from .variables import MyVariable
 
 
-class TestEntity(Entity):
-    def get_variable(self, variable_name):
-        result = TestVariable(self)
+class MyEntity(Entity):
```
MattiSG, line 8: I understand we have a naming conflict with the `pytest` API. However, these new names are not satisfactory.
What about `Stub` or `Fixture`? 🙂

tests/fixtures/variables.py
```diff
@@ -2,7 +2,7 @@
 from openfisca_core.variables import Variable
 
 
-class TestVariable(Variable):
+class MyVariable(Variable):
```
MattiSG, line 5: I understand we have a naming conflict with the `pytest` API. However, these new names are not satisfactory.
What about `Stub` or `Fixture`? 🙂

tests/web_api/test_calculate.py
```diff
@@ -1,23 +1,24 @@
...
 
-def post_json(client, data = None, file = None):
+def post_json(test_client, data = None, file = None):
     if file:
         file_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'assets', file)
-        with open(file_path, 'r') as file:
-            data = file.read()
-    return client.post('/calculate', data = data, content_type = 'application/json')
+        with open(file_path, encoding = "utf-8") as f:
+            data = f.read()
```
MattiSG, line 16: > `file` was better 😉

tests/core/test_simulation_builder.py
```diff
@@ -45,18 +46,20 @@ def __init__(self):
 @pytest.fixture
 def enum_variable():
 
+    class MyEnum(Enum):
```
MattiSG, line 49: I understand we have a naming conflict with the `pytest` API. However, these new names are not satisfactory.
What about `Stub` or `Fixture`? 🙂

MattiSG (changes requested):
Very impressive and long overdue cleanup work 👏  Congratulations and thanks @maukoquiroga for doing a part of the Augeas cleaning.

Document in CHANGELOG:

- Removal of obsolete migration scripts.
- Usage of string formatting `f`.
- Removal of Python 2 legacy pragmas.
- Systematic usage of early exits.
- Richer stack traces (through `from error`).

_Disclaimer: considering the size of this PR, we paired with @maukoquiroga and he answered each question I had in person, so there is little trace of the discussions here. This would not have been doable asynchronously._

openfisca_core/indexed_enums/enum_array.py
```diff
@@ -10,12 +10,30 @@
...
+        ...     OWNER = "Owner"
+        ...     TENANT = "Tenant"
+        ...     FREE_LODGER = "Free lodger"
+        ...     HOMELESS = "Homeless"
+
+        >>> array = numpy.array([1])
+        >>> enum_array = EnumArray(array, Housing)
+
+        >>> enum_array.possible_values
+        <enum 'Housing'>
```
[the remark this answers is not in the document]
maukoquiroga: There's already a PR for this actually.

openfisca_core/tools/test_runner.py
```diff
@@ -230,9 +216,9 @@ def should_ignore_variable(self, variable_name):
 
         return variable_ignored or variable_not_tested
 
-    def repr_failure(self, excinfo):
+    def repr_failure(self, excinfo, __arg = None):
```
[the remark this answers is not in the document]
maukoquiroga: This method is inherited from another class, where the arity differs. The linter complains, and the solution is appropriate, but cryptic.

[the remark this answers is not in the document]
maukoquiroga: Found a better solution.

tests/core/test_yaml.py
maukoquiroga: Removed.

maukoquiroga:
I think I went through all of your very helpful suggestions @MattiSG. This PR is in no way exhaustive in terms of style —for example we still. have quote use inconsistency— but it is a very good ground for further improvement.

setup.py
maukoquiroga: Even if not pinned, some constraint needs to be in place, otherwise cache will never update them, and we will end with errors like the one we have now in CI.

nikhilwoodruff (approved):
This looks like a nice set of improvements IMO @maukoquiroga.

benjello (approved):

setup.py
```diff
@@ -1,95 +1,97 @@
...
+    "pyupgrade >= 2.37.3, < 3.0.0",
     ] + api_requirements
 
 setup(
-    name = 'OpenFisca-Core',
-    version = '35.9.0',
-    author = 'OpenFisca Team',
-    author_email = 'contact@openfisca.org',
+    name = "OpenFisca-Core",
+    version = "36.0.0.rc1",
```
MattiSG, line 55:
```suggestion
    version = "36.0.0-rc1",
```

MattiSG (changes requested):
This is too big for me to realistically review one more time, especially since the changeset is even bigger now. I have to trust that the initial review followed by pairing is enough 🙂 

I believe that, in this case, the usual review approach is unfit for purpose. I see that you intend to release this as `36.0.0.rc1`. Indeed, I suggest that we introduce for this case, both as a way to unblock and as an experiment for future similar cases, a prerelease system. Could we publish this as `36.0.0-rc1` (with a dash, not a dot, to abide by [SemVer§9](https://semver.org/#spec-item-9) and try it out / ask for it to be tried out on several country packages? 🙂 If no issues arise, then we'll publish as `36.0.0`! 😃 If there are issues, then we'll iterate and fix them.

maukoquiroga:
> This is too big for me to realistically review one more time, especially since the changeset is even bigger now. I have to trust that the initial review followed by pairing is enough 🙂
> 
> I believe that, in this case, the usual review approach is unfit for purpose. I see that you intend to release this as `36.0.0.rc1`. Indeed, I suggest that we introduce for this case, both as a way to unblock and as an experiment for future similar cases, a prerelease system. Could we publish this as `36.0.0-rc1` (with a dash, not a dot, to abide by [SemVer§9](https://semver.org/#spec-item-9) and try it out / ask for it to be tried out on several country packages? 🙂 If no issues arise, then we'll publish as `36.0.0`! 😃 If there are issues, then we'll iterate and fix them.

The easier for that is then to improve the tooling with `poetry` or `hatch` to make it as easy as possible to hand publish.

MattiSG:
I understand changing the dependency manager will make it easier in the future, but I would prefer we avoid putting this on the critical path of unblocking this PR, since this change will certainly yield additional decisions and issues.

maukoquiroga:
@MattiSG #1160

PR #1048, opened 2021-09-29
[1/4] [test-built] Fix yaml tests bug

maukoquiroga:
Depended upon by #1058

#### Bug fix

- In _test_yaml.py_:
  - Fix yaml tests loading —required for testing against the built version.


MattiSG (approved):
I am not sure I fully understand how this fixes testing against the cached version (I assume that is because the current system uses `pkg` resources to access fixture data), but this looks cleaner, tests still pass, and there is the same number of tests run as before 🙂

MattiSG:
Thanks @maukoquiroga! 👏

PR #1049, opened 2021-09-29
Fix slow tests

maukoquiroga:
Just for archiving.


maukoquiroga:
@MattiSG This is just a prototype I did while ago, and created a pull request for sharing it with the community, however I'm not working on it anymore —and probably won't, but I didn't this to die in my local computer 😄

PR #1059, opened 2021-10-01
[3/4] [test-built] Adapt package organisation

maukoquiroga:
#### Non-technical changes

- Move published code to `src`.
- Use `setup.cfg` for metadata.


MattiSG:
Thanks @maukoquiroga for this suggestion!

It is not very clear to us what is the aim of this change, and we worry that it would break import paths, making this a major.

Coud you please make explicit the reason for this change? 🙂

PR #1060, opened 2021-10-07
Fix make (build should be "phony")

bonjourmauko:
#### Non-technical changes

- Remove `setup.py` from `make build`.
  - Otherwise it always run.


MattiSG:
I was about to click on “approve” when this got merged. How could it be merged without a review? 😮

MattiSG:
Crazy 🤓

Issue #1061, opened 2021-10-07
Improve package's doctests


bonjourmauko:
Hi there!

I really enjoy OpenFisca, but I recently encountered an issue.

### Here is what I did:

```
pytest openfisca_core openfisca_web_api --maxfail 0 --continue-on-collection-errors --quiet
```

### Here is what I expected to happen:

```
XXX passed, X xfailed, XXX warnings in X.XXs
```

### Here is what actually happened:

```
========================================================================================= short test summary info ==========================================================================================
FAILED openfisca_core/commons/formulas.py::openfisca_core.commons.formulas.apply_thresholds
FAILED openfisca_core/commons/formulas.py::openfisca_core.commons.formulas.switch
FAILED openfisca_core/holders/holder.py::openfisca_core.holders.holder.Holder.get_memory_usage
...
ERROR openfisca_core/scripts/measure_numpy_condition_notations.py::test_switch_select
```

### Context:

Python provides `doctests` to help keep a testable documentation.

They provide some benefits:

- They help maintainers and contributors to spot bugs and improvement opportunities.
- They helps reusers to understand the code quickly
- They help separating contexts (unit, functional, ...)
- They help produce the [official doc](https://openfisca.org/doc)

### Complications:

Today the code coverage by doctests is mild, they are not run, and most of them are broken. 

Some observed consequences:

- The code is in overall hard to grasp without a consequent time investment, thus hard to contribute to
- The code is hard to test unitarily due to its current design
- There are then very few unit tests
- The documentation ~is~ was [broken](https://github.com/openfisca/openfisca-doc/issues/244)

### Proposal:

Fix, complete, and run doctests systematically!

### TODO

- [x] #1250
- [x] #1249
- [x] #1254 
- [ ] #1266
- [x] #1267
- [ ] #1268
- [ ] #1269
- [ ] #1276
- [ ] 9.  Improve `openfisca_core.projectors` doctests
- [ ] 10. Improve `openfisca_core.reforms` doctests
- [ ] 11. Improve `openfisca_core.simulation_builder` doctests
- [ ] 12. Improve `openfisca_core.simulations` doctests
- [ ] 13. Improve `openfisca_core.taxbenefitsystems` doctests
- [ ] 14. Improve `openfisca_core.taxscales` doctests
- [ ] 15. Improve `openfisca_core.tracers` doctests
- [ ] 16. Improve `openfisca_core.variables` doctests
- [ ] 17. Improve `openfisca_web_api.loader` doctests

PR #1070, opened 2021-10-17
Allow complete group encapsulation

### LINKED ISSUE #1041: Allow complete group encapsulation
In `openfisca-uk`, we have entities that are guaranteed to be completely hierachical. For example, a person can be a member of a benefit unit (nuclear family) and a household, but every person in the benefit unit will be in the same household. Variables at the benefit unit level often depend on properties of the household, and right now this is usually done via:
(for example, say `household_region` is at the household level, and `benunit_region` is at the benefit unit level)
```python
benunit_region = benunit.value_from_first_person(benunit.members.household("household_region", period))
```
I wonder if this could be simplified to:
```python
benunit_region = benunit.household("household_region", period)
```
by making the `.household` function available to group entities (e.g. benefit unit) which specify that their members will have the same `household` membership.

nikhilwoodruff:
Hope this is useful - this fixes #1041 , by adding a shortcut to a containing group entity provided. I've added a test to `test_formulas`.

Essentially, say we have three entities, person, family and household, and we know that all members of a family will be in the same household (but not necessarily vice-versa). Say we also have a variable `household_level_variable` at the household level, and we want to project the result of that variable to the contained families. Currently, I think we have to use:

```python
family_level_variable = family.value_from_first_person(family.members.household("household_level_variable", period))
```

With this, we can instead use:

```python
family_level_variable = family.household("household_level_variable", period)
```

provided, that when we defined the family entity (using either `build_entity` or the `Entity` class), we set the optional keyword argument `containing_entities = ["household"]`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 35.7.0 [#1070](https://github.com/openfisca/openfisca-core/pulls/1070)
+
+#### New Features
+
+- Add group population shortcut to containing groups entities
+
 ## 35.6.0 [#1054](https://github.com/openfisca/openfisca-core/pull/1054)
 
 #### New Features
--- a/openfisca_core/entities/group_entity.py
+++ b/openfisca_core/entities/group_entity.py
@@ -2,8 +2,25 @@ from openfisca_core.entities import Entity, Role
 
 
 class GroupEntity(Entity):
-    """
-    Represents an entity composed of several persons with different roles, on which calculations are run.
+    """Represents an entity containing several other entities with different roles.
+
+    A :class:`.GroupEntity` represents an :class:`.Entity` containing 
+    several other :class:`.Entity` with different :class:`.Role`, and on
+    which calculations can be run.
+
+    Args:
+        key: A key to identify the group entity.
+        plural: The ``key``, pluralised.
+        label: A summary description.
+        doc: A full description.
+        roles: The list of :class:`.Role` of the group entity.
+        containing_entities: The list of keys of group entities whose 
+            members are guaranteed to be a superset of this group's entities.
+
+    .. versionchanged:: 3.7.0     
+        Added ``containing_entities``, that allows the defining of group entities 
+        which entirely contain other group entities.
+
     """
 
     def __init__(self, key, plural, label, doc, roles, containing_entities = ()):
--- a/openfisca_core/entities/group_entity.py
+++ b/openfisca_core/entities/group_entity.py
@@ -6,7 +6,7 @@ class GroupEntity(Entity):
     Represents an entity composed of several persons with different roles, on which calculations are run.
     """
 
-    def __init__(self, key, plural, label, doc, roles):
+    def __init__(self, key, plural, label, doc, roles, containing_entities = []):
         super().__init__(key, plural, label, doc)
         self.roles_description = roles
         self.roles = []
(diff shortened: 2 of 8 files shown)


MattiSG (dismissed):
Thanks @nikhilwoodruff for this contribution! 😃 This seems to address the point you opened yourself in #1041.

- For value assessment: @MaxGhenis, you upvoted #1041, can you confirm this implementation would solve your concern as well? 🙂
- For vision compatibility assessment, @benjello and @sandcha, could you please comment on this minor extension of the API surface? Does it seem idiomatic enough in OpenFisca? Do you see good uses in the tax and benefit systems you maintain? 🙂
- For technical implementation: I see a good test suite. @maukoquiroga, @benjello do you have any feedback on the implementation? 🙂

On my side as a gatekeeper, my main feedback is on semver and CHANGELOG, as outlined in the comments 🙂 I would also add that some documentation on this new feature would be an important addition, to enable its discoverability. @maukoquiroga, considering #1061, would you recommend to use doctests or to use [openfisca/openfisca-doc](https://github.com/openfisca/openfisca-doc)?

> As soon as the version number and link are corrected, if I'm too long to review again, it is okay for another maintainer to dismiss my review if it is the only blocking point.

MattiSG:
Thanks @nikhilwoodruff for your fast reaction! 😃

> As a side note, regarding the version bump and in order to save you effort in the future: due to the number of contributors and the asynchronous nature of contributions on OpenFisca, it is very common that version numbers conflict. In order to minimise the administrative burden, we tend to update the version number and rebase one last time after getting review approval, as otherwise the number of upstream merges can become a bit daunting :wink:

MattiSG:
> Is there something I need to do to run the code coverage test?

No, unfortunately it seems our CI config is not properly set up to handle incoming PRs from external organisations. The token needed to pass the CI coverage step is not passed to such PRs for security reasons, but the step is still run and thus fails instead of being cancelled. This is something to be fixed in #1057.

nikhilwoodruff:
> > For value assessment: @MaxGhenis, you upvoted #1041, can you confirm this implementation would solve your concern as well? slightly_smiling_face
> 
> Yes, this is great, thanks @nikhilwoodruff. Just wanted to check if this works for enums? I don't see one in the test.

Just tried to get a MWE (I know we've had an issue in the past where projection loses the Enum decodability), but found it does work! [Here's a MWE](https://gist.github.com/nikhilwoodruff/6d4d8cb271191f863445ac0df39d9085) - I tested:
* person Enum -> household Enum
* household Enum -> person Enum
* household Enum -> family Enum (with this PR)

bonjourmauko (approved):

guillett:
Many years later, I think this PR should be reverted. To me, the code improvement is not worth the added complexity.

@nikhilwoodruff you thought that you had to use
```python
family_level_variable = family.value_from_first_person(family.members.household("household_level_variable", period))
```

so this PR allows
```python
family_level_variable = family.household("household_level_variable", period)
```

To me, without this PR, it is possible to write a not-so-complex and clearer alternative: 
```python
family_level_variable = family.first_person.household("household_level_variable", period)
```

Moreover @benjello [comment and raised concern](https://github.com/openfisca/openfisca-core/pull/1070#issuecomment-945790912) were not taken into account as no validation step was added.

guillett:
Some searches on GitHub reveal (unsurprinsingly) that this is only used in PolicyEngine

https://github.com/search?q=org%3Aopenfisca+containing_entities&type=code
https://github.com/search?q=containing_entities%3D&type=code&p=2

I will open a RFC issue suggesting to revert this PR.

PR #1081, opened 2021-11-02
Fix message bug

### LINKED ISSUE #1072: Possible misnamed attribute in error handler
Hi there!

When debugging a whole host of unrelated errors, at some point I managed to trigger the error handler in `Population.check_array_compatible_with_entity`.  However, printing that error message threw another error, that `Population` has no attribute `key`. 

I couldn't disagree with this new exception - I wonder if in [this line](https://github.com/openfisca/openfisca-core/blob/da995e477e46276c0e2627d9e0e11eb33861d2cc/openfisca_core/populations/population.py#L47), should `self.key` instead be `self.entity.key` like other occurrences? I couldn't find any assignment or other usage of `Population.key`.

nikhilwoodruff:
This should fix #1072 . I'll add version change information, but the action seemed to fail quite quickly for some reason.

#### Technical changes

- Use `self.entity.key` rather than `self.key` when generating an error message for a mis-sized population array.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### 35.7.3 [#1081](https://github.com/openfisca/openfisca-core/pull/1081)
 
-- Fix misnamed attribute in mis-sized population error message
+- Correct error message in case of mis-sized population 
 
 ### 35.7.2 [#1057](https://github.com/openfisca/openfisca-core/pull/1057)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 35.7.3 [#1081](https://github.com/openfisca/openfisca-core/pull/1081)
+
+- Fix misnamed attribute in mis-sized population error message
+
 ### 35.7.2 [#1057](https://github.com/openfisca/openfisca-core/pull/1057)
 
 #### Technical changes
--- a/openfisca_core/populations/population.py
+++ b/openfisca_core/populations/population.py
@@ -44,7 +44,7 @@ class Population:
     def check_array_compatible_with_entity(self, array):
         if not self.count == array.size:
             raise ValueError("Input {} is not a valid value for the entity {} (size = {} != {} = count)".format(
-                array, self.key, array.size, self.count))
+                array, self.entity.key, array.size, self.count))
 
     def check_period_validity(self, variable_name, period):
         if period is None:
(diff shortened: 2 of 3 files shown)


benjello (approved):

nikhilwoodruff:
Thanks @benjello for the approval - just want to make sure this won't break anything, is there a reason the CircleCI changes are refusing to run?

benjello:
@HAEKADI @MattiSG @maukoquiroga any hint about the failing CircleCi job ?

HAEKADI:
After a rebase the new GitHub Actions config should run smoothly. The CircleCI check is just a left over from the previous config and will be removed but it is not required for the merge. cc @sandcha

sandcha:
Thank you @HAEKADI! 
Build stopped on CircleCI. cc @MattiSG  
A rebase should solve the current building issue. Let me know if I can help on anything else @nikhilwoodruff :)

MattiSG (dismissed):
It seems `openfisca-core` was added as a submodule, it should not be here! 😉

MattiSG:
I triggered the build 🙂

As the CI says, the blocking point now is the version number that should be bumped.

@nikhilwoodruff can you update the `CHANGELOG` and `setup.py`? I think this is just a patch version update, so version `35.7.3` should be good 👍 
Let us know if you need any help with this! 🙂

MattiSG:
(as a side note, I suggest you `git rm src/openfisca-core` locally, as you accidentally added it as a submodule in https://github.com/openfisca/openfisca-core/pull/1081/commits/97fc9005a4dc271acfe15a94c457372ec61cca07)

nikhilwoodruff:
Oh thanks @MattiSG - not sure how that happened. I've just added an entry to the changelog and bumped the version.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+### 35.7.3 [#1081](https://github.com/openfisca/openfisca-core/pull/1081)
+
+- Fix misnamed attribute in mis-sized population error message
```
MattiSG, line 5: We try to keep the changelog user-oriented and not get too technical in it 🙂 
```suggestion
- Correct error message in case of mis-sized population 
```

MattiSG (approved):
Thanks @nikhilwoodruff! 🎉

MattiSG:
(I added a commit mostly to trigger the build, which apparently does not get triggered from third-party pull requests 🤔)

HAEKADI:
@MattiSG I opened a PR #1083 to solve the issue.

PR #1086, opened 2021-11-27
Fix Enum projection bug

nikhilwoodruff:
This PR fixes a bug where `GroupPopulation.value_nth_person(EnumArray) -> Array[int]`, instead of the expected `GroupPopulation.value_nth_person(EnumArray) -> EnumArray`. I've extended the previous test case to cover this as well, to avoid code duplication.


nikhilwoodruff:
@benjello I added a test to cover this, should I add more? I think it covers the only use case.

benjello (approved):

benjello:
@MattiSG @maukoquiroga we need an additional review here ;-)

nikhilwoodruff:
Hey @sandcha @MattiSG @maukoquiroga - if you've got a spare second and this looks OK to add, would be much appreciated as we're having to use some workarounds in the US system (string type variable copies of affected Enums) that end up causing their own bugs. Thanks in advance!

openfisca_core/populations/group_population.py
```diff
@@ -267,6 +267,10 @@ def value_nth_person(self, n, array, default = 0):
         # The map is needed b/c the order of the nth persons of each household in the persons vector is not necessarily the same than the household order.
         result[nb_persons_per_entity > n] = array[members_map][positions[members_map] == n]
 
+        # Preserve Enum dtype
+        if isinstance(array, EnumArray):
+            result = EnumArray(result, array.possible_values)
```
sandcha, line 272: Thank you for this fix that also brings some consistency with the `value_from_person(...)` treatment for enums.

openfisca_core/populations/group_population.py
```diff
@@ -267,6 +267,10 @@ def value_nth_person(self, n, array, default = 0):
         # The map is needed b/c the order of the nth persons of each household in the persons vector is not necessarily the same than the household order.
         result[nb_persons_per_entity > n] = array[members_map][positions[members_map] == n]
 
+        # Preserve Enum dtype
```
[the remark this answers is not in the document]
nikhilwoodruff: Oh yes, you're right. Thanks.

tests/core/test_projectors.py
```diff
@@ -287,7 +287,19 @@ class projected_family_level_variable(Variable):
         def formula(family, period):
             return family.household("household_level_variable", period)
 
-    system.add_variables(household_level_variable, projected_family_level_variable)
+    class decoded_projected_family_level_variable(Variable):
```
augusto-herrmann, line 290: I wonder why the class is named like this, when Python class naming conventions is to use [PascalCase](https://namingconvention.org/python/class-naming.html). But then again, I'm not familiar enough with the code of this project so I'm sure there is a perfectly valid reason for this.

nikhilwoodruff:
> @nikhilwoodruff From a "logistical" point of view, git rebase is favored in openfisca-core in order to get a linear history (easier to read etc.). Could you use git rebase instead of git merge?
> 
> > You can for example compare their impacts by looking into the branch history with `git log --pretty=format:'%h - %an : %s' --date-order --graph`.

Thanks @sandcha - I've updated the version number and changelog, so all the checks now pass. As for the rebasing, sounds good; I don't think I have permissions to merge the PR though?

nikhilwoodruff:
Re-merging after a recent PR was merged to ensure the versioning is still up-to-date here.

nikhilwoodruff:
Thanks @benjello, @sandcha and @augusto-herrmann for the help with this PR!

PR #1094, opened 2021-12-10
Support string parameter types

### LINKED ISSUE #1092: Support `str` and/or `date` parameters
Is there a technical reason why OpenFisca-Core doesn't allow string or date types in parameters? When I try to add one, it fails the validation check in the parameter initialisation. Some legislative parameters are date-based - e.g. pension changes often affect only those born after a specific date. Happy to file a PR adding this if it's not objectionable.

nikhilwoodruff:
This PR fixes #1092 by supporting string parameters. I've added a test for it, and removed an old test which would now be incorrect (testing that a string parameter fails).


benjello:
@sandcha @MattiSG @maukoquiroga @guillett : do you remember why we forbid string parameters ?

MattiSG (changes requested):
The Web API would also have to be tested for exposing such types.

sandcha:
@nikhilwoodruff Do you have all the information you need about the Web API test suggested above?

nikhilwoodruff:
> @nikhilwoodruff Do you have all the information you need about the Web API test suggested above?

Thanks @MattiSG and @sandcha - not sure I fully understand what exactly the Web API should test for, could you provide any more detail?

MattiSG:
The test you added tests the _Python API_ of OpenFisca 👍 That's a great starting point!

However, by enabling new data types to be exposed, this changeset also changes the _Web API_ (the module that exposes parameter values and calculations over RESTful HTTP). The web API is packaged in its own module (`web_api` at the root of the repository, alongside `core`). In order to ensure that the parameters are also properly exposed over the Web API, you need to also add tests for that behaviour 🙂 

This is done by editing the [`tests/web_api/test_parameters.py`](https://github.com/openfisca/openfisca-core/blob/master/tests/web_api/test_parameters.py) file. I suggest that you add a new test that would try to `get` a String value. For that, you will need to add one such parameter in the [Country Template](https://github.com/openfisca/country-template/), which will also enable discovery for new users.

Finally, of course, [updating the documentation](https://github.com/openfisca/openfisca-doc/blob/master/source/coding-the-legislation/legislation_parameters.md) would be a great final touch for users to know about this new feature 😉

guillett:
Superseeded by #1334

PR #1111, opened 2022-02-22
Test Python 3.8 and remove Python 3.6

### LINKED ISSUE #1035: Test Python 3.8
Python 3.10 will also be launched in about 6 weeks.

### LINKED ISSUE #1110: Remove Python 3.6 tags
Only Python 3.7 is tested, but setup.py and the README badge both show Python 3.6. Python 3.6 has also reached end of life. I would suggest removing references to it.

MaxGhenis:
#### Technical changes

- Test Python 3.8 (fixes #1035)
- Remove Python 3.6 support in setup.py (fixes #1110)

Creating as a draft as I'm not sure I've done the Python version matrix correctly. I'm guessing some of the jobs should not be matrixed. I also need to adjust my linter.

NB: Python 3.8.12 is the latest version of Python 3.8, per https://www.python.org/downloads.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -7,7 +7,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        # Patch version must be specified to avoid any cache confusion, since the cache key depends on the full Python version. If left unspecified, different patch versions could be allocated between jobs, and any such difference would lead to a cache not found error.
+        # Patch version must be specified to avoid any cache confusion, since
+        # the cache key depends on the full Python version.
+        # If left unspecified, different patch versions could be allocated
+        # between jobs, and any such difference would lead to a cache not
+        # found error.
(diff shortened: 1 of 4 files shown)


MaxGhenis:
numpy 1.17.0 is failing to install on Python 3.8. I don't understand why it's starting at numpy 1.19.0 to retrieve the most recent three versions, rather than 1.22.0, which is the latest numpy version and compatible with Python 3.8 (not 3.7).

@maukoquiroga I'd missed your #1031 when starting this. Could this serve as a simpler starting point to at least get 3.8 running?

MattiSG:
Thanks @MaxGhenis for this suggestion, and sorry for not having been able to collaborate on it to face the stronger complexity than it initially looked 😔
Superseded by #1168, which will hopefully provide a scalable way to test Python versions in a matrix. Looking forward to progressively bring support for newer versions of Python!

PR #1113, opened 2022-03-22
Include bracket rates and thresholds in scale descendants

nikhilwoodruff:
#### Technical changes

- `ParameterScale.get_descendants` now includes bracket rates and thresholds.

### DIFF
--- a/tests/core/tax_scales/test_tax_scale_descendants.py
+++ b/tests/core/tax_scales/test_tax_scale_descendants.py
@@ -1,0 +1,47 @@
+def test_tax_scale_descendants_are_complete():
+    """Tests that calling the `get_descendant` function of a ParameterScale returns the rates and thresholds as Parameters.
+    """
+
+    from openfisca_core.parameters import ParameterNode, Parameter
+
+    parameters = ParameterNode("parameters", data={
+        "root_node": {
+            "first_child": {
+                "scale_parameter": {
+                    "brackets": [
+                    {
+                        "rate": {
+                            "2022-01-01": 0.1,
+                        },
+                        "threshold": {
+                            "2022-01-01": 0,
+                        }
+                    },
+                    {
+                        "rate": {
+                            "2022-01-01": 0.2,
+                        },
+                        "threshold": {
+                            "2022-01-01": 100,
+                        }
+                    }
+                    ]
+                },
+                "normal_parameter": {
+                    "2022-01-01": 5
+                }
+            }
+        }
+    })
+
+    # Get descendants which are parameters (ignore nodes)
+    parameter_descendants = list(filter(lambda p: isinstance(p, Parameter), parameters.get_descendants()))
+
+    # Check that the expected names are in the descendants
+    parameter_names = list(map(lambda p: p.name, parameter_descendants))
+    assert all(name in parameter_names for name in [
+        "root_node.first_child.scale_parameter[0].rate",
+        "root_node.first_child.scale_parameter[0].threshold",
+        "root_node.first_child.scale_parameter[1].rate",
+        "root_node.first_child.scale_parameter[1].threshold",
+    ]), "ParameterScale descendants don't include bracket thresholds and rates"


nikhilwoodruff:
OK I think I got everything:
- [x] Unit test added and passed
- [x] Versioning added (patch bump)
- [x] Current tests and lint checks passing

@benjello @sandcha @MattiSG a review on this would be much appreciated, thanks!

benjello (approved):

sandcha:
Hello, Thank you for this fix! Just in case, did/could you test it on the Web API? 
> These, for example, should work: `openfisca serve` should launch the Web API without error and a GET to `/parameter/{parameterID}` should return the parameter information without error.

openfisca_core/parameters/parameter_scale.py
```diff
@@ -62,7 +62,14 @@ def __repr__(self):
...
     def get_descendants(self):
-        return iter(())
+        returned_any_results = False
+        for bracket in self.brackets:
+            for allowed_key_name in bracket._allowed_keys:
+                if hasattr(bracket, allowed_key_name):
+                    yield getattr(bracket, allowed_key_name)
+                    returned_any_results = True
+        if not returned_any_results:
+            return iter(())
```
sandcha, line 72: Do you have an example where this situation occurs? 🤔 It could be added as a test.

sandcha (changes requested):
Tested locally on the Web API and: 
* The command `openfisca serve` runs without error 🙌 
* Requesting the ParameterScale [taxes.social_security_contribution](https://github.com/openfisca/country-template/blob/3.13.2/openfisca_country_template/parameters/taxes/social_security_contribution.yaml) works and gives the same result as before this PR 🙌 
  > Request (and `jq` formatting): `curl http://127.0.0.1:5000/parameter/taxes.social_security_contribution | jq`
* But requesting all the parameters lists now scale elements as parameters which would be an issue for Web API users. For example, the list contains: 
  - `taxes.social_security_contribution[INDEX].threshold` 
  - and `taxes.social_security_contribution[INDEX].rate` 

  when, for this scale it should only contain `taxes.social_security_contribution`. 
  Fix this issue before merging this PR?
  > Request (and `jq` formatting): `curl http://127.0.0.1:5000/parameters`

---
And, later, if you both agree that this Web API issue is fixed, you can dismiss my review.

nikhilwoodruff:
> Tested locally on the Web API and:
> 
> * The command `openfisca serve` runs without error raised_hands
> * Requesting the ParameterScale [taxes.social_security_contribution](https://github.com/openfisca/country-template/blob/3.13.2/openfisca_country_template/parameters/taxes/social_security_contribution.yaml) works and gives the same result as before this PR raised_hands
>   > Request (and `jq` formatting): `curl http://127.0.0.1:5000/parameter/taxes.social_security_contribution | jq`
> * But requesting all the parameters lists now scale elements as parameters which would be an issue for Web API users. For example, the list contains `taxes.social_security_contribution[INDEX].threshold` and `taxes.social_security_contribution[INDEX].rate` when, for this scale it should only contain `taxes.social_security_contribution`. Fix this issue before merging this PR?
>   > Request (and `jq` formatting): `curl http://127.0.0.1:5000/parameters`
> 
> And, later, if you both agree that this Web API issue is fixed, you can dismiss my review.

Thanks for this review, @sandcha. On your last point, I'm not sure if I follow: this PR intends to do exactly as you say. For example, in the UK we have a progressive income tax scale, but the rates have specific names ("basic rate", "higher rate", "additional rate"). We'd like to surface these parameters without having to take them outside of their parameter scale, using `get_descendants`. What do you think?

MattiSG:
> We'd like to surface these parameters without having to take them outside of their parameter scale

If I understand correctly, in the parameter scale, they would only be accessible through their numeric index. How would exposing them this way enable you to name them? 🤔

### Alternative implementations

I understand the value in using a value both as a named one and as part of a parameter scale. The only way I know of this being implemented in the OF-FR corpus is by defining a legal constant (“plafond de la sécurité sociale”) as a parameter and expressing scales as multiples of it (`threshold_unit: PSS`). @openfisca/france-contrib do you know of other ways this has been implemented? Did you ever want to have such a feature?

### Weight

The main issue I can see with this change is how much larger the `/parameters` route data would become. Currently, the aim of listing parameters is to make them discoverable, but not to expose all of their data and structure.

@sandcha since you ran this branch locally, could you please provide us with the file size increase for a large corpus such as OF-FR? As a baseline, `curl -L -I https://api.fr.openfisca.org/latest/parameters | grep content-length` yields `904421`, i.e. 883 kB. What would be the added weight with this PR?

nikhilwoodruff:
> 

Thanks @MattiSG: 

> > We'd like to surface these parameters without having to take them outside of their parameter scale
> 
> If I understand correctly, in the parameter scale, they would only be accessible through their numeric index. How would exposing them this way enable you to name them? thinking

Ah, I should mention we've added a `name` attribute to the parameter metadata for PolicyEngine, which overrides the default naming convention when it's served over the OpenFisca-US/UK API ([here's the basic rate example](https://github.com/PolicyEngine/openfisca-uk/blob/9af64745506817c7fc485a6e669dcd1f967d6972/openfisca_uk/parameters/tax/income_tax/rates/uk.yaml#L16)).

> 
> ### Alternative implementations
> I understand the value in using a value both as a named one and as part of a parameter scale. The only way I know of this being implemented in the OF-FR corpus is by defining a legal constant (“plafond de la sécurité sociale”) as a parameter and expressing scales as multiples of it (`threshold_unit: PSS`). @openfisca/france-contrib do you know of other ways this has been implemented? Did you ever want to have such a feature?

This is a very interesting idea I hadn't thought of. Would this be the first feature that enables `Parameter`s to depend on other `Parameter`s?

> 
> ### Weight
> The main issue I can see with this change is how much larger the `/parameters` route data would become. Currently, the aim of listing parameters is to make them discoverable, but not to expose all of their data and structure.
> 
> @sandcha since you ran this branch locally, could you please provide us with the file size increase for a large corpus such as OF-FR? As a baseline, `curl -L -I https://api.fr.openfisca.org/latest/parameters | grep content-length` yields `904421`, i.e. 883 kB. What would be the added weight with this PR?

Yes, this is a good point I hadn't considered.

MattiSG:
> Would this be the first feature that enables `Parameter`s to depend on other `Parameter`s?

In the example I gave, this “dependency” is only made clear through documentation and variables handle the (basic multiplication) maths. There is no explicit parameter linking that would be discoverable or handled by OF-Core.

Thanks for the additional information on `name`! This is much clearer now. But, on the other hand, the value of adding this to Core without this change becomes weaker 🙃 Note for future consolidation: this use case complements the proposals in https://github.com/openfisca/openfisca-france/issues/1672#issuecomment-939986515.

PR #1128, opened 2022-05-25
Remove ambiguous docstring references

MattiSG:
Some docstring refs are ambiguous. Until now, these warnings were not raised. Following [updates](https://github.com/openfisca/openfisca-doc/pull/264) in the documentation generation system, these references now raise errors when compiled and block documentation updates. This changeset removes links that are ambiguous so that the documentation can be generated again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+### 35.8.2 [#1128](https://github.com/openfisca/openfisca-core/pull/1128)
+
+#### Technical changes
+
+- Remove ambiguous links in docstrings.
+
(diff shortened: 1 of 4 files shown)


benjello (approved):
Approving to unblock even if do not know what :any: does ...

MattiSG:
I understand that `:any:` instructs the documentation builder to look for this definition in any class or module, even third-party ones. This makes it faster to write the doc as you don't need to find the fully-qualified namespace for a function, but it might be ambiguous.
In the cases I changed, I did not have enough time nor knowledge to find the full names, but the ambiguity led the engine to not know which link to create, which following the dependencies updates in openfisca/openfisca-doc#264 became blocking for compiling the doc.

PR #1131, opened 2022-05-30
Fix docstrings

maukoquiroga:
https://github.com/openfisca/openfisca-doc/pull/264

@MattiSG

### DIFF
--- a/openfisca_core/taxbenefitsystems/tax_benefit_system.py
+++ b/openfisca_core/taxbenefitsystems/tax_benefit_system.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Optional, Sequence
+from typing import Any, Dict, Optional, Sequence
 
 import copy
 import glob
@@ -39,7 +39,7 @@ class TaxBenefitSystem:
 
     """
     _base_tax_benefit_system = None
-    _parameters_at_instant_cache = None
+    _parameters_at_instant_cache: Optional[Dict[Any, Any]] = None
     person_key_plural = None
     preprocess_parameters = None
     baseline = None  # Baseline tax-benefit system. Used only by reforms. Note: Reforms can be chained.
@@ -50,8 +50,8 @@ class TaxBenefitSystem:
         # TODO: Currently: Don't use a weakref, because they are cleared by Paste (at least) at each call.
         self.parameters: Optional[ParameterNode] = None
         self._parameters_at_instant_cache = {}  # weakref.WeakValueDictionary()
-        self.variables = {}
-        self.open_api_config = {}
+        self.variables: Dict[Any, Any] = {}
+        self.open_api_config: Dict[Any, Any] = {}
         # Tax benefit systems are mutable, so entities (which need to know about our variables) can't be shared among them
         if entities is None or len(entities) == 0:
             raise Exception("A tax and benefit sytem must have at least an entity.")
--- a/openfisca_core/taxbenefitsystems/tax_benefit_system.py
+++ b/openfisca_core/taxbenefitsystems/tax_benefit_system.py
@@ -142,13 +142,15 @@ class TaxBenefitSystem:
 
         return variable
 
-    def add_variable(self, variable):
-        """
-        Adds an OpenFisca variable to the tax and benefit system.
+    def add_variable(self, variable: Variable) -> Variable:
+        """Adds an OpenFisca variable to the tax and benefit system.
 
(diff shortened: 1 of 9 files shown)


MattiSG:
I removed the `py:obj reference target not found:` warning by removing the `nitpicky` option in Sphinx. This hides it but does not make it go away entirely. See more in https://github.com/openfisca/openfisca-doc/pull/264#issuecomment-1157876485.

There are now problems detected by linting files in Core, that are not detected when building the doc:

```
openfisca_core/parameters/parameter.py:50: error: Incompatible types in assignment (expression has type "None", variable has type "str")
openfisca_core/parameters/parameter.py:52: error: Incompatible types in assignment (expression has type "None", variable has type "str")
openfisca_core/parameters/parameter.py:59: error: Incompatible types in assignment (expression has type "Optional[Any]", variable has type "str")
openfisca_core/parameters/parameter.py:67: error: Incompatible types in assignment (expression has type "Optional[Any]", variable has type "str")
openfisca_core/taxbenefitsystems/tax_benefit_system.py:52: error: Need type annotation for "_parameters_at_instant_cache" (hint: "_parameters_at_instant_cache: Dict[<type>, <type>] = ...")
openfisca_core/taxbenefitsystems/tax_benefit_system.py:53: error: Need type annotation for "variables" (hint: "variables: Dict[<type>, <type>] = ...")
openfisca_core/taxbenefitsystems/tax_benefit_system.py:54: error: Need type annotation for "open_api_config" (hint: "open_api_config: Dict[<type>, <type>] = ...")
openfisca_core/reforms/reform.py:76: error: "None" has no attribute "parameters"
```

Many of those points, while relevant, seem to have potential major implications, such as setting the default `description` of a `Parameter` to `""` instead of `None`. I don't feel experienced enough in Python docstrings to fully understand the consequences.

MattiSG:
I managed to make the compiler happy on all those items. However, the type hints I gave were often too wide, and some I purely removed as I did not know what to set. This is not a satisfactory nor relevant use of a static checker. Considering the added friction and the lack of documentation on how to handle such cases, I am in favour of removing type checks altogether.

openfisca_core/taxbenefitsystems/tax_benefit_system.py
```diff
@@ -231,7 +236,7 @@ def load_extension(self, extension):
         """
         Loads an extension to the tax and benefit system.
 
-        :param string extension: The extension to load. Can be an absolute path pointing to an extension directory, or the name of an OpenFisca extension installed as a pip package.
+        :param str extension: The extension to load. Can be an absolute path pointing to an extension directory, or the name of an OpenFisca extension installed as a pip package.
```
benjello, line 239: Shouldn't this be converted to google docstring style too ?

openfisca_core/taxbenefitsystems/tax_benefit_system.py
```diff
@@ -412,9 +418,9 @@ def get_variables(self, entity = None):
         """
         Gets all variables contained in a tax and benefit system.
 
-        :param .Entity entity: If set, returns only the variable defined for the given entity.
+        :param Entity entity: If set, returns only the variable defined for the given entity.
```
benjello, line 421: Shouldn't this be converted to google docstring style too ?

openfisca_core/variables/variable.py
```diff
@@ -309,7 +309,7 @@ def get_formula(self, period = None):
         If no period is given and the variable has several formula, return the oldest formula.
 
         :returns: Formula used to compute the variable
-        :rtype: .Formula
+        :rtype: :any:`Formula`
```
benjello, line 312: Shouldn't this be converted to google docstring style too ?

benjello (commented):
No particular expertise, just trying to help ..

openfisca_core/taxbenefitsystems/tax_benefit_system.py
MattiSG: Unfortunately I don't think there has been a clear decision made on the docstring style to be used. Missing this, I really just try to make the compiler happy so we can get the documentation to compile. Currently, there are no such style checks. I would prefer to defer unifying style to when we have a consensus on the style, and then use automatic formatters and add a linter on it.

MattiSG:
Thanks @benjello for your help! If you did not spot any issue during review yet prefer not to approve, I will approve this PR myself so we can unblock the documentation compilation.

benjello (approved):

PR #1137, opened 2022-07-28
Fix pristine editable install

### LINKED ISSUE #1136: Pristine editable install with make throws pytest error
Hi there!

I ~really~ so much enjoy OpenFisca ~but~ that I recently encountered an issue and I'm reporting it.

### Here is what I did:

I tried to pristine install the library in editable mode with make:

```sh
make install
```

To receive this message:

```
[⚙] Install project dependencies…
```

It works, but I'm getting an error related to `pytest`:

```
pyenv: pytest: command not found
...
[⚙] Install project dependencies… 
```

Python 3.9.13
OpenFisca 35.8.4

## Context

I identify more as a:

- Maintainer _(I work on OpenFisca's shared infrastructure)_.

maukoquiroga:
Fixes #1136 

#### Technical changes

- Fix pylint dependency in fresh editable installations
  - Ignore pytest requirement, used to collect test cases, if it is not yet installed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+### 35.8.5 [#1137](https://github.com/openfisca/openfisca-core/pull/1137)
+
+#### Technical changes
+
+- Fix pylint dependency in fresh editable installations
+  - Ignore pytest requirement, used to collect test cases, if it is not yet installed.
+
 ### 35.8.4 [#1131](https://github.com/openfisca/openfisca-core/pull/1131)
 
 #### Technical changes
--- a/openfisca_tasks/install.mk
+++ b/openfisca_tasks/install.mk
@@ -2,7 +2,7 @@
 install:
 	@$(call print_help,$@:)
 	@pip install --upgrade pip twine wheel
-	@pip install --editable .[dev] --upgrade --use-deprecated=legacy-resolver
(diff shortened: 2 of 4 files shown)


MattiSG (commented):
Thank you for this contribution! 
Could you please give an example to reproduce the error you encountered and intend to fix with this changeset? 🙂 I don't understand how it is possible to run `make test-core` without `pytest` installed, yet `openfisca` installed.

MattiSG:
Sorry, I just saw the link to #1136 🤦 Trying to reproduce now 🙂

MattiSG (approved):
- I could reproduce the issue described in #1136.
- Maintaining `make install` on top of the installation procedure described in the README is consistent with #1040.
- I checked `flake8` v4 changelog, and I believe the [breaking changes](https://flake8.pycqa.org/en/latest/release-notes/4.0.0.html#backwards-incompatible-changes) are not impactful for reusers.

While I believe silencing errors has potential side effects (see https://github.com/openfisca/openfisca-core/pull/1138#discussion_r933472293) and is not optimal (I would prefer that we stop `test-core` from being called when `make install` is called!), I can approve the current changeset.

maukoquiroga:
Thanks @MattiSG , I agree and actually find make increasingly painful to maintain as a tool. I hope we converge sooner than later to a more democratic and sustainable task manager for OpenFisca.

PR #1141, opened 2022-07-31
Fix doc test & build dependency errors

### LINKED ISSUE #1126: Install from scratch not working due to jinja transitive upgrade
Hi there!

### Here is what I did:

```
pip install --upgrade pip twine wheel
pip install openfisca-core[web-api]
pip install openfisca-france-fiscalite-miniere

openfisca serve --country-package openfisca_france_fiscalite_miniere
```

The server should be up and running

A lot of stacktraces with some jinjja escape error:

```
  File "/usr/local/bin/openfisca", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.7/site-packages/openfisca_core/scripts/openfisca_command.py", line 64, in main
    from openfisca_web_api.scripts.serve import main
  File "/usr/local/lib/python3.7/site-packages/openfisca_web_api/scripts/serve.py", line 7, in <module>
    from openfisca_web_api.app import create_app
  File "/usr/local/lib/python3.7/site-packages/openfisca_web_api/app.py", line 17, in <module>
    handle_import_error(error)
  File "/usr/local/lib/python3.7/site-packages/openfisca_web_api/errors.py", line 9, in handle_import_error
    raise ImportError("OpenFisca is missing some dependencies to run the Web API: '{}'. To install them, run `pip install openfisca_core[web-api]`.".format(error))
ImportError: OpenFisca is missing some dependencies to run the Web API: 'cannot import name 'escape' from 'jinja2' (/usr/local/lib/python3.7/site-packages/jinja2/__init__.py)'. To install them, run `pip install op
enfisca_core[web-api]`.
```

I was able to fix this by upgrading flask to the 2.1.2 version (in the setup.py it's the 1.1.4 specified)

I'm not sure this is the correct fix though, but from stackoverflow here: https://stackoverflow.com/questions/71718167/importerror-cannot-import-name-escape-from-jinja2 this may indeed be the root cause

## Context

I identify more as a:

### LINKED ISSUE #1136: Pristine editable install with make throws pytest error

### LINKED ISSUE #1140: Documentation testing broken since flake8 upgrade
Hi there!

### Here is what I did:

I created a PR #1138 

All tests to pass.

Doc tests [are failing](https://github.com/openfisca/openfisca-core/runs/7598050479?check_suite_focus=true) since the merging of #1137 .

It sems to me that #1137 did not create the problem, but solved it partially.

## Context

I identify more as a:

maukoquiroga:
Fixes #1136
Fixes #1140

#### Technical changes

- Fix `pycodestyle` dependency in fresh editable installations.
- Fix `flake8` dependency in fresh editable installations.

Note: this PR is the continuation of #1137, but seems the solution was not complete. This PR should unblock doc tests failing because of dependency mismatch.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,6 +27,7 @@ api_requirements = [
     'flask >= 1.1.4, < 2.0.0',
     'flask-cors == 3.0.10',
     'gunicorn >= 20.0.0, < 21.0.0',
+    'jinja2 >= 2.11.0, < 3.0.0',
     'werkzeug >= 1.0.1, < 2.0.0',
     ]
 
@@ -34,7 +35,7 @@ dev_requirements = [
     'autopep8 >= 1.4.0, < 1.6.0',
     'coverage == 6.0.2',
     'darglint == 1.8.0',
-    'flake8 >= 3.9.0, < 4.0.0',
+    'flake8 >= 5.0.0, < 5.1.0',
     'flake8-bugbear >= 19.3.0, < 20.0.0',
     'flake8-docstrings == 1.6.0',
     'flake8-print >= 3.1.0, < 4.0.0',
@@ -42,8 +43,8 @@ dev_requirements = [
     'mypy == 0.910',
     'openfisca-country-template >= 3.10.0, < 4.0.0',
     'openfisca-extension-template >= 1.2.0rc0, < 2.0.0',
-    'pycodestyle >= 2.7.0, < 2.8.0',
-    'pylint == 2.10.2',
+    'pycodestyle >= 2.9.0, < 2.10.0',
+    'pylint >= 2.13.0, < 2.14.0',
     ] + api_requirements
 
 setup(
--- a/setup.py
+++ b/setup.py
@@ -23,18 +23,18 @@ general_requirements = [
     ]
 
 api_requirements = [
-    'markupsafe >= 2.0.1, < 2.1.0',
-    'flask >= 1.1.4, < 2.0.0',
+    'markupsafe == 2.0.1',  # While flask revision < 2
+    'flask == 1.1.4',
     'flask-cors == 3.0.10',
     'gunicorn >= 20.0.0, < 21.0.0',
-    'werkzeug >= 1.0.1, < 2.0.0',
+    'werkzeug >= 1.0.0, < 2.0.0',
     ]
 
 dev_requirements = [
     'autopep8 >= 1.4.0, < 1.6.0',
     'coverage == 6.0.2',
     'darglint == 1.8.0',
-    'flake8 >= 3.9.0, < 4.0.0',
+    'flake8 >= 4.0.0, < 4.1.0',
     'flake8-bugbear >= 19.3.0, < 20.0.0',
     'flake8-docstrings == 1.6.0',
     'flake8-print >= 3.1.0, < 4.0.0',
@@ -42,7 +42,6 @@ dev_requirements = [
     'mypy == 0.910',
     'openfisca-country-template >= 3.10.0, < 4.0.0',
     'openfisca-extension-template >= 1.2.0rc0, < 2.0.0',
-    'pycodestyle >= 2.7.0, < 2.8.0',
     'pylint == 2.10.2',
     ] + api_requirements
 
--- a/setup.py
+++ b/setup.py
@@ -32,19 +32,20 @@ api_requirements = [
     ]
 
 dev_requirements = [
-    'autopep8 >= 1.4.0, < 1.6.0',
-    'coverage == 6.0.2',
-    'darglint == 1.8.0',
-    'flake8 >= 5.0.0, < 5.1.0',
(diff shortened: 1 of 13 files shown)


setup.py
```diff
@@ -23,25 +23,26 @@
...
-    'werkzeug >= 1.0.0, < 2.0.0',
+    'werkzeug >= 1.0.1, < 2.0.0',
     ]
 
 dev_requirements = [
     'autopep8 >= 1.4.0, < 1.6.0',
     'coverage == 6.0.2',
     'darglint == 1.8.0',
-    'flake8 >= 4.0.0, < 4.1.0',
+    'flake8 >= 3.9.0, < 4.0.0',
```
MattiSG, line 37: I am not sure I understand: this reverts #1137. Are you saying that the problem documented in #1136 was not coming from this dependency? Yet we saw that it fixed it. Downgrading `flake8` and forcing it to v3 does not seem like a sustainable solution.

MattiSG (commented):
I would like us to understand the reason why the tests passed in #1137 and failed afterwards on `master`. What differs? Until we understand this, all fixes are bound to be uncertain. We cannot iterate on production versions by publishing a new patch every day.

MattiSG:
As described in https://github.com/openfisca/openfisca-core/issues/1140#issuecomment-1200902579, I believe there is no way we currently can reproduce the documented failure, so we cannot know if publishing this version will fix the problem. I was trustful for #1137, but I cannot approve this PR under these circumstances now, especially considering that it currently reverts #1137 yet introduces other untestable dependency changes.

I would fast approve a pure #1137 revert, considering that the issue #1136 it fixed was extremely minor (misleading log with no consequence), or could consider approving this if #1140 included a repeatable procedure for testing changes. I would also ask that the `1> /dev/null` silencers in the `test-doc-install` step are removed so we can follow better what is happening.

maukoquiroga:
> As described in [#1140 (comment)](https://github.com/openfisca/openfisca-core/issues/1140#issuecomment-1200902579), I believe there is no way we currently can reproduce the documented failure, so we cannot know if publishing this version will fix the problem. I was trustful for #1137, but I cannot approve this PR under these circumstances now, especially considering that it currently reverts #1137 yet introduces other untestable dependency changes.
> 
> I would fast approve a pure #1137 revert, considering that the issue #1136 it fixed was extremely minor (misleading log with no consequence), or could consider approving this if #1140 included a repeatable procedure for testing changes. I would also ask that the `1> /dev/null` silencers in the `test-doc-install` step are removed so we can follow better what is happening.

Actually the `flake8` upgrade is a knight unrelated to the fix.

maukoquiroga:
Fantastic, so what I get now:

```
flake8.exceptions.FailedToLoadPlugin: Flake8 failed to load plugin "pycodestyle.missing_whitespace_after_import_keyword" due to module 'pycodestyle' has no attribute 'missing_whitespace_after_import_keyword'.
```

So this is a layer 8 error: instead of downgrading pycodestyle, I proposed upgrading flask8. 

It concurred with the silencing and we got no clue of what was happing.

Let's remove the silencers!

maukoquiroga:
Now by suppressing the silencers we have actionable logs:

```
The conflict is caused by:
    sphinx 4.5.0 depends on importlib-metadata>=4.4; python_version < "3.10"
    flake8 4.0.1 depends on importlib-metadata<4.3; python_version < "3.8"
    sphinx 4.5.0 depends on importlib-metadata>=4.4; python_version < "3.10"
    flake8 4.0.0 depends on importlib-metadata<4.3; python_version < "3.8"
```

https://github.com/openfisca/openfisca-core/runs/7611320470?check_suite_focus=true

Which confirms the need to rollback that upgrade.

Rollback or no, 

> Or could consider approving this if https://github.com/openfisca/openfisca-core/issues/1140 included a repeatable procedure for testing changes. I would also ask that the 1> /dev/null silencers in the test-doc-install step are removed so we can follow better what is happening.

I think this PR now provides an increase in our understanding of what's going on.

setup.py
maukoquiroga: Upgrading flake8 introduced a dependency conflict with sphinx, that was being silenced by `test-doc`.

maukoquiroga:
As I put it in the last commit:

_Versions >= 4.0.0 of flake8 require a library version, for clients using
Python < 3,8, that is incompatible with the same library requirement by
Sphinx >= 4.5._

PR #1142, opened 2022-07-31
Deprecate Instant.period

maukoquiroga:
Depends on #1160 

#### Deprecations

- In _periods.Instant_:
  - Remove `period`, method used to build a `Period` from an `Instant`.
  - This method created an upward circular dependency between `Instant` and `Period` causing lots of trouble.
  - The functionality is still provided by `periods.period` and the `Period` constructor.

#### Migration details

- Replace `some_period.start.period` and similar methods with `Period((unit, some_period.start, 1))`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Changelog
 
+# 37.0.0 [#1142](https://github.com/openfisca/openfisca-core/pull/1142)
+
+#### Deprecations
+
+- In _periods.Instant_:
+  - Remove `period`, method used to build a `Period` from an `Instant`.
+  - This method created an upward circular dependency between `Instant` and `Period` causing lots of trouble.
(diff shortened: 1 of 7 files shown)


MattiSG:
I will postpone reviewing this PR until the one it depends on is approved, and suggest that this one is set to draft. Indeed, I fail to see how it could be approved when #1141 might not go through.

Do you truly need #1141 for this one to be reviewed? 🙂 As an alternative to switching to draft, I suggest to remove the dependency and rebase straight on `master`. Yes, `test-docs` might fail until #1141 is fixed, but that is true of all PRs, not only of this one. This would enable reviews to be parallelised and proper solutions found for #1140 without blocking all the others.

maukoquiroga:
cc @MattiSG

benjello (approved):

PR #1143, opened 2022-07-31
Re-add support for dispatch/divide by days

maukoquiroga:
Fixes govzeroaotearoa/openfisca-aotearoa#15
Depends on #1141 

#### New features

- Introduce support for the ``day`` date unit in `holders.set_input_dispatch_by_period` and `holders.set_input_divide_by_period`
  - Allows for dispatching values per day, for example, to provide a daily (week, fortnight) to an yearly variable.
  - Inversely, allows for calculating the daily (week, fortnight) value of a yearly input.

### DIFF
--- a/openfisca_core/holders/tests/test_helpers.py
+++ b/openfisca_core/holders/tests/test_helpers.py
@@ -57,22 +57,67 @@ def three_days(beggining_of_year):
     return Period((periods.DAY, beggining_of_year, 3))
 
 
-def test_set_input_dispatch_by_period(
+@pytest.mark.parametrize("definition_period, values, expected", [
+    [periods.YEAR, [43800.], [43800. * 3]],
+    [periods.MONTH, [3650.], [3650. * 12 * 3]],
+    # [periods.DAY, [120.], [131400.]],
+    ])
+def test_set_input_dispatch_by_period_over_3_years(
         Income,
+        definition_period,
         population,
         three_years,
+        values,
+        expected,
         ):
-    """Yearly income propagates evenly to the following years."""
-
-    Income.definition_period = periods.MONTH
+    Income.definition_period = definition_period
     income = Income()
     holder = Holder(income, population)
-    values = [1000.]
 
     holders.set_input_dispatch_by_period(holder, three_years, values)
 
-    known_periods = holder.get_known_periods()
-    period_count = len(known_periods)
+    assert sum(map(holder.get_array, holder.get_known_periods())) == expected
 
-    assert period_count == 36  # Three years in months
-    assert sum(map(holder.get_array, known_periods)) == [36000.]
+
+@pytest.mark.parametrize("definition_period, values, expected", [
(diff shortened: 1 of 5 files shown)


MattiSG:
I will postpone reviewing this PR until the one it depends on is approved, and suggest that this one is set to draft. Indeed, I fail to see how it could be approved when #1141 might not go through.

Do you truly need #1141 for this one to be reviewed? 🙂 As an alternative to switching to draft, I suggest to remove the dependency and rebase straight on `master`. Yes, `test-docs` might fail until #1141 is fixed, but that is true of all PRs, not only of this one. This would enable reviews to be parallelised and proper solutions found for #1140 without blocking all the others.

maukoquiroga:
Done!

maukoquiroga:
I'm not sure if this is a new feature, or a bug-fix as the behaviour this adds was introduced as a replacement of a behaviour that was deprecated, however the pull request forgot to include `days` in that migration.

benjello:
@maukoquiroga : looks very good to me.
Just a question: how does this PR handles month day size, february and leeap year for divide_by_period ?
It does not seem to be tested although there are a bunch of tests for that in `Period` and `Instant`.

maukoquiroga:
> @maukoquiroga : looks very good to me. Just a question: how does this PR handles month day size, february and leeap year for divide_by_period ? It does not seem to be tested although there are a bunch of tests for that in `Period` and `Instant`.

I didn't do a lot more than «reenabling» a feature that already existed.

If you see in the tests, month days are handled appropriately (31 + 28 + 30).

Maybe test should be more precise about leap years.

maukoquiroga:
(By the way, I think that holders implements it's own logic in terms of equivalences between days, months and years. We should see which is better, and keep just one in periods, as holder should not handle that kind of logic IMHO).

maukoquiroga:
@benjello @MattiSG could you please take a look at this pull request again?

maukoquiroga:
cc @MattiSG @benjello

openfisca_core/holders/tests/test_helpers.py
```diff
@@ -0,0 +1,98 @@
...
+def population(people):
+    population = Population(people)
+    population.count = 1
+    return population
+
+
+@pytest.mark.parametrize("dispatch_unit, definition_unit, values, expected", [
+    [periods.YEAR, periods.YEAR, [1.], [3.]],
+    [periods.YEAR, periods.MONTH, [1.], [36.]],
+    [periods.YEAR, periods.DAY, [1.], [1096.]],
```
benjello, line 42: Why it is not 1095 = 3 * 365 ? (Is it because a year is approx. 395.25 ?) and you rounded the number ?
It is different from the division counter-part.

maukoquiroga: It is just because there's a leap year with 366 days.

I'm counting calendar days instead of multiplying by 365.

benjello (approved):

PR #1145, opened 2022-08-01
Drop doc build tester

maukoquiroga:
Fixes #1140 
Fixes #1141

#### Technical changes

- Removes the automatic documentation build check
  - It has been proven difficult to maintain, specifically due _dependency hell_ and a very contrived build workflow.

PR #1149, opened 2022-08-08
Introduce error margin by variable in tests

benjello:

#### New features

- Introduce variable dependent error margins int YAML tests
  - Allows for different error margin across variables

### DIFF
--- a/openfisca_core/tools/test_runner.py
+++ b/openfisca_core/tools/test_runner.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 
+import collections
 import warnings
 import sys
 import os
@@ -214,14 +215,28 @@ class YamlItem(pytest.Item):
 
         actual_value = self.simulation.calculate(variable_name, period)
 
+        absolute_error_margin = self.test.get('absolute_error_margin')
+        variable_absolute_error_margin = (
+            absolute_error_margin.get(variable_name, absolute_error_margin.get("default"))
+            if isinstance(absolute_error_margin, collections.Mapping)
+            else absolute_error_margin
+            )
+
+        relative_error_margin = self.test.get('relative_error_margin')
+        variable_relative_error_margin = (
+            relative_error_margin.get(variable_name, relative_error_margin.get("default"))
+            if isinstance(relative_error_margin, collections.Mapping)
+            else relative_error_margin
+            )
+
         if entity_index is not None:
             actual_value = actual_value[entity_index]
         return assert_near(
             actual_value,
             expected_value,
-            absolute_error_margin = self.test.get('absolute_error_margin'),
+            absolute_error_margin = variable_absolute_error_margin,
             message = f"{variable_name}@{period}: ",
-            relative_error_margin = self.test.get('relative_error_margin'),
+            relative_error_margin = variable_relative_error_margin,
             )
 
     def should_ignore_variable(self, variable_name: str):
--- a/openfisca_core/tools/test_runner.py
+++ b/openfisca_core/tools/test_runner.py
@@ -1,6 +1,5 @@
-# -*- coding: utf-8 -*-
+from __future__ import annotations
 
-import collections
 import warnings
 import sys
 import os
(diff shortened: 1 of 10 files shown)


nikhilwoodruff (approved):
Great, thanks @benjello. This will be useful in both UK and US systems.

benjello:
Doc building fails and should be fixed by #1145.

@maukoquiroga and @MattiSG please ping me when the later is merged so I can move on with this PR.
Thanks !

maukoquiroga:
Hi @benjello ! Thanks for your contribution. Could you please write a couple of lines so I or an other reviewers can understand the motivation for this PR? Just 

1. Need
2. Complication to solve the need with current code base
3. What you propose and why you propose it this way and not another way

Quickly from reading the code this is what I got:

1. We already accept calculation level error margins, which are absolute. 
2. However, because quality of data is not always homogeneous, in some cases, outlier errors in one variable calculation makes the whole calculation to fail. Which is generally OK except when we want to observe one variable in isolation.
3. You propose to define 
  a. A calculation-level error margin that is called absolute in terms of the desired result.
  b. A variable-level error margin that is relative to the absolute one and the desired result.
  c. The `default` error is absolute and is a scalar, e.g. `100 €`.
  d. The `relative` error is relative to the `default` and is a percentage, e.g. `5%`.

openfisca_core/tools/test_runner.py
```diff
@@ -212,14 +213,28 @@ def check_variable(self, variable_name, expected_value, period, entity_index = N
 
         actual_value = self.simulation.calculate(variable_name, period)
 
+        absolute_error_margin = self.test.get('absolute_error_margin')
+        variable_absolute_error_margin = (
+            absolute_error_margin.get(variable_name, absolute_error_margin.get("default"))
```
maukoquiroga, line 218: Why call it default and not just absolute error margin?

openfisca_core/tools/test_runner.py
```diff
@@ -212,14 +213,28 @@ def check_variable(self, variable_name, expected_value, period, entity_index = N
...
+            if isinstance(absolute_error_margin, collections.Mapping)
```
maukoquiroga, line 219: Can't we rather assume that if not declared, relative error margin is 0%?

openfisca_core/tools/test_runner.py
```diff
@@ -212,14 +213,28 @@ def check_variable(self, variable_name, expected_value, period, entity_index = N
...
         if entity_index is not None:
             actual_value = actual_value[entity_index]
         return assert_near(
             actual_value,
             expected_value,
-            absolute_error_margin = self.test.get('absolute_error_margin'),
+            absolute_error_margin = variable_absolute_error_margin,
             message = f"{variable_name}@{period}: ",
-            relative_error_margin = self.test.get('relative_error_margin'),
+            relative_error_margin = variable_relative_error_margin,
```
maukoquiroga, line 237: How did this work previously, do you have an example?

tests/fixtures/yaml_tests/failing_test_variable_absolute_error_margin.yaml
```diff
@@ -0,0 +1,9 @@
+- name: "Failing test: result out of variable specific absolute error margin"
```
maukoquiroga, line 1: I'm not clear here: the absolute error margin is for the whole calculation or just for the variable?

tests/fixtures/yaml_tests/failing_test_variable_absolute_error_margin.yaml
```diff
@@ -0,0 +1,9 @@
...
+  period: 2015-01
+  absolute_error_margin:
+    default: 100
+    income_tax: 50
```
maukoquiroga, line 5: Ah, so here you declare an absolute for income tax overriding the abdolute of the calculaiton ?

tests/fixtures/yaml_tests/failing_test_variable_relative_error_margin.yaml
```diff
@@ -0,0 +1,9 @@
+- name: "Failing test: result out of variable specific relative error margin"
+  period: 2015-01
+  relative_error_margin:
+    default: 1
```
maukoquiroga, line 4: I don't understand here the logic. Relative error margin to relative to what? Are we here again just overriding the relative margin error just fro income tax? 

Maybe an example where this is useful including other variables could come handy to better grasp the intent.

tests/fixtures/yaml_tests/test_relative_error_margin.yaml
```diff
@@ -5,3 +5,14 @@
     salary: 2000
   output:
     income_tax: 290 # 300
+
+
+- name: "Result within variable relative error margin"
+  period: 2015-01
+  relative_error_margin:
+    default: .001
+    income_tax: 0.05
```
maukoquiroga, line 14: So here we demonstrate that test passes because the result is within the lower and upper bounds (+-15), but should otherwise fail with the general one (+-3). Could we show that in the test to it's clear?

maukoquiroga:
This goes beyond the PR because error margin logic already exist in OF, but to be clear, when we say _income tax has a relative margin error of 5%_, we're saying just that _income tax can have an absolute margin error of `+- result * relative error margin`_ right?

openfisca_core/tools/test_runner.py
benjello: You want to be able to easily (continuously) deviate from the current situation where you have a global error margin

openfisca_core/tools/test_runner.py
benjello: Current tests clearly shows how margins can be used today.

tests/fixtures/yaml_tests/failing_test_variable_absolute_error_margin.yaml
benjello: The default is for all the variables that are not directly specified.
Imagine you set a 1% relative error margin. 
```yaml
relative_error_margin: .01
```
But you find out that want to have a less strict, say 5%, error margin on some variables.
You can easily modify the previous line to:
```yaml
relative_error_margin: 
- default: .01
- my_variable: .05
```

tests/fixtures/yaml_tests/failing_test_variable_relative_error_margin.yaml
benjello: relative error margin = abs(computed_value - target_value / target_value)  is here 100%.

An example is pension there are some variables where it is very difficult to compute up to 1% error.
And other where it is crucial.

benjello:
> This goes beyond the PR because error margin logic already exist in OF, but to be clear, when we say _income tax has a relative margin error of 5%_, we're saying just that _income tax can have an absolute margin error of `+- result * relative error margin`_ right?

See above for the definition of relative error margin

benjello:
@maukoquiroga : does my answer clarify the PR. May be we can make it clearer in the doc after merging the PR.

maukoquiroga:
> @maukoquiroga : does my answer clarify the PR. May be we can make it clearer in the doc after merging the PR.

Yes, I understand better.

The reason I ask all this questions is to make it easier for any reviewer out there to get the logic and review the code, and eventually accept/reject the proposal.

I certainly have no grounds to reject it now that I understand the value it adds, however I wasn't able to review the actual implementation of it without that understanding.

I hope to give you a proper review soon if nobody gives you one before.

openfisca_core/tools/test_runner.py
benjello: It is the default of the absolute error margin

benjello:
@nikhilwoodruff @maukoquiroga : I do need a final approval to merge this PR ;-)

maukoquiroga (approved):

PR #1150, opened 2022-08-25
Add a depth-max for test logs

benjello:
#### New features

- Introduce a maximal depth for compuation logs
  - Allows for limiting the depth of the computation log chain

### DIFF
--- a/openfisca_core/scripts/openfisca_command.py
+++ b/openfisca_core/scripts/openfisca_command.py
@@ -41,6 +41,7 @@ def get_parser():
         parser.add_argument('-v', '--verbose', action = 'store_true', default = False, help = "increase output verbosity")
         parser.add_argument('-o', '--only-variables', nargs = '*', default = None, help = "variables to test. If specified, only test the given variables.")
         parser.add_argument('-i', '--ignore-variables', nargs = '*', default = None, help = "variables to ignore. If specified, do not test the given variables.")
+        parser.add_argument('-d', '--depth_max', type = int, default = None, help = "Maximal depth of the log. If not specified, display all the computation")
 
         return parser
 
--- a/openfisca_core/scripts/openfisca_command.py
+++ b/openfisca_core/scripts/openfisca_command.py
@@ -38,10 +38,11 @@ def get_parser():
         parser.add_argument('-p', '--pdb', action = 'store_true', default = False, help = "drop into debugger on failures or errors")
         parser.add_argument('--performance-graph', '--performance', action = 'store_true', default = False, help = "output a performance graph in a 'performance_graph.html' file")
         parser.add_argument('--performance-tables', action = 'store_true', default = False, help = "output performance CSV tables")
-        parser.add_argument('-v', '--verbose', action = 'store_true', default = False, help = "increase output verbosity")
+        parser.add_argument('-v', '--verbose', action = 'store_true', default = False, help = "increase output verbosity. If specified, output the entire calculation trace.")
+        parser.add_argument('-a', '--aggregate', action = 'store_true', default = False, help = "increase output verbosity to aggregate. If specified, output the avg, max, and min values of the calculation trace. This flag has no effect without --verbose.")
+        parser.add_argument('-d', '--max-depth', type = int, default = None, help = "set maximal verbosity depth. If specified, output the calculation trace up to the provided depth. This flag has no effect without --verbose.")
         parser.add_argument('-o', '--only-variables', nargs = '*', default = None, help = "variables to test. If specified, only test the given variables.")
         parser.add_argument('-i', '--ignore-variables', nargs = '*', default = None, help = "variables to ignore. If specified, do not test the given variables.")
-        parser.add_argument('-d', '--max_depth', type = int, default = None, help = "Maximal depth of the log. If not specified, display all the computation")
 
         return parser
 
--- a/openfisca_core/scripts/openfisca_command.py
+++ b/openfisca_core/scripts/openfisca_command.py
@@ -41,7 +41,7 @@ def get_parser():
         parser.add_argument('-v', '--verbose', action = 'store_true', default = False, help = "increase output verbosity")
         parser.add_argument('-o', '--only-variables', nargs = '*', default = None, help = "variables to test. If specified, only test the given variables.")
         parser.add_argument('-i', '--ignore-variables', nargs = '*', default = None, help = "variables to ignore. If specified, do not test the given variables.")
-        parser.add_argument('-d', '--depth_max', type = int, default = None, help = "Maximal depth of the log. If not specified, display all the computation")
+        parser.add_argument('-d', '--max_depth', type = int, default = None, help = "Maximal depth of the log. If not specified, display all the computation")
 
         return parser
 
--- a/openfisca_core/scripts/run_test.py
+++ b/openfisca_core/scripts/run_test.py
@@ -22,6 +22,7 @@ def main(parser):
         'name_filter': args.name_filter,
         'only_variables': args.only_variables,
         'ignore_variables': args.ignore_variables,
+        'depth_max': args.depth_max,
         }
 
     paths = [os.path.abspath(path) for path in args.path]
--- a/openfisca_core/scripts/run_test.py
+++ b/openfisca_core/scripts/run_test.py
@@ -19,10 +19,11 @@ def main(parser):
         'performance_graph': args.performance_graph,
         'performance_tables': args.performance_tables,
         'verbose': args.verbose,
+        'aggregate': args.aggregate,
+        'max_depth': args.max_depth,
         'name_filter': args.name_filter,
         'only_variables': args.only_variables,
         'ignore_variables': args.ignore_variables,
-        'depth_max': args.depth_max,
         }
 
     paths = [os.path.abspath(path) for path in args.path]
--- a/openfisca_core/tools/test_runner.py
+++ b/openfisca_core/tools/test_runner.py
@@ -146,6 +146,7 @@ class YamlItem(pytest.Item):
         verbose = self.options.get('verbose')
         performance_graph = self.options.get('performance_graph')
         performance_tables = self.options.get('performance_tables')
+        depth_max = self.options.get('depth_max')
 
         try:
             builder.set_default_period(period)
@@ -165,15 +166,15 @@ class YamlItem(pytest.Item):
         finally:
             tracer = self.simulation.tracer
             if verbose:
-                self.print_computation_log(tracer)
+                self.print_computation_log(tracer, depth_max)
             if performance_graph:
                 self.generate_performance_graph(tracer)
             if performance_tables:
                 self.generate_performance_tables(tracer)
 
-    def print_computation_log(self, tracer):
+    def print_computation_log(self, tracer, depth_max = None):
         print("Computation log:")  # noqa T001
-        tracer.print_computation_log()
+        tracer.print_computation_log(depth_max = depth_max)
 
     def generate_performance_graph(self, tracer):
         tracer.generate_performance_graph('.')
--- a/openfisca_core/tools/test_runner.py
+++ b/openfisca_core/tools/test_runner.py
@@ -144,9 +144,10 @@ class YamlItem(pytest.Item):
         period = self.test.get('period')
         max_spiral_loops = self.test.get('max_spiral_loops')
         verbose = self.options.get('verbose')
+        aggregate = self.options.get('aggregate')
+        max_depth = self.options.get('max_depth')
         performance_graph = self.options.get('performance_graph')
         performance_tables = self.options.get('performance_tables')
-        depth_max = self.options.get('depth_max')
 
         try:
             builder.set_default_period(period)
@@ -166,15 +167,15 @@ class YamlItem(pytest.Item):
         finally:
             tracer = self.simulation.tracer
             if verbose:
-                self.print_computation_log(tracer, depth_max)
+                self.print_computation_log(tracer, aggregate, max_depth)
             if performance_graph:
                 self.generate_performance_graph(tracer)
             if performance_tables:
                 self.generate_performance_tables(tracer)
 
-    def print_computation_log(self, tracer, depth_max = None):
+    def print_computation_log(self, tracer, aggregate, max_depth):
         print("Computation log:")  # noqa T001
-        tracer.print_computation_log(depth_max = depth_max)
+        tracer.print_computation_log(aggregate, max_depth)
 
     def generate_performance_graph(self, tracer):
         tracer.generate_performance_graph('.')
--- a/openfisca_core/tracers/computation_log.py
+++ b/openfisca_core/tracers/computation_log.py
@@ -73,7 +73,7 @@ class ComputationLog:
             max_depth: Optional[int],
             ) -> List[str]:
 
-        if max_depth is not None and (depth > max_depth):
+        if max_depth is not None and depth > max_depth:
             return []
 
         node_log = [self._print_line(depth, node, aggregate, max_depth)]
--- a/openfisca_core/tracers/computation_log.py
+++ b/openfisca_core/tracers/computation_log.py
@@ -35,6 +35,7 @@ class ComputationLog:
             node: tracers.TraceNode,
             depth: int,
             aggregate: bool,
+            depth_max: int = None,
             ) -> List[str]:
 
         def print_line(depth: int, node: tracers.TraceNode) -> str:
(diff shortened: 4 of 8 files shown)


nikhilwoodruff (approved):
Thanks @benjello! While the general OpenFisca framework is really developer-friendly for debugging, overly dense output sometimes makes debugging more difficult. This will help a lot for the UK and US systems!

openfisca_core/scripts/openfisca_command.py
```diff
@@ -41,6 +41,7 @@ def build_test_parser(parser):
         parser.add_argument('-v', '--verbose', action = 'store_true', default = False, help = "increase output verbosity")
         parser.add_argument('-o', '--only-variables', nargs = '*', default = None, help = "variables to test. If specified, only test the given variables.")
         parser.add_argument('-i', '--ignore-variables', nargs = '*', default = None, help = "variables to ignore. If specified, do not test the given variables.")
+        parser.add_argument('-d', '--depth_max', type = int, default = None, help = "Maximal depth of the log. If not specified, display all the computation")
```
maukoquiroga, line 44:
```suggestion
        parser.add_argument('-l', '--max-log-depth', type = int, default = None, help = "Maximal log depth. If specified, cap log verbosity to the provided calculation depth.")
```

openfisca_core/scripts/run_test.py
```diff
@@ -22,6 +22,7 @@ def main(parser):
         'name_filter': args.name_filter,
         'only_variables': args.only_variables,
         'ignore_variables': args.ignore_variables,
+        'depth_max': args.depth_max,
```
maukoquiroga, line 25:
```suggestion
        'max_log_depth': args.max_log_depth,
```

openfisca_core/tools/test_runner.py
```diff
@@ -146,6 +146,7 @@ def runtest(self):
         verbose = self.options.get('verbose')
         performance_graph = self.options.get('performance_graph')
         performance_tables = self.options.get('performance_tables')
+        depth_max = self.options.get('depth_max')
```
maukoquiroga, line 149:
```suggestion
        max_log_depth = self.options.get('max_log_depth')
```

openfisca_core/tools/test_runner.py
```diff
@@ -165,15 +166,15 @@ def runtest(self):
         finally:
             tracer = self.simulation.tracer
             if verbose:
-                self.print_computation_log(tracer)
+                self.print_computation_log(tracer, depth_max)
```
maukoquiroga, line 169: This seems to break or interfere with the _verbose_ option, is it what we want?

maukoquiroga: In that case, the whole feature could be just called _verbosity depth_.

benjello: Yes the computation log is not produced if the verbose option is not triggered.
Ok for verbosity depth. Thanks !

maukoquiroga:
Hi @benjello and @nikhilwoodruff ! Thanks for this contribution.

There's already an `aggregate` option for the tracer, this is a different feature?
https://openfisca.org/doc/openfisca-python-api/tracer.html#openfisca_core.tracers.Tracer.print_computation_log

openfisca_core/tools/test_runner.py
maukoquiroga: Ok let me change my suggestions.

maukoquiroga:
Also @benjello could you please:

1. Provide a command to test the feature in the command line? :)
2. Tracer is unit tested, so for this PR to get accepted I suggest adding a unit test

benjello:
> Hi @benjello and @nikhilwoodruff ! Thanks for this contribution.
> 
> There's already an `aggregate` option for the tracer, this is a different feature? https://openfisca.org/doc/openfisca-python-api/tracer.html#openfisca_core.tracers.Tracer.print_computation_log

Yes completely different. The agreggate option is for test on population data.

openfisca_core/tracers/full_tracer.py
```diff
@@ -121,8 +121,8 @@ def flat_trace(self) -> tracers.FlatTrace:
     def _get_time_in_sec(self) -> float:
         return time.time_ns() / (10**9)
 
-    def print_computation_log(self, aggregate = False):
-        self.computation_log.print_log(aggregate)
+    def print_computation_log(self, depth_max = None, aggregate = False):
```
maukoquiroga, line 124: This is a breaking change, I suggest appending after aggregate.

benjello: Please do !

openfisca_core/tools/test_runner.py
benjello: Please do !

openfisca_core/scripts/openfisca_command.py
maukoquiroga: @benjello Makes sense?

maukoquiroga (changes requested):
This is a great contribution @benjello :)

I have no questioning on the value of this feature.

I do however, in order to integrate it, need we work on three aspects of it:

1. Having a reproducible way of testing this. I suggest we complete https://openfisca.org/doc/openfisca-python-api/openfisca_test.html with an example of using this flag.

2. That we unit test this feature, in a similar fashion that the `aggregate` one.

3. That we test whether `aggregate` and this one work together or not, and that we make it clear for the user.

openfisca_core/scripts/run_test.py
maukoquiroga, line 25:
```suggestion
        'aggregate': args.aggregate,
        'depth': args.depth,
```

openfisca_core/tools/test_runner.py
maukoquiroga, line 149:
```suggestion
        aggregate = self.options.get('aggregate')
        max_depth = self.options.get('max_depth')        
```

openfisca_core/tools/test_runner.py
maukoquiroga, line 169: Here depends on `average` and `max-depth` compatibility, a proposal.

openfisca_core/tools/test_runner.py
```diff
@@ -165,15 +166,15 @@ def runtest(self):
         finally:
             tracer = self.simulation.tracer
             if verbose:
```
maukoquiroga, line 168: Here depends on `average` and `max-depth` compatibility, a proposal:

```suggestion
            if verbose and aggregate:
                self.print_computation_log(tracer, aggregate = True)

            if verbose and max_depth:
                self.print_computation_log(tracer, max_depth = max_depth)

            if verbose:
                self.print_computation_log(tracer)
```

openfisca_core/scripts/openfisca_command.py
benjello: @maukoquiroga : aggregate and calculation depth are orthogonal.
Depth is about showing the value of calculation when you have variable A depending on B depending on C etc.
You may wwant to stop at B (`max-depth = 2`).

Aggregate is about showing `mean`/`max`/`min` of the variables A or B o C on your whole population.

openfisca_core/tools/test_runner.py
```diff
@@ -165,15 +166,15 @@ def runtest(self):
...
-                self.print_computation_log(tracer)
+                self.print_computation_log(tracer, depth_max)
             if performance_graph:
                 self.generate_performance_graph(tracer)
             if performance_tables:
                 self.generate_performance_tables(tracer)
 
-    def print_computation_log(self, tracer):
+    def print_computation_log(self, tracer, depth_max = None):
```
maukoquiroga, line 175:
```suggestion
    def print_computation_log(self, tracer, aggregate = False, max_depth = None):
```

openfisca_core/tools/test_runner.py
```diff
@@ -165,15 +166,15 @@ def runtest(self):
...
         print("Computation log:")  # noqa T001
-        tracer.print_computation_log()
+        tracer.print_computation_log(depth_max = depth_max)
```
maukoquiroga, line 177:
```suggestion
        tracer.print_computation_log(aggregate, max_depth)
```

openfisca_core/tracers/full_tracer.py
maukoquiroga, line 124:
```suggestion
    def print_computation_log(self, aggregate = False, max_depth = None):
```

openfisca_core/tracers/computation_log.py
```diff
@@ -35,6 +35,7 @@ def _get_node_log(
             node: tracers.TraceNode,
             depth: int,
             aggregate: bool,
+            depth_max: int = None,
```
maukoquiroga, line 38:
```suggestion
            max_depth: Optional[int] = None,
```

openfisca_core/tracers/computation_log.py
```diff
@@ -76,18 +81,18 @@ def _flatten(
             ) -> List[str]:
         return [item for _list in list_of_lists for item in _list]
 
-    def lines(self, aggregate: bool = False) -> List[str]:
+    def lines(self, aggregate: bool = False, depth_max: int = None) -> List[str]:
```
maukoquiroga, line 84:
```suggestion
    def lines(self, aggregate: bool = False, max_depth: Optional[int] = None) -> List[str]:
```

openfisca_core/tracers/computation_log.py
```diff
@@ -76,18 +81,18 @@ def _flatten(
...
         depth = 1
 
         lines_by_tree = [
-            self._get_node_log(node, depth, aggregate)
+            self._get_node_log(node, depth, aggregate, depth_max = depth_max)
```
maukoquiroga, line 88:
```suggestion
            self._get_node_log(node, depth, aggregate, max_depth)
```

openfisca_core/tracers/computation_log.py
```diff
@@ -76,18 +81,18 @@ def _flatten(
...
             for node
             in self._full_tracer.trees
             ]
 
         return self._flatten(lines_by_tree)
 
-    def print_log(self, aggregate = False) -> None:
+    def print_log(self, aggregate = False, depth_max = None) -> None:
```
maukoquiroga, line 95:
```suggestion
    def print_log(self, aggregate = False, max_depth = None) -> None:
```

openfisca_core/tracers/computation_log.py
```diff
@@ -99,5 +104,5 @@ def print_log(self, aggregate = False) -> None:
 
         This mode is more suited for simulations on a large population.
         """
-        for line in self.lines(aggregate):
+        for line in self.lines(aggregate, depth_max):
```
maukoquiroga, line 107:
```suggestion
        for line in self.lines(aggregate, max_depth):
```

openfisca_core/tracers/computation_log.py
```diff
@@ -60,12 +61,16 @@ def print_line(depth: int, node: tracers.TraceNode) -> str:
 
             return f"{indent}{node.name}<{node.period}> >> {formatted_value}"
 
+        if (depth_max is not None) and (depth > depth_max):
```
maukoquiroga, line 64: Can't it be `None` ?

openfisca_core/tracers/computation_log.py
```diff
@@ -60,12 +61,16 @@ def print_line(depth: int, node: tracers.TraceNode) -> str:
...
+            return [""]
+
         node_log = [print_line(depth, node)]
 
         children_logs = [
-            self._get_node_log(child, depth + 1, aggregate)
+            self._get_node_log(child, depth + 1, aggregate, depth_max = depth_max)
             for child
             in node.children
+            if self._get_node_log(child, depth + 1, aggregate, depth_max = depth_max) != [""]
```
maukoquiroga, line 73: Is this line reachable? I think as you return before, you never get here.

openfisca_core/tracers/computation_log.py
maukoquiroga, line 64:
```suggestion
        if max_depth is not None and depth > max_depth:
```

openfisca_core/tracers/computation_log.py
```diff
@@ -60,12 +61,16 @@ def print_line(depth: int, node: tracers.TraceNode) -> str:
...
             return f"{indent}{node.name}<{node.period}> >> {formatted_value}"
 
+        if (depth_max is not None) and (depth > depth_max):
+            return [""]
+
         node_log = [print_line(depth, node)]
 
         children_logs = [
-            self._get_node_log(child, depth + 1, aggregate)
+            self._get_node_log(child, depth + 1, aggregate, depth_max = depth_max)
```
maukoquiroga, line 70:
```suggestion
            self._get_node_log(child, depth + 1, aggregate, max_depth)
```

openfisca_core/tracers/computation_log.py
```diff
@@ -60,12 +61,16 @@ def print_line(depth: int, node: tracers.TraceNode) -> str:
 
             return f"{indent}{node.name}<{node.period}> >> {formatted_value}"
 
+        if (depth_max is not None) and (depth > depth_max):
+            return [""]
```
maukoquiroga, line 65:
```suggestion
            return None
```

openfisca_core/tracers/computation_log.py
maukoquiroga, line 73:
```suggestion
```

openfisca_core/tracers/computation_log.py
```diff
@@ -60,12 +61,16 @@ def print_line(depth: int, node: tracers.TraceNode) -> str:
...
 
         children_logs = [
-            self._get_node_log(child, depth + 1, aggregate)
+            self._get_node_log(child, depth + 1, aggregate, depth_max = depth_max)
             for child
             in node.children
+            if self._get_node_log(child, depth + 1, aggregate, depth_max = depth_max) != [""]
             ]
 
         return node_log + self._flatten(children_logs)
```
maukoquiroga, line 76:
```suggestion
        return node_log + self._compact(self._flatten(children_logs))
        
    def _compact(self, list: List[Optional[str]]) -> List[str]:
        return [item for item in list if item is not None] 
```

openfisca_core/tracers/computation_log.py
```diff
@@ -76,18 +81,18 @@ def _flatten(
             ) -> List[str]:
         return [item for _list in list_of_lists for item in _list]
```
maukoquiroga, line 82:
```suggestion
    def _flatten(
            self,
            list_of_lists: List[List[Optional[str]]],
            ) -> List[Optional[str]]:
        return [item for list in list_of_lists for item in list]
```

openfisca_core/tracers/computation_log.py
benjello: I had problems with types. that's why I used that ugly ` != [""]` ;-)
Since there are many good suggestions @maukoquiroga, I propose that you push all of them and I will react to them afterwards if you don't mind ?

maukoquiroga: Ah was it about types, sorry to read that.

openfisca_core/scripts/openfisca_command.py
maukoquiroga: Being orthogonal means they should play well together then.

openfisca_core/tracers/computation_log.py
maukoquiroga: These are just suggestions, you can not agree with them.

openfisca_core/scripts/openfisca_command.py
benjello: >  This flag is incompatible with --aggregate."
This is thus not relevant.

openfisca_core/tracers/computation_log.py
benjello: I agree with you that it is ugly but I didn't find a proper way to avoid it so I suggested that you handle it.
BTW I will write the test ASAP.

benjello:
Thank you very much @maukoquiroga !

benjello:
> I do however, in order to integrate it, need we work on three aspects of it:
> 
>     1. Having a reproducible way of testing this. I suggest we complete https://openfisca.org/doc/openfisca-python-api/openfisca_test.html with an example of using this flag.
> 

There is no example of output in that page so I do not know what example to use.

It is actually a problem of the online doc. It does not come with easily reproductible examples ...

benjello:
@maukoquiroga : could we go on and merge (and open a related issue in the openfisca-doc) ?

maukoquiroga (approved):

PR #1151, opened 2022-09-07
Add type hints to some variables

nikhilwoodruff:
This helps IDEs like VS Code recognise and suggest class properties.

#### New features

- Add type hints for all instances of `variable_name` in function declarations.
- Add type hints for some `Simulation` and `Population` properties.


nikhilwoodruff:
Hmm, there's a lot of unexpected issues here.

MattiSG:
Indeed! I suggest you focus on [the results of `make lint`](https://github.com/openfisca/openfisca-core/runs/8223591396?check_suite_focus=true#step:5:36), which you can run locally for faster feedback 🙂

openfisca_core/entities/entity.py
```diff
@@ -17,24 +20,35 @@ def __init__(self, key, plural, label, doc):
...
+        variable: Optional[Variable]
+        entity: Entity
+
+        variable = self.get_variable(variable_name, check_existence = True)
+
+        if variable is not None:
+            entity = variable.entity
+
         # Should be this:
         # if variable_entity is not self:
```
benjello, line 47: Why not removing also these lines ?

maukoquiroga:
Weirdly the status checks do not seem to run on our side.

openfisca_core/entities/entity.py
```diff
@@ -17,24 +20,35 @@ def __init__(self, key, plural, label, doc):
...
+    def check_variable_defined_for_entity(self, variable_name: str) -> None:
+        variable: Optional[Variable]
+        entity: Entity
+
+        variable = self.get_variable(variable_name, check_existence = True)
+
+        if variable is not None:
+            entity = variable.entity
+
         # Should be this:
```
maukoquiroga, line 46:
```suggestion
-        # Should be this:
-        # if variable_entity is not self:
```

openfisca_core/entities/entity.py
maukoquiroga, line 47:
```suggestion
```

PR #1160, opened 2022-11-16
Remove constraint on packages

### LINKED ISSUE #1158: Simplify testing of new major versions
For smoke testing purposes, the Core has a dependency on both [Country Template](https://github.com/openfisca/country-template/) and [Extension Template](https://github.com/openfisca/extension-template/). This enables running their test suites with the latest Core code.

While powerful and useful, this technique leads to a major obstacle in delivering major releases of the Core. Indeed, both of these packages in turn have a dependency to the core, and this dependency is pinned to a specific major version to ensure stability.

These two points combined create a circular dependency that prevents testing major releases: if I bump the major Core version to version N, the Template dependencies bring irresolvable constraints to version N-1.

This is typically handled by creating a new branch for each repository, relaxing the version constraint, and updating the core dependency to these new branches. This means that every such change needs temporary modifications on at least three repositories, making them very impractical. See for example https://github.com/openfisca/openfisca-core/pull/989, leading to https://github.com/openfisca/country-template/pull/128 and https://github.com/openfisca/extension-template/pull/47.

A solution to enable testing from the Core with any Core version would significantly ease the delivery of new major versions, which will be needed to finally unblock many PRs and features that have been waiting for way too long.

maukoquiroga:
Fixes #1158

#### New Features

- Lighter install by removing test packages from systematic install.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@
 
 #### New Features
 
-- Introduce a maximal depth for compuation logs
+- Introduce a maximal depth for computation logs
   - Allows for limiting the depth of the computation log chain
 
 ### 35.8.6 [#1145](https://github.com/openfisca/openfisca-core/pull/1145)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 35.12.0 [#1160](https://github.com/openfisca/openfisca-core/pull/1160)
+
+#### New Features
+
+- Lighter install by removing test packages from systematic install.
+
 ### 35.11.2 [#1166](https://github.com/openfisca/openfisca-core/pull/1166)
 
 #### Technical changes
--- a/openfisca_tasks/install.mk
+++ b/openfisca_tasks/install.mk
@@ -1,19 +1,17 @@
-## Install overall dependencies
-install-deps:
-	@$(call print_help,$@:)
-	@python -m pip install --upgrade pip twine wheel
-	@python -m pip install --upgrade --no-dependencies openfisca-country-template
-	@python -m pip install --upgrade --no-dependencies openfisca-extension-template
-
-## Install project dependencies.
-install: install-deps
-	@$(call print_help,$@:)
-	@python -m pip install --upgrade --editable ".[dev]"
-
-## Uninstall project dependencies.
+## Uninstall project's dependencies.
 uninstall:
 	@$(call print_help,$@:)
-	@python -m pip freeze | grep -v "^-e" | sed "s/@.*//" | xargs python -m pip uninstall -y
+	@pip freeze | grep -v "^-e" | sed "s/@.*//" | xargs pip uninstall -y
+
+## Install project's overall dependencies
+install-deps:
+	@$(call print_help,$@:)
+	@pip install --upgrade pip twine wheel
+
+## Install project's development dependencies.
+install-edit:
+	@$(call print_help,$@:)
+	@pip install --upgrade --editable ".[dev]"
 
 ## Delete builds and compiled python files.
 clean:
--- a/openfisca_tasks/install.mk
+++ b/openfisca_tasks/install.mk
@@ -1,13 +1,19 @@
-## Install project dependencies.
-install:
+## Install overall dependencies
+install-deps:
 	@$(call print_help,$@:)
-	@pip install --upgrade pip twine wheel
-	@pip install --editable ".[dev]" --upgrade
+	@python -m pip install --upgrade pip twine wheel
+	@python -m pip install --upgrade --no-dependencies openfisca-country-template
+	@python -m pip install --upgrade --no-dependencies openfisca-extension-template
+
+## Install project dependencies.
+install: install-deps
+	@$(call print_help,$@:)
+	@python -m pip install --upgrade --editable ".[dev]"
 
 ## Uninstall project dependencies.
 uninstall:
 	@$(call print_help,$@:)
-	@pip freeze | grep -v "^-e" | sed "s/@.*//" | xargs pip uninstall -y
+	@python -m pip freeze | grep -v "^-e" | sed "s/@.*//" | xargs python -m pip uninstall -y
 
 ## Delete builds and compiled python files.
 clean:
--- a/openfisca_tasks/lint.mk
+++ b/openfisca_tasks/lint.mk
@@ -11,7 +11,7 @@ check-syntax-errors: .
 ## Run linters to check for syntax and style errors.
 check-style: $(shell git ls-files "*.py")
 	@$(call print_help,$@:)
-	@flake8 $?
+	@python -m flake8 $?
 	@$(call print_pass,$@:)
 
 ## Run linters to check for syntax and style errors in the doc.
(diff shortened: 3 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+### 36.0.0 [#1160](https://github.com/openfisca/openfisca-core/pull/1160)
+
+#### Breaking changes
+
+- This pull request changes the definition of what is to be considered as a breaking change in a non-retro-compatible way.
+- Before, a breaking change in core automatically forced a breaking change in all it's dependants.
+- Now, a breaking change in core is a breaking change in any dependant relying on actually deprecated features and not just a major version pump in core.
```
benjello, line 9:
```suggestion
- Now, a breaking change in core is a breaking change in any dependant relying on actually deprecated features and not just a major version bump in core.  
```

benjello (changes requested):

openfisca_tasks/lint.mk
```diff
@@ -11,7 +11,7 @@ check-syntax-errors: .
 ## Run linters to check for syntax and style errors.
 check-style: $(shell git ls-files "*.py")
 	@$(call print_help,$@:)
-	@flake8 $?
+	@python -m flake8 $?
```
MattiSG, line 14: Unless I am mistaken and this is really needed for “removing constraint on packages”, this is a [rider](https://en.wikipedia.org/wiki/Rider_(legislation)).
In order to ease review, focus discussion and avoid blocking otherwise useful improvements, please make sure this PR only contains changes that match its description.

openfisca_tasks/install.mk
```diff
@@ -1,13 +1,19 @@
+## Install overall dependencies
+install-deps:
```
MattiSG, line 2: If I understand correctly, `openfisca-country-template` and `openfisca-extension-template` are only needed for testing. If that is indeed the case, then the task should be renamed and the above comment corrected:
```suggestion
# Enable regression testing with template repositories
install-test-deps:
```

openfisca_tasks/install.mk
```diff
@@ -1,13 +1,19 @@
...
+	@$(call print_help,$@:)
+	@python -m pip install --upgrade pip twine wheel
+	@python -m pip install --upgrade --no-dependencies openfisca-country-template
+	@python -m pip install --upgrade --no-dependencies openfisca-extension-template
+
 ## Install project dependencies.
-install:
+install: install-deps
```
MattiSG, line 9: If the template packages are only needed for testing, should we really install them systematically?

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+### 36.0.0 [#1160](https://github.com/openfisca/openfisca-core/pull/1160)
+
+#### Breaking changes
+
+- This pull request changes the definition of what is to be considered as a breaking change in a non-retro-compatible way.
+- Before, a breaking change in core automatically forced a breaking change in all it's dependants.
+- Now, a breaking change in core is a breaking change in any dependant relying on actually deprecated features and not just a major version bump in core.
```
MattiSG, line 9: Why is this a breaking change for reusers? 🙂 In my view this is a minor bump with new feature: lighter install by removing test packages from systematic install.

setup.py
```diff
@@ -40,15 +40,13 @@
     'flake8-print >= 3.1.0, < 4.0.0',
     'flake8-rst-docstrings == 0.2.3',
     'mypy == 0.910',
-    'openfisca-country-template >= 3.10.0, < 4.0.0',
-    'openfisca-extension-template >= 1.2.0rc0, < 2.0.0',
```
MattiSG, line 44: Have you tried:

```suggestion
     'openfisca-country-template --no-dependencies',
     'openfisca-extension-template --no-dependencies',
```

? 🙂

MattiSG: If that does not work, I would find it useful to add a comment in this dependencies list explaining that those will be installed with `make`, in order to make them discoverable.

MattiSG (changes requested):
Thanks @maukoquiroga for this exciting proposal!

- It relies on the fact that the two templates do not have any dependency that would not be provided by the Core. At this stage (and for the foreseeable future), this is the case: they both only depend on Core. However, this should be documented at least in a comment.
- The current README install procedure for development mode suggests `pip install --editable .[dev] --use-deprecated=legacy-resolver`. If I understand correctly, one would now need to use `make install` (which is consistent with https://github.com/openfisca/openfisca-core/issues/1040), so this should be updated. A workaround would be to make sure that `make test` installs these test dependencies.
- We lose the versioning of templates in testing, which means that builds are not reproducible: they might pass or fail one day and yield a different result with a different Template version. This is a regression, but I find it an acceptable cost considering that this is an edge case because of the slow change pace of Template.
- Can you please confirm that this solves https://github.com/openfisca/openfisca-core/issues/1158, and link it appropriately if that's the case? 🙂

maukoquiroga:
> Thanks @maukoquiroga for this exciting proposal!
> 
>     * It relies on the fact that the two templates do not have any dependency that would not be provided by the Core. At this stage (and for the foreseeable future), this is the case: they both only depend on Core. However, this should be documented at least in a comment.

Correct. This can either be documented in a comment, or by creating a lock file in the packages (I'm doing that now).

> 
>     * The current README install procedure for development mode suggests `pip install --editable .[dev] --use-deprecated=legacy-resolver`. If I understand correctly, one would now need to use `make install` (which is consistent with #1040), so this should be updated. A workaround would be to make sure that `make test` installs these test dependencies.

Yes, do don't use the legacy resolver anymore 🥳  I'll update the README.

> 
>     * We lose the versioning of templates in testing, which means that builds are not reproducible: they might pass or fail one day and yield a different result with a different Template version. This is a regression, but I find it an acceptable cost considering that this is an edge case because of the slow change pace of Template.

That is fixable with a bit of tooling, but we can leave that for a next pull request.

However, for now, I think it is an acceptable cost.

> 
>     * Can you please confirm that this solves #1158, and link it appropriately if that's the case? 🙂

maukoquiroga:
>     * Can you please confirm that this solves #1158, and link it appropriately if that's the case? 🙂

Only partially.

There are at least two more pull requests required, one in country, and one in extension, to confirm all works as expected.

maukoquiroga:
> > ```
> > * Can you please confirm that this solves #1158, and link it appropriately if that's the case? 🙂
> > ```
> 
> Only partially.
> 
> There are at least two more pull requests required, one in country, and one in extension, to confirm all works as expected.

UPDATE: yes, it actually fixes the aforesaid issue.

openfisca_tasks/install.mk
maukoquiroga: Well, I assume people installing from source using make are in that use case.

setup.py
maukoquiroga: It does not work. I'll leave it as a comment then,

benjello (approved):

openfisca_tasks/publish.mk
```diff
@@ -1,11 +1,11 @@
 .PHONY: build
 
-## Install openfisca-core for deployment and publishing.
+## Build & install openfisca-core for deployment and publishing.
 build:
 	@## This allows us to be sure tests are run against the packaged version
 	@## of openfisca-core, the same we put in the hands of users and reusers.
 	@$(call print_help,$@:)
-	@pip install --upgrade pip build twine
```
MattiSG, line 8: Why do we remove the upgrade of `pip` and `twine` from here?

maukoquiroga: Because they're not strictly needed for building.

maukoquiroga:
> I do not understand why we remove installing and upgrading `pip` and `twine` from the `build` step, and believe this could lead to errors when `pip` is too old on the host. However, I will approve this PR to unblock it, on the grounds that CI already tested this to work. I would still appreciate an explanation on the reasoning behind removing these upgrades, which I consider a rider.

They are still being run. Yet, as the installation of dependencies was split, without this change they would have been installed twice instead of just once.

MattiSG:
For reference, installing of `twine` is made more properly as an extra dependency group in #1168.

Issue #1159, opened 2022-11-15
Use release candidates instead of peer reviews for approval of large changesets


MattiSG:
### Problem statement

Very large changesets to the codebase are necessary when implementing major overhauls, even when they have no functional impact. For example, when refactoring vocabulary or architecture, or when changing formatting.

Since 2016 and until now, mandatory peer reviews have been the way in which quality is managed. While this system has demonstrated its power in eliminating the vast majority of regressions and broken production builds as well as in sharing code knowledge, it also has a cost and tends to slow down changes.

In the specific case of very large changesets, reviews might be so expensive (> 4 hours review for a PR) that they wait for very long before being done: they are always postponed by core staff having to handle more urgent matters, and never taken by volunteers for whom it is too costly.
As time passes, the changeset becomes irrelevant and needs major investment in rebases, worsening the problem as previous reviews have to be dismissed.
This leads to major overhauls and necessary maintenance never being deployed, with PRs waiting for months if not years and creating a deadlock of obsolete dependencies and architecture.

### State of the art

Different approaches have been tried and failed:

- increasing investment from sponsors on specific topics (more PRs are opened but not more are merged);
fast dirty reviews (author usually does not dare to act upon them);
- RFCs / specification (the implementation still needs to be reviewed).

### Proposal

I instead suggest a novel approach for OpenFisca: deploying release candidates (RC), that is versions of the package that are in the next major version, but not automatically installed, and asking for country packages to test these RCs.

1. A pull request is opened for a changeset of at least _L_ [SLoC](https://en.wikipedia.org/wiki/Source_lines_of_code).
2. At least _P_ people (author and _P - 1_ reviewers or _P_ reviewers) comment that:
    - The changeset is too large for manual review.
    - Ask for activation of the pre-release process.
    - Commit to trying the pre-release on a specific country package.
3. The version number is marked as pre-release according to [SemVer§9](https://semver.org/#spec-item-9), with suffix `-rc.X`, where X is the iteration number.
4. All automated tests pass.
5. The pre-release is deployed manually.
6. Reviewers approve or request changes to the changeset based on the result of their testing of the pre-release on the country package they committed to testing with.
7. Once at least _Q_ people out of the _P_ have approved the PR and none have requested changes, it is considered valid.
8. The version number is changed to remove the pre-release marking.
9. The pull request is merged and automatically deployed.

### Process

I open this issue to foster a discussion and generate consensus. I expect participants to either:

[...]

PR #1167, opened 2022-11-30
Add typing to data storage module

maukoquiroga:
#### Technical changes

- Add typing to `data_storage`.


openfisca_core/types/_domain.py
```diff
@@ -1,3 +1,20 @@
...
+the sum of their properties. If two data objects hold the same identifier, even
+if the data they hold is different, they are equal but not fungible.
+
+Examples:
+    >>> from openfisca_core import entities
+
+    >>> this = entities.Entity(1, "a", "b", "c")
+    >>> that = entities.Entity(1, "d", "f", "g")
+    >>> this == that
+    True
```
benjello, line 14: This is not clear to me (but I am not an expert).
Could you elaborate more so I can understand what you are achieving with this lines of code ?

maukoquiroga: Hi! The idea is that, as long as an entity and another entity have the same id, they're equal, even if they have different, let's say, documentation.

openfisca_core/data_storage/on_disk_storage.py
```diff
@@ -1,85 +1,124 @@
...
+    """Class responsible for storing and retrieving calculated vectors on disk.
+
+    Attributes:
+        _enums: ?
+        _files: ?
+        is_eternal: ?
+        storage_dir: Path to store calculated vectors.
+        preserve_storage_dir: ?
+
+    Args:
```
MattiSG, line 26: What is the `args` of a class?

MattiSG (commented):
This improves documentation significantly, seems to add interesting value checks, and adds types in a way that I don't feel confident checking.

I invoke https://github.com/openfisca/openfisca-core/issues/1159 pre-approval and am willing to try this changeset on the Country Template.

openfisca_core/data_storage/on_disk_storage.py
maukoquiroga: The constructor arguments: https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html

openfisca_core/types/_domain.py
benjello: How is it used ? Could you provide an example ?

openfisca_core/entities/entity.py
```diff
@@ -43,7 +49,7 @@ def check_variable_defined_for_entity(self, variable_name: str) -> None:
         if variable is not None:
             entity = variable.entity
 
-        if entity.key != self.key:
+        if self != entity:
```
maukoquiroga, line 52: @benjello here.

maukoquiroga:
@benjello What do you think?

openfisca_core/holders/holder.py
```diff
@@ -63,43 +125,64 @@ def clone(self, population):
...
+            value = store.get(period)
+
+            if value is not None:
+                return value
+
+        return None
 
     def get_memory_usage(self) -> MemoryUsage:
-        """Get data about the virtual memory usage of the Holder.
+        """Gets data about the virtual memory usage of the Holder.
```
benjello, line 185: I would use [imperative](https://stackoverflow.com/questions/59902102/why-is-imperative-mood-important-for-docstrings) here and in other places

maukoquiroga:
> Hi @maukoquiroga : I am not sure I can review this PR since I am not qualified enough. May be @maukoquiroga @sandcha @benoit-cty or @eraviart can give you a better review than myself.
> 
> If you point me to some specific problem where I can actually help, I would do it.

As it is, it is mainly documentation (moving the data storage is a breaking change, but can't be downplayed by doing a doublwe import). 

However, I'd like to have your opinion in where this could go, maybe in a second pull request, or in this one:

1. Packaging together Holder and Storage.
2. Order storage to make it more easily extensible.

In principle, everything that is way too beyond documentation (including types) and refactoring should maybe extracted into a second pull request, so this one can be patch or minor.

In terms of `style`, I'm trying everywhere to apply [[STYLEGUIDE.md]] (wip).

benoit-cty:
Hello Mauko, Can you run OpenFisca-France tests with this version of Core ?

PR #1171, opened 2022-12-06
Introduce a way to run tests in parallel

### LINKED ISSUE #1025: Introduce a way to run test in parallel

maukoquiroga:
Fixes #1025 

#### Technical changes

- Introduce a way to run test in parallel.
- Check `openfisca test --help` for usage instructions.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -60,7 +60,7 @@ jobs:
           key: build-${{ env.pythonLocation }}-${{ hashFiles('setup.py') }}-${{ github.sha }}
 
       - name: Run openfisca-core tests
-        run:  make test-core
+        run:  make test-core openfisca_args="--workers auto"
 
       - name: Submit coverage to Coveralls
         run: |
@@ -89,7 +89,7 @@ jobs:
           key: build-${{ env.pythonLocation }}-${{ hashFiles('setup.py') }}-${{ github.sha }}
 
       - name: Run Country Template tests
-        run: make test-country
+        run: make test-country openfisca_args="--workers auto"
 
   test-extension-template:
     runs-on: ubuntu-20.04
@@ -113,7 +113,7 @@ jobs:
           key: build-${{ env.pythonLocation }}-${{ hashFiles('setup.py') }}-${{ github.sha }}
 
       - name: Run Extension Template tests
-        run: make test-extension
+        run: make test-extension openfisca_args="--workers auto"
 
   check-numpy:
     runs-on: ubuntu-20.04
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -60,7 +60,7 @@ jobs:
           key: build-${{ env.pythonLocation }}-${{ hashFiles('setup.py') }}-${{ github.sha }}
 
       - name: Run openfisca-core tests
-        run:  make test-core openfisca_args="--workers auto"
+        run:  make test-core
 
       - name: Submit coverage to Coveralls
         run: |
@@ -89,7 +89,7 @@ jobs:
           key: build-${{ env.pythonLocation }}-${{ hashFiles('setup.py') }}-${{ github.sha }}
 
       - name: Run Country Template tests
-        run: make test-country openfisca_args="--workers auto"
+        run: make test-country
 
   test-extension-template:
     runs-on: ubuntu-20.04
@@ -113,7 +113,7 @@ jobs:
           key: build-${{ env.pythonLocation }}-${{ hashFiles('setup.py') }}-${{ github.sha }}
 
       - name: Run Extension Template tests
-        run: make test-extension openfisca_args="--workers auto"
+        run: make test-extension
 
   check-numpy:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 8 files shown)


.github/workflows/workflow.yml
```diff
@@ -60,7 +60,7 @@ jobs:
           key: build-${{ env.pythonLocation }}-${{ hashFiles('setup.py') }}-${{ github.sha }}
 
       - name: Run openfisca-core tests
-        run:  make test-core
+        run:  make test-core openfisca_args="--workers auto"
```
MattiSG, line 63: Does this make sense when there is always only one core in CI?

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+### 37.0.1 [#1171](https://github.com/openfisca/openfisca-core/pull/1171)
+
+#### Technical changes
+
+- Introduce a way to run test in parallel.
+- Check `openfisca test --help` for usage instructions.
```
MattiSG, line 8:
```suggestion
- Introduce option `--workers` to `openfisca test to run tests in parallel`.
  - Check `openfisca test --help` for usage instructions.
- Add short option `-g` to `openfisca` as a synonym to `--performance-graph`.
```

setup.py
```diff
@@ -57,12 +57,13 @@
     'mypy == 0.910',
     'pycodestyle >= 2.8.0, < 2.9.0',
     'pylint == 2.10.2',
+    'pytest-parallel < 1.0.0',
```
MattiSG, line 60: The current version is `0.1.1`. According to SemVer, this means [no warranties are made](https://semver.org/#spec-item-4) on the API. I am thus in favour of locking the minor version.
```suggestion
    'pytest-parallel < 0.2.0',
```

openfisca_core/tools/test_runner.py
```diff
@@ -138,6 +140,9 @@ def run_tests(
     if options.get('verbose'):
         argv.append('--verbose')
 
+    if workers is not None:
+        argv.append(f"--workers={workers[0]}")
```
MattiSG, line 144: Why not default to `auto` and let users override that to `1` if they want so? 🙂 I don't see a reason to not use it if it is safe, and I don't see a reason to offer (let alone use) it if it is not safe.

MattiSG (commented):
In https://github.com/openfisca/openfisca-core/issues/1025, @benoit-cty mentioned issues with `Xdist` that were likely to come up with `pytest-parallel`. Do we not have them here?

The suggested library is not maintained: https://github.com/kevlened/pytest-parallel/issues/104. Our stack is already suffering from outdated dependencies and I don't find it wise to add to this.

What is the performance gain we get from this implementation on some large models?

.github/workflows/workflow.yml
maukoquiroga: I think there are two actually. Haven't checked though.

openfisca_core/tools/test_runner.py
maukoquiroga: Yes workers are safe, yet we need to see how/if they work properly on windows. Otherwise I agree with you.

What really brings improvement however are to run several threads per process, yet out tests are not thread-safe :/

maukoquiroga:
> In #1025, @benoit-cty mentioned issues with `Xdist` that were likely to come up with `pytest-parallel`. Do we not have them here?
> 
> The suggested library is not maintained: [kevlened/pytest-parallel#104](https://github.com/kevlened/pytest-parallel/issues/104). Our stack is already suffering from outdated dependencies and I don't find it wise to add to this.
> 
> What is the performance gain we get from this implementation on some large models?

Fair enough. Performance is only remarkable for county packages like France. Implementing `xdist` requires changes to the test command. I propose we deal with those, needed nonetheless to upgrade to newest `pytest`.

PR #1179, opened 2022-12-19
Fix dependency conflicts

### LINKED ISSUE #1176: Missing importlib_metadata requirement

maukoquiroga:
Fixes #1176 

#### Bug fix

- Do not install dependencies outside the `setup.py`
  - Dependencies installed outside the `setup.py` are not taken into account by `pip`'s dependency resolver.
  - In case of conflicting transient dependencies, the last library installed will "impose" its dependency version.
  - This makes the installation and build of the library non-deterministic and prone to unforeseen bugs caused by external changes in dependencies' versions.

#### Note

A definite way to solve this issue is to clearly separate library dependencies (with a `virtualenv`) and a universal dependency installer for CI requirements (like `pipx`), taking care of:

1. Always running tests inside the `virtualenv` (for example with `nox`).
2. Always building outside of the `virtualenv` (for example with `poetry` installed by `pipx`).

Moreover, it is indeed even better to have a lock file for dependencies, similar to what is it proposed in #1015 (with `pip freeze`) or with tools providing such features (`poetry`, `pipenv`, etc.).


benjello (commented):
Looks good to me. But I am reluctant to go beyond comment and approve this PR since I am not an expert and I already done enough mistakes lately to take the risk so early ;-)

maukoquiroga:
Would be great if you could test it @eraviart . I did test it on the New Zealand's package and seems to work fine.

eraviart (approved):

maukoquiroga:
Thanks @eraviart . As I elaborate in https://github.com/openfisca/openfisca-core/issues/1176#issuecomment-1364935194, it does not make sense to me to yank 38.0.2 as it actually unblocks the problem introduced in 35.11.1 regarding `importlib-metadata`. Because we've already released several majors since then, it doesn't seem practical to yank all of these versions either, unless there is a consensus to do so. Therefore, I will modify the CHANGELOG of this PR to reflect that, by recommending using this version, but removing the `has been yanked` part.

maukoquiroga:
Changelog updated to reflect that `38.0.2` actually restored a working publication.

PR #1181, opened 2023-03-22
Upgrade Python & use the latest version of each dependency

eraviart:
#### Breaking changes

- Upgrade every dependencies to their latest versions.
- Upgrade to Python >= 3.9


eraviart:
Currently, the ArrayLike type in `types/_data.py` has been relaxed and the typing tests have been deactivated, because the monkey patching of ArrayLike no more works and fixing it is way above my pay grade.

eraviart:
> Currently, the ArrayLike type in `types/_data.py` has been relaxed and the typing tests have been deactivated, because the monkey patching of ArrayLike no more works and fixing it is way above my pay grade.

Furthermore the `check-types` & `lint-typing-strict` tests have been commented out, because they detect dozens of problems in OpenFisca typings. Since these typings are some form of sugar and nobody in LexImpact is currently able to fix them, we would like to merge this MR without these tests

benjello (approved):

maukoquiroga:
Thanks @eraviart ! The monkey patch is useless now that we bumped Numpy up. I'll clean that up at some point.

PR #1182, opened 2023-05-03
Fix YamlItem in test_runner.py to clone TBS in case of reforms

Allan-CodeWorks:
#### Technical changes

- Method `_get_tax_benefit_system()` of class `YamlItem` in file  `openfisca_core/tools/test_runner.py` will now clone the TBS when applying reforms to avoid running tests with previously reformed TBS.

#### Notes
In the case we want to run **2 tests or more** with differents `reforms` and no `extensions` we were sharing the same baseline so we were executing following tests with the TBS set in previous tests.

### DIFF
--- a/tests/core/tools/test_runner/test_yaml_runner.py
+++ b/tests/core/tools/test_runner/test_yaml_runner.py
@@ -92,6 +92,12 @@ def test_variable_not_found():
     assert excinfo.value.variable_name == "unknown_variable"
 
 
+def test_tax_benefit_systems_with_cache():
+    baseline = TaxBenefitSystem()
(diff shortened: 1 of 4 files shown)


tests/core/tools/test_runner/test_yaml_runner.py
```diff
@@ -22,7 +22,7 @@ def get_package_metadata(self):
         return {"name": "Test", "version": "Test"}
 
     def apply_reform(self, path):
-        return Reform(self)
```
guillett, line 25: :scream:

guillett (approved):

PR #1186, opened 2023-06-26
Skip type-checking tasks

maukoquiroga:
Related to #1181 

#### Technical changes

- Skip type-checking tasks
  - Before their definition was commented out but still run with `make test`
  - Now they're skipped but not commented, which is needed to fix the underlying issues


maukoquiroga:
LGTM @benjello ?

benjello (approved):

PR #1187, opened 2023-06-27
Document Role

maukoquiroga:
#### Technical changes

- Document `Role`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+###  41.0.1 [#1187](https://github.com/openfisca/openfisca-core/pull/1187)
+
+#### Technical changes
+
+- Document `Role`.
+
 # 41.0.0 [#1189](https://github.com/openfisca/openfisca-core/pull/1189)
 
 #### Breaking changes
@@ -12,8 +18,8 @@ The Web API was very prone to crashing, timeouting at startup because of the tim
 
 #### New Features
 
-* Allows for dispatching and dividing inputs over a broader range.
-  * For example, divide a monthly variable by week.
+- Allows for dispatching and dividing inputs over a broader range.
+  - For example, divide a monthly variable by week.
 
 ###  40.0.1 [#1184](https://github.com/openfisca/openfisca-core/pull/1184)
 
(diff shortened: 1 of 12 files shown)


MattiSG (commented):
Thanks @maukoquiroga!

I can't evaluate the relevance of the documentation itself. @benjello maybe?

benjello:
@maukoquiroga could you run the openfisca-france test suite with this version of core so we stress test it in real conditions and avoid some problem down the roads.

maukoquiroga:
@benjello :

> 2380 passed, 313 warnings in 873.19s (0:14:33)

maukoquiroga:
> Thanks @maukoquiroga!
> 
> I can't evaluate the relevance of the documentation itself. @benjello maybe?

Thanks @MattiSG . The documentation contains unit tests, maybe you could have an input on that.

MattiSG (approved):
At this stage, I see mostly syntax improvements, and addition of documentation. I prefer to switch to a shipping mindset and unblock this PR by applying the idea that some documentation is better than no documentation, and will approve even though I cannot guarantee the accuracy of the documentation contents themselves.

Thank you for your work and patience @maukoquiroga.

maukoquiroga:
Thanks @MattiSG :) With lots of syntax improvements and doc, my hope is it'll be easier to propose and review big changes in the future without breaking things unadvertently.

benjello:
@maukoquiroga : do you think this PR may produce this [bug](https://github.com/openfisca/openfisca-survey-manager/actions/runs/6381132067/job/17317034553#step:6:222) ?

maukoquiroga:
Hi @benjello, at first view, no (but I encountered a similar issue while working in this PR). While #1194 is a first fix, `Role` continues to be non-hashable because its immutability can't be assured. Thanks for the fix!

benjello:
#1194 fixed the trouble with the `openfisca-survey-manager` bug

maukoquiroga:
After checking, in fact yes, sorry: this PR makes Role unhashable by giving it an `__eq__`. If a `Role` is intended to actually be hashable, then indeed `__eq__` should be rermoved.

maukoquiroga:
I'll propose a PR in 5m. I'll just add a test in the doc so we know `Role` is meant to be hashable. No need for the `__hash__` method, just removing `__eq__`.

PR #1188, opened 2023-07-10
Remove comments extraction not used and very slow on newer python versions

guillett:
Comments extraction is not tested, doesn't seem to be used and is making the API very very very slow.

I am currently refactoring this section, it remains the source code extraction and generation but for comments it looks simpler as the logic can be removed.

#### Breaking changes

- In _Variable.get_introspection_data_:
  - `comments` removed from the tuple returned

232 seconds spent for an unused feature here:

Screenshot 2023-07-10 at 18-30-38 full_7minutes_profile3

A single long `findsource` trace (and I'm working on it, I extracted that PR from my work)

Screenshot 2023-07-10 at 18-36-42 without_comments

### DIFF
--- a/openfisca_core/variables/variable.py
+++ b/openfisca_core/variables/variable.py
@@ -364,12 +364,10 @@ class Variable:
         """
         Get instrospection data about the code of the variable.
 
-        :returns: (comments, source file path, source code, start line number)
+        :returns: (source file path, source code, start line number)
         :rtype: tuple
 
         """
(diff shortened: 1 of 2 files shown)


MattiSG (approved):
Congratulations for the performance analysis!

guillett:
I would like to merge this PR with #1189 to avoid two major version bumps. I will wait a couple of days and if it looks like it's better to split them, I'll merge them separately.

PR #1189, opened 2023-07-10
Refactor introspection data generation to get a much quicker startup time

guillett:
Python version bump made OpenFisca Web API much much slower.
It takes 5 minutes or so (cf discussion on Slack).

This slow startup is mainly due to the introspection data generation and more precisely to the python files parsing.

## before

Screenshot 2023-07-17 at 17-26-04 full_7minutes_profile3

## after

Screenshot 2023-07-17 at 17-27-49 fix_1946

**similar to #1188**

#### Breaking changes

- In _Variables_:
  - `get_introspection_data` no longer have parameters as well as calling functions

#### Technical notes

I think it is worth sharing my development strategy:

I worked with a country package (France) and started `gunicorn run --config config.py` with
> `run.py`
```python
from openfisca_core.scripts import build_tax_benefit_system
from openfisca_web_api.app import create_app


country_package = 'openfisca_france'

tax_benefit_system = build_tax_benefit_system(
    country_package_name = country_package,
    extensions = [],
    reforms = []
)

application = create_app(tax_benefit_system)
```
and
> `config.py`
```python
import cProfile
import pstats
from io import StringIO
import logging
import os
import tempfile
import time

timeout=1200

profiles = {}
def post_fork(server, worker):
    worker.start_time = time.time()
    profile = cProfile.Profile()
    profile.enable()
    profiles[worker.pid] = profile
    worker.log.info("PROFILING %d" % (worker.pid))

def worker_abort(worker):
    tf = tempfile.NamedTemporaryFile(delete = False)
    worker.log.info("PROFILING RESULT %d: http://127.0.0.1:8081/snakeviz/%s" % (worker.pid, tf.name))
    profiles[worker.pid].dump_stats(tf.name)

def post_worker_init(worker):
    tf = tempfile.NamedTemporaryFile(delete = False)
    worker.log.info("PROFILING RESULT %d: http://127.0.0.1:8081/snakeviz/%s" % (worker.pid, tf.name))
    profiles[worker.pid].dump_stats(tf.name)
```

It allowed me to get visual clues with snakeviz on areas of potential improvements (as well as slow interesting portions I could comment out).

I also cloned https://github.com/python/cpython to have a better understanding of underlying functions. However, I didn't dig much in code in C I mainly stayed in Python and I recycled code snippets in our contexts.

### DIFF
--- a/openfisca_web_api/loader/variables.py
+++ b/openfisca_web_api/loader/variables.py
@@ -39,7 +39,7 @@ def build_source_url(
 
 
 def build_formula(
-    formula, country_package_metadata, source_file_path, tax_benefit_system
+    formula, country_package_metadata, source_file_path
 ):
     source_code, start_line_number = inspect.getsourcelines(formula)
 
@@ -59,22 +59,22 @@ def build_formula(
 
 
 def build_formulas(
-    formulas, country_package_metadata, source_file_path, tax_benefit_system
+    formulas, country_package_metadata, source_file_path
 ):
     return {
         start_date: build_formula(
-            formula, country_package_metadata, source_file_path, tax_benefit_system
+            formula, country_package_metadata, source_file_path
         )
         for start_date, formula in formulas.items()
     }
 
 
-def build_variable(variable, country_package_metadata, tax_benefit_system):
+def build_variable(variable, country_package_metadata, ):
     (
         source_file_path,
         source_code,
         start_line_number,
-    ) = variable.get_introspection_data(tax_benefit_system)
+    ) = variable.get_introspection_data()
     result = {
         "id": variable.name,
         "description": variable.label,
@@ -99,8 +99,7 @@ def build_variable(variable, country_package_metadata, tax_benefit_system):
         result["formulas"] = build_formulas(
             variable.formulas,
             country_package_metadata,
-            source_file_path,
-            tax_benefit_system,
+            source_file_path
         )
 
         if variable.end:
@@ -116,6 +115,6 @@ def build_variable(variable, country_package_metadata, tax_benefit_system):
 
 def build_variables(tax_benefit_system, country_package_metadata):
     return {
-        name: build_variable(variable, country_package_metadata, tax_benefit_system)
+        name: build_variable(variable, country_package_metadata)
         for name, variable in tax_benefit_system.variables.items()
     }
--- a/openfisca_web_api/loader/variables.py
+++ b/openfisca_web_api/loader/variables.py
@@ -71,7 +71,6 @@ def build_formulas(
 
 def build_variable(variable, country_package_metadata, tax_benefit_system):
     (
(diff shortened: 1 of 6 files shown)


openfisca_web_api/loader/variables.py
```diff
@@ -59,23 +59,22 @@ def build_formula(
...
         start_date: build_formula(
-            formula, country_package_metadata, source_file_path, tax_benefit_system
+            formula, country_package_metadata, source_file_path
         )
         for start_date, formula in formulas.items()
     }
 
 
-def build_variable(variable, country_package_metadata, tax_benefit_system):
+def build_variable(variable, country_package_metadata, ):
```
MattiSG, line 72: Is it intentional to have a trailing comma?
```suggestion
def build_variable(variable, country_package_metadata):
```

MattiSG (commented):
Thank you for this changeset!
I don't understand enough of the goals of the introspection nor of the suggested alternative code to approve or request changes. I like how less parameters are needed with that rewrite.
Could you maybe share some data on that “much quicker startup time”? 🙂

guillett:
Oh sorry, I forgot to give more contexte. I updated the initial comment.

MattiSG (commented):
Looks amazing, thank you very much for the clarifications!

I would just need the changelog and version change (to understand if this is a minor or major) to approve 🙂

MattiSG (approved):
In order to speed up the process, and after re-reading the edited first comment, approve with the following requests:

1. Make that a breaking change (major bump).
2. Rephrase changelog entry to “`get_introspection_data` no longer has parameters nor calling functions”

I am open to discussing the major vs minor if there is clear demonstration that this was not part of the public API (I am not familiar enough with the code to be certain myself).

maukoquiroga:
> Thank you for this changeset! I don't understand enough of the goals of the introspection nor of the suggested alternative code to approve or request changes. I like how less parameters are needed with that rewrite. Could you maybe share some data on that “much quicker startup time”? 🙂

@MattiSG Every variable has formulas, which are not defined in core but directly in each country package. For core to be able to use these formulas, the `Variable` class parses the AST, their internal sub-Python representation, and evaluates them, so they get defined in memory. (This is one of the reasons static analysis of OpenFisca at run-time is not possible: the formulas refered to internally have no tie with the actual code read from the packages source files).

maukoquiroga:
@guillett What does it imply that _no longer have parameters as well as calling functions_? What I could get from the changeset is that comments are no longer parsed (why were there being parsed is yet as mistery for me).

If you could add more specific information about the actual changes it would be great :)

If I'm not mistaken, @benjello @eraviart this mechanism bypasses all compile-time and interpreter optimisations. 

Another possible concern is that performance scales badly (time increases exponentially the bigger a country package).

Maybe @Morendil looked into this or @fpagnoux?

PR #1192, opened 2023-09-26
Add tests to entities

maukoquiroga:
#### Technical changes

- Add tests to `entities`.

### DIFF
--- a/openfisca_core/entities/entity.py
+++ b/openfisca_core/entities/entity.py
@@ -44,16 +44,11 @@ class Entity:
             entity = variable.entity
 
         if entity.key != self.key:
-            message = os.linesep.join(
-                [
-                    "You tried to compute the variable '{}' for the entity '{}';".format(
-                        variable_name, self.plural
-                    ),
-                    "however the variable '{}' is defined for '{}'.".format(
-                        variable_name, entity.plural
-                    ),
-                    "Learn more about entities in our documentation:",
-                    "<https://openfisca.org/doc/coding-the-legislation/50_entities.html>.",
-                ]
+            message = (
+                f"You tried to compute the variable '{variable_name}' for",
+                f"the entity '{self.plural}'; however the variable",
+                f"'{variable_name}' is defined for '{entity.plural}'.",
+                "Learn more about entities in our documentation:",
+                "<https://openfisca.org/doc/coding-the-legislation/50_entities.html>.",
             )
-            raise ValueError(message)
+            raise ValueError(os.linesep.join(message))
--- a/openfisca_core/entities/entity.py
+++ b/openfisca_core/entities/entity.py
@@ -25,7 +25,7 @@ class Entity:
 
     def check_role_validity(self, role: Any) -> None:
         if role is not None and not isinstance(role, Role):
-            raise ValueError("{} is not a valid role".format(role))
+            raise ValueError(f"{role} is not a valid role")
 
     def get_variable(
         self,
@@ -46,10 +46,10 @@ class Entity:
         if entity.key != self.key:
             message = os.linesep.join(
                 [
-                    "You tried to compute the variable '{0}' for the entity '{1}';".format(
+                    "You tried to compute the variable '{}' for the entity '{}';".format(
                         variable_name, self.plural
                     ),
-                    "however the variable '{0}' is defined for '{1}'.".format(
+                    "however the variable '{}' is defined for '{}'.".format(
                         variable_name, entity.plural
                     ),
                     "Learn more about entities in our documentation:",
--- a/openfisca_core/entities/entity.py
+++ b/openfisca_core/entities/entity.py
@@ -12,7 +12,7 @@ class Entity:
     Represents an entity (e.g. a person, a household, etc.) on which calculations can be run.
     """
 
-    def __init__(self, key, plural, label, doc):
+    def __init__(self, key: str, plural: str, label: str, doc: str) -> None:
         self.key = key
         self.label = label
         self.plural = plural
--- a/openfisca_core/entities/group_entity.py
+++ b/openfisca_core/entities/group_entity.py
@@ -35,7 +35,7 @@ class GroupEntity(Entity):
         doc: str,
         roles: Sequence[HasKey],
         containing_entities: Sequence[str] = (),
-    ) -> object:
+    ) -> None:
         super().__init__(key, plural, label, doc)
         self.roles_description = roles
         self.roles = []
--- a/openfisca_core/entities/group_entity.py
+++ b/openfisca_core/entities/group_entity.py
@@ -1,8 +1,10 @@
-from collections.abc import Sequence
+from collections.abc import Iterable, Mapping
+from typing import Any
+
+from itertools import chain
 
 from .entity import Entity
 from .role import Role
-from .typing import HasKey
 
 
 class GroupEntity(Entity):
@@ -21,10 +23,6 @@ class GroupEntity(Entity):
         containing_entities: The list of keys of group entities whose members
             are guaranteed to be a superset of this group's entities.
 
-    .. versionchanged:: 35.7.0
-        Added ``containing_entities``, that allows the defining of group
-        entities which entirely contain other group entities.
-
     """  # noqa RST301
 
     def __init__(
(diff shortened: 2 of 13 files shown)


openfisca_core/entities/group_entity.py
```diff
@@ -23,7 +27,15 @@ class GroupEntity(Entity):
...
-    def __init__(self, key, plural, label, doc, roles, containing_entities=()):
+    def __init__(
+        self,
+        key: str,
+        plural: str,
+        label: str,
+        doc: str,
+        roles: Sequence[HasKey],
+        containing_entities: Sequence[str] = (),
+    ) -> object:
```
MattiSG, line 38: Since this is a constructor, couldn't we be more specific? 🙂 
```suggestion
    ) -> GroupEntity:
```

openfisca_core/entities/group_entity.py
```diff
@@ -23,7 +27,15 @@ class GroupEntity(Entity):
...
     """  # noqa RST301
 
-    def __init__(self, key, plural, label, doc, roles, containing_entities=()):
+    def __init__(
+        self,
+        key: str,
+        plural: str,
+        label: str,
+        doc: str,
+        roles: Sequence[HasKey],
```
MattiSG, line 36: Is there any reason why the roles should be ordered? I don't see one and would recommend to use `Set` or `Iterable` instead.

openfisca_core/entities/group_entity.py
```diff
@@ -23,7 +27,15 @@ class GroupEntity(Entity):
```
MattiSG, line 27: > Can we remove the `versionchanged` information along the way? This has its place in the changelog, but not in the inline doc.

openfisca_core/entities/typing.py
```diff
@@ -1,9 +1,18 @@
+from abc import abstractmethod
 from typing import Protocol
 
 
-class Entity(Protocol):
-    """Entity protocol.
+class HasKey(Protocol):
```
MattiSG, line 5: I feel that the main value brought by the  `Entity` protocol is signalling the type by its name. I am not convinced by the value of making this generic. In my view, it makes the entities module less readable by adding a more generic name, and the protocol is so trivial it has little intrinsic value.

I thus suggest to keep the `Entity` name at this stage and rename it to a more generic name only if it gets extracted from the `entities` module.

openfisca_core/entities/tests/test_role.py
```diff
@@ -0,0 +1,24 @@
...
+
+import pytest
+
+from openfisca_core import entities
+
+from ..typing import HasKey
+
+
+@pytest.fixture
+def entity() -> Any:
```
MattiSG, line 11: Shouldn't this be
```suggestion
def entity() -> HasKey:
```
or rather, with my suggested rollback:
```suggestion
def entity() -> Entity:
```

openfisca_core/entities/tests/test_group_entity.py
```diff
@@ -0,0 +1,84 @@
...
+@pytest.fixture
+def parent() -> str:
+    """A key."""
+
+    return "parent"
+
+
+@pytest.fixture
+def uncle() -> str:
+    """Another key."""
```
MattiSG, line 19: > I'm not sure these comments add much value in the context of tests 🙂 for next time, feel free not to add them.

openfisca_core/entities/tests/test_group_entity.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def test_init_when_doc_indented() -> None:
+    """De-indent the ``doc`` attribute if it is passed at initialisation."""
+
+    key = "\tkey"
+    doc = "\tdoc"
+    group_entity = entities.GroupEntity(key, "label", "plural", doc, [])
+    assert group_entity.key == key
+    assert group_entity.doc != doc
```
MattiSG, line 66: This does not test that there is de-indentation, but that the strings are different. This approximation is too broad and can give false negatvies too easily. What about:

```suggestion
    assert group_entity.doc == doc.lstrip()
```

openfisca_core/entities/tests/test_entity.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def test_init_when_doc_indented() -> None:
+    """De-indent the ``doc`` attribute if it is passed at initialisation."""
+
+    key = "\tkey"
+    doc = "\tdoc"
+    entity = entities.Entity(key, "label", "plural", doc)
+    assert entity.key == key
+    assert entity.doc != doc
```
MattiSG, line 11: This does not test that there is de-indentation, but that the strings are different. This approximation is too broad and can give false negatvies too easily. What about:

```suggestion
    assert group_entity.doc == doc.lstrip()
```

MattiSG (commented):
Interesting changes, thank you!

openfisca_core/entities/group_entity.py
benjello: I would not change that ionce it might be used by package dependent on core (and even core) when filling population if the roles are indexed by an integer.

openfisca_core/entities/group_entity.py
maukoquiroga: This was a typo that was already there, thanks!

openfisca_core/entities/group_entity.py
maukoquiroga: After spending a couple of hours on this... good part of the code depends on roles and sub-roles being ordered.

maukoquiroga: I prefer not to touch that for now, not before the aforesaid parts of the code are properly tested and documented.

openfisca_core/entities/typing.py
maukoquiroga: Makes sense. And you're right, the main use here is to "duck-reference" classes between them. That improves static analysis and avoids run time problems (cyclic imports, inheritance, etc).

openfisca_core/entities/tests/test_group_entity.py
maukoquiroga: Can I check the linter in another PR?

maukoquiroga: Done

openfisca_core/entities/group_entity.py
maukoquiroga: BTW @MattiSG at least in Python, `iterable` is a generic protocol including `list` (ordered, mutable), `tuple` (ordered, immutable), `dict` (not-ordered, mutable), and `set` (not-ordered, sort of mutable).

@benjello Why? You mean, for example, I have a GroupEntity of 4 people. You'll use role/subrole's index to determine each person's order within the group? (I actually had strange results when I tried to remove indexation).

maukoquiroga:
@benjello This PR opened a couple of questions that I pose here for documentation purposes:

1. Why the reliance on roles/subroles indexes?
2. Why do we need bidirectional navigation entity <--> role?
3. Why do we need bidirectional navigation entity <--> TBS?
4. How does / where does the magic happen with the `members_role` projector?
5. Why defining entity.CHILD and not something like entity.roles.where(key="child")?

These are all breaking-changes, but refactoring each could greatly facilitate maintenance and new features.

maukoquiroga:
Is there anything missing for LTGM @benjello @MattiSG?

openfisca_core/entities/entity.py
```diff
@@ -46,10 +46,10 @@ def check_variable_defined_for_entity(self, variable_name: str) -> None:
         if entity.key != self.key:
             message = os.linesep.join(
                 [
-                    "You tried to compute the variable '{0}' for the entity '{1}';".format(
+                    "You tried to compute the variable '{}' for the entity '{}';".format(
```
benjello, line 49: What about using f-stings here too ?

openfisca_core/entities/entity.py
```diff
@@ -46,10 +46,10 @@ def check_variable_defined_for_entity(self, variable_name: str) -> None:
...
                         variable_name, self.plural
                     ),
-                    "however the variable '{0}' is defined for '{1}'.".format(
+                    "however the variable '{}' is defined for '{}'.".format(
```
benjello, line 52: What about using f-stings here too ?

benjello (approved):
Some suggestions left to the author of the PR to adopt or not.

PR #1195, opened 2023-10-03
Make role hashable

maukoquiroga:
#### Technical changes

- Make `Role` hashable
- Details:
    - By introducing `__eq__`, naturally `Role` became unhashable, because equality was calculated based on a property of `Role` (`role.key == another_role.key`), and no longer structurally (`"1" == "1"`).
    - This changeset removes `__eq__`, as `Role` is being used downstream as a hasheable object, and adds a test to ensure `Role`'s hasheability

### DIFF
--- a/openfisca_core/entities/role.py
+++ b/openfisca_core/entities/role.py
@@ -41,11 +41,11 @@ class Role:
         >>> str(role)
         'Role(parent)'
 
+        >>> {role}
+        {Role(parent)}
+
         >>> role.key
         'parent'
-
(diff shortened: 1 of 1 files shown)


openfisca_core/entities/role.py
```diff
@@ -41,12 +41,12 @@ class Role:
         >>> str(role)
         'Role(parent)'
 
+        >>> {role}
```
maukoquiroga, line 44: @benjello

Issue #1198, opened 2023-10-10
Upgrade Gunicorn Dependency to Resolve SIGTERM Handling Issue


baptou12:
Hi there!

I really enjoy OpenFisca, but I recently encountered an issue.

### Here is what I did:

We recently discovered that when a SIGTERM was sent to openfisca web-api, it could raise some errors : 

<details>
<summary>Stacktrace</summary>

```
--- Logging error ---
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/logging/__init__.py", line 1113, in emit
    stream.write(msg + self.terminator)
RuntimeError: reentrant call inside <_io.BufferedWriter name='<stderr>'>
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/logging/__init__.py", line 1113, in emit
    stream.write(msg + self.terminator)
  File "/home/main/aides_jeunes/venv/lib/python3.11/site-packages/gunicorn/arbiter.py", line 242, in handle_chld
    self.reap_workers()
  File "/home/main/aides_jeunes/venv/lib/python3.11/site-packages/gunicorn/arbiter.py", line 530, in reap_workers
    self.log.warning(
  File "/home/main/aides_jeunes/venv/lib/python3.11/site-packages/gunicorn/glogging.py", line 261, in warning
    self.error_log.warning(msg, *args, **kwargs)
  File "/usr/local/lib/python3.11/logging/__init__.py", line 1501, in warning
    self._log(WARNING, msg, args, **kwargs)
  File "/usr/local/lib/python3.11/logging/__init__.py", line 1634, in _log
    self.handle(record)
  File "/usr/local/lib/python3.11/logging/__init__.py", line 1644, in handle
    self.callHandlers(record)
  File "/usr/local/lib/python3.11/logging/__init__.py", line 1706, in callHandlers
    hdlr.handle(record)
  File "/usr/local/lib/python3.11/logging/__init__.py", line 978, in handle
    self.emit(record)
  File "/usr/local/lib/python3.11/logging/__init__.py", line 1118, in emit
    self.handleError(record)
  File "/usr/local/lib/python3.11/logging/__init__.py", line 1031, in handleError
    sys.stderr.write('--- Logging error ---\n')
RuntimeError: reentrant call inside <_io.BufferedWriter name='<stderr>'>
Call stack:
  File "/home/main/aides_jeunes/venv/bin/gunicorn", line 8, in <module>
[...]

PR #1199, opened 2023-10-10
chore: bump gunicorn version to `~21`

baptou12:
fix: https://github.com/openfisca/openfisca-core/issues/1198

#### Technical changes

- Bump Gunicorn version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 41.2.0 [#1199](https://github.com/openfisca/openfisca-core/pull/1199)
+
+#### Technical changes
+
+- Fix `openfisca-core` Web API error triggered by `Gunicorn` < 22.0.
+  - Bump `Gunicorn` major revision to fix error on Web API.
+    Source: https://github.com/benoitc/gunicorn/issues/2564
+
 ### 41.1.2 [#1192](https://github.com/openfisca/openfisca-core/pull/1192)
 
 #### Technical changes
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+### 41.2.0 [#1199](https://github.com/openfisca/openfisca-core/pull/1199)
```
maukoquiroga, line 3:
```suggestion
## 41.2.0 [#1199](https://github.com/openfisca/openfisca-core/pull/1199)
```

setup.py
```diff
@@ -43,7 +43,7 @@
 api_requirements = [
     "Flask >=2.2.3, < 3.0",
     "Flask-Cors >=3.0.10, < 4.0",
-    "gunicorn >=20.1.0, < 21.0",
+    "gunicorn >=20.1.0, < 22.0",
```
maukoquiroga, line 46: If the error was fixed in the latest major, I propose this:

```suggestion
    "gunicorn >=21.0, < 22.0",
```

Excepting `numpy`, it is preferable to avoid requirements spanning more than one major.

maukoquiroga: Also, now that I think of it, without the contraint, if another library happens to require a lower version of gunicorn, the `pip` resolver can silently decide to use a version where the bug is not yet corrected.

baptou12: Thx, I made the change

maukoquiroga (approved):
Top !

PR #1202, opened 2023-12-08
fix: require fully specifing entities when expanding axes

bonjourmauko:
#### Technical changes

- Check that ententies are fully specified when expanding over axes.


openfisca_core/simulations/simulation_builder.py
```diff
@@ -145,17 +207,17 @@ def build_from_entities(self, tax_benefit_system, input_dict):
...
+    def build_from_variables(
+        self, tax_benefit_system: TaxBenefitSystem, input_dict: AbbrParams
+    ) -> Simulation:
         """
         Build a simulation from a Python dict ``input_dict`` describing variables values without expliciting entities.
 
         This method uses :any:`build_default_simulation` to infer an entity structure
 
         Example:
+            >>> {'salary': {'2016-10': 12000}}
```
Morendil, line 219: Nitpick: why remove the function call from the example ? As I understand it example in docstrings are executable and this is often helpful (if somewhat redundant with a good test suite).

openfisca_core/simulations/simulation_builder.py
```diff
@@ -52,24 +75,36 @@ def build_from_dict(self, tax_benefit_system, input_dict):
...
         """
         Build a simulation from a Python dict ``input_dict`` fully specifying entities.
 
         Examples:
 
-        >>> simulation_builder.build_from_entities({
-            'persons': {'Javier': { 'salary': {'2018-11': 2000}}},
-            'households': {'household': {'parents': ['Javier']}}
-            })
+            >>> {
```
Morendil, line 104: Same remark (nitpick) - why remove the executable part of this example ?

openfisca_core/simulations/_type_guards.py
```diff
@@ -0,0 +1,22 @@
...
+from typing_extensions import TypeGuard
+
+from .typing import AbbrParams, AxesParams, ExplParams, ImplParams, Params
+
+
+def is_abbr_spec(params: Params, items: Iterable[str]) -> TypeGuard[AbbrParams]:
+    return any(key in items for key in params.keys()) or params == {}
+
+
+def is_impl_spec(params: Params, items: Iterable[str]) -> TypeGuard[ImplParams]:
```
Morendil, line 13: I don't really understand what this function does - what seems awkward about it is that it's named like a predicate, but returns a set operation.

openfisca_core/simulations/simulation_builder.py
```diff
@@ -84,11 +119,17 @@ def build_from_entities(self, tax_benefit_system, input_dict):
...
 
         helpers.check_type(input_dict, dict, ["error"])
-        axes = input_dict.pop("axes", None)
+
+        axes: list[list[AxisParams]] | None = None
+
+        if is_axes_spec(input_dict):
+            axes = input_dict.pop("axes")
+
+        params: NoAxParams = typing.cast(NoAxParams, input_dict)
```
Morendil, line 128: This should be named NoAxisParams or NoAxesParams. An Ax is a weapon ;)

I feel a bit queasy about "No" as a suffix. For one thing we have things like NoSQL, NoEstimates and look how well they've turned out ;) Or another thing that comes to mind is how many things are called "wireless" that are not primarily characterized by the absence of wires.

Perhaps something like UnitaryParams would be a better focus on what the type *is* rather than on what it *isn't*.

Morendil (approved):
Other than some quibbles and nitpicks, and the type guards which I don't understand well enough to comment on, but don't look like they affect behaviour, I understand the logic behind theses changes and I believe they effect the intent of the PR as described on Github and in the Changelog.

openfisca_core/simulations/simulation_builder.py
bonjourmauko: The thing is that they do not work, and the linter was complaining. For this PR, we could actually leave it as it was. Ideally, however, I've been fixing doctests and adding them to the test suite, and I'd definitely do that in another PR.

openfisca_core/simulations/_type_guards.py
bonjourmauko: I'll document these functions before merging. TL; DR : [type narrowing](https://mypy.readthedocs.io/en/stable/type_narrowing.html)

openfisca_core/simulations/simulation_builder.py
bonjourmauko: Yeah that is chapter 1 of Uncle Bob's clean code : beware of name encodings. I'll change it to something more semantic.

openfisca_core/simulations/_type_guards.py
bonjourmauko: I'll document these functions before merging. TL; DR : [type narrowing](https://mypy.readthedocs.io/en/stable/type_narrowing.html)

We're asking: is this dict of that type? If the answer is yes, we narrow the type of the dict. In this particular case is very useful because it allows to "bring to life" a full range of concepts that lie in our heads: fully defined entities, explicit and implicit, etc. At runtime, these act just like a regular _isinstance_ would if these were class instances instead of plain dicts.

PR #1205, opened 2024-02-05
Add a look back constraint on spiral errors

guillett:
#### New features

- Add a look back constraint on spiral errors


openfisca_core/simulations/simulation.py
```diff
@@ -412,17 +416,12 @@ def invalidate_cache_entry(self, variable: str, period):
...
-        # intermediate values computed (to avoid impacting performance) but we mark them
-        # for deletion from the cache once the calculation ends.
-        count = 0
-        for frame in reversed(self.tracer.stack):
-            self.invalidate_cache_entry(str(frame["name"]), frame["period"])
-            if frame["name"] == variable:
-                count += 1
-                if count > self.max_spiral_loops:
-                    break
+        invalidate_entries = False
```
Morendil, line 419: I notice that this delta removes a comment. The comment noted that the stack was visited in reverse order; I'd assume that was a significant decision. I'm a bit worried to see this delta use a different order AND remove the comment.

"The tests still pass, so it must be OK" is an objection that I could understand, but perhaps in the context of OF-Core it's not as convincing as it would be in other contexts…

guillett: Thanks for your comment @Morendil. Indeed, I changed the logic because the `break` logic was not general enough.
In order to avoid silent regressions, I updated and improved the relevant tests in #1206.

guillett: I realized that the intention could be improved incrementally. I update the function code. In my opinion, it is a much clearer version.

bonjourmauko (changes requested):
Thanks a lot @guillett!

This looks like a very interesting proposal. If I'm getting it right this represents a major advance in spirals logic and therefore also a breaking change.

In order to accept this pull request, as a reviewer, it will help to have a better description of how this changes the current domain logic and, for impacted users, how can they migrate in case of experiencing errors after the upgrade. Also, given it is –it seems– a breaking change, I suggest you add the corresponding tests.

PR #1215, opened 2024-08-05
Bump twine version to allow deployment

benjello:
#### Technical changes

- Bump twine version to allow deployment.


benoit-cty (approved):
Merci !

PR #1220, opened 2024-09-16
[3/17] Fix doc & type definitions in the entities module

bonjourmauko:
Depends on #1219
Depended upon by #1146 

#### Technical changes

- Fix doc & type definitions in the entities module

PR #1146, opened 2022-08-01
Add weeks

maukoquiroga:
Supersedes #979 

# TODO

- [ ] Add baseline test coverage for periods
- [ ] Add baseline test coverage for holders
- [ ] Add `DateUnit` to replace string date units
- [ ] Implement ISO Week Calendar
- [ ] Implement week offsets
- [ ] Implement week input dispatchs
- [ ] Implement week input divides
- [ ] Split instant in two (ISO format and ISO calendar)
- [ ] Complete doc (code)
- [ ] Complete doc (mainstream)
- [ ] Simplify tests — or use a specific tool for brute-force testing
- [ ] Add typing
- [ ] Deprecate obsolete methods
- [ ] Implement output dispatch
- [ ] Implement output divide
- [ ] Adapt variables
- [ ] Adapt holders
- [ ] Adapt populations
- [ ] Adapt simulations
- [ ] ?

PR #1221, opened 2024-09-16
Fix doc & type definitions in the projectors module

bonjourmauko:
Depends on #1220
Depended upon by #1146 

#### Technical changes

- Fix doc & type definitions in the projectors module


openfisca_core/commons/types.py
benoit-cty, line 1: Why re-creating types here instead of importing them directly ?

bonjourmauko: I'm still unsure what is the best way to go.

openfisca_core/periods/helpers.py
```diff
@@ -125,7 +141,7 @@ def period(value) -> Period:
         Period((<DateUnit.DAY: 'day'>, Instant((2021, 1, 1)), 1))
 
         >>> period(DateUnit.ETERNITY)
-        Period((<DateUnit.ETERNITY: 'eternity'>, Instant((1, 1, 1)), inf))
+        Period((<DateUnit.ETERNITY: 'eternity'>, Instant((1, 1, 1)), 0))
```
benjello, line 144: Should be -1 here ?

openfisca_core/periods/helpers.py
```diff
@@ -164,15 +180,9 @@ def period(value) -> Period:
         return Period((DateUnit.DAY, instant(value), 1))
 
     # We return an "eternity-period", for example
-    # ``<Period(('eternity', <Instant(1, 1, 1)>, inf))>``.
+    # ``<Period(('eternity', <Instant(1, 1, 1)>, 0))>``.
```
benjello, line 183: Should be -1 here ?

PR #1223, opened 2024-09-17
[9/17] Typing & maintenance to periods

bonjourmauko:
Depends on #1220 
Depended upon by #1146 

#### Technical changes

- Update `pendulum'


openfisca_core/periods/_parsers.py
```diff
@@ -34,7 +34,7 @@ def _parse_period(value: str) -> Optional[Period]:
         return None
 
     # Check for a non-empty string.
-    if not (value and isinstance(value, str)):
+    if len(value) == 0:
```
benjello, line 37: Are we completely sure that value will be a string ?

openfisca_core/periods/helpers.py
```diff
@@ -162,15 +180,9 @@ def period(value) -> Period:
         return Period((DateUnit.DAY, instant(value), 1))
 
     # We return an "eternity-period", for example
-    # ``<Period(('eternity', <Instant(1, 1, 1)>, inf))>``.
+    # ``<Period(('eternity', <Instant(1, 1, 1)>, 0))>``.
```
benjello, line 183: Are we sure we do not break anything with this change ?

openfisca_core/periods/instant_.py
```diff
@@ -195,6 +209,10 @@ def offset(self, offset, unit):
             return self.__class__((date.year, date.month, date.day))
         return None
 
-    @property
-    def year(self):
-        return self[0]
+    @classmethod
+    def eternity(cls) -> t.Instant:
```
benjello, line 213: Why do we need this ?

benjello (changes requested):
There is an empty file `openfisca_core/periods/py.typed`

You should check these changes against `openfisca-france` and `openfisca-survey-manager` since it may surface some problems that should be documented in the changelog to ease transition and to diagnose the right version bump.

openfisca_core/periods/_parsers.py
bonjourmauko: No, it just tests for the declared types in the method signature. I willl make it broader, just in case.

openfisca_core/periods/helpers.py
bonjourmauko: It may happen, but it would be strange, as it will imply that you handle by default periods as floats and not as integers (integers are a subtype of floats). If that was the case, it would be a waste of memory, you really just need ints here. Should I mark it as a major bump and announce it?

openfisca_core/periods/instant_.py
bonjourmauko: Creating eternity instants was repeated through four times. I thought it made it easier to test. That makes it definitely a minor bump, thanks for seeing it. I'll update the PR.

openfisca_core/periods/helpers.py
benjello: I think so.

bonjourmauko: The constraint is to make `Instant` a pure `int` construct.  Ideally, to do the same with `Period`, mimicking what is already done with `indexed_enums` (that on another PR). Does it make sense?

openfisca_core/periods/instant_.py
bonjourmauko: Add it to the changelog.

openfisca_core/periods/helpers.py
benjello: OK

openfisca_core/periods/_parsers.py
bonjourmauko: Actually, yes. This is an internal function, so we're sure these are `str`.

bonjourmauko:
> There is an empty file `openfisca_core/periods/py.typed`

That is pure documentation. Is to let others know the code is typed. See: https://peps.python.org/pep-0561/

> You should check these changes against `openfisca-france` and `openfisca-survey-manager` since it may surface some problems that should be documented in the changelog to ease transition and to diagnose the right version bump.

OK.

bonjourmauko:
> There is an empty file `openfisca_core/periods/py.typed`
> 
> You should check these changes against `openfisca-france` and `openfisca-survey-manager` since it may surface some problems that should be documented in the changelog to ease transition and to diagnose the right version bump.

It all works @benjello :)

In France, the only check that fails is the Conda build, but that is because we're testing against a URL and not a PyPi release, which is normal, as this has yet to be merged and released.

The important is that the tests pass.

benjello (approved):

PR #1253, opened 2024-10-01
docs(entities): fix docs

bonjourmauko:
Fixes #1252 

#### Documentation

- Fix documentation of `entities`

PR #1255, opened 2024-10-01
test(entities): fix doctests

### LINKED ISSUE #1254: Fix entities doctests
Part of #1061

bonjourmauko:
Fixes #1254 
Depends on #1253 
Supersedes #1034  

#### New features

- Make `CoreEntity` public
  - Allows for more easily creating customised entities.

#### Technical changes

- Add missing doctests.

### DIFF
--- a/openfisca_core/entities/_core_entity.py
+++ b/openfisca_core/entities/_core_entity.py
@@ -9,19 +9,27 @@ from . import types as t
 from .role import Role
 
 
-class _CoreEntity:
+class CoreEntity:
     """Base class to build entities from.
 
     Args:
-        __key: A key to identify the ``_CoreEntity``.
-        __plural: The ``key`` pluralised.
-        __label: A summary description.
-        __doc: A full description.
-        *__args: Additional arguments.
+        *__args: Any arguments.
+        **__kwargs: Any keyword arguments.
+
+    Examples:
+        >>> from openfisca_core import entities
+        >>> from openfisca_core.entities import types as t
+
+        >>> class Entity(entities.CoreEntity):
+        ...     def __init__(self, key):
+        ...         self.key = t.EntityKey(key)
+
+        >>> Entity("individual")
+        Entity(individual)
 
     """
 
-    #: A key to identify the ``_CoreEntity``.
+    #: A key to identify the ``CoreEntity``.
     key: t.EntityKey
 
     #: The ``key`` pluralised.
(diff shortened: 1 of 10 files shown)


bonjourmauko:
Poke @benjello @benoit-cty

openfisca_core/entities/_core_entity.py
```diff
@@ -72,11 +73,46 @@ def get_variable(
...
+            >>> this.get_variable("tax")
+
+            >>> this.get_variable("tax", check_existence=True)
+            Traceback (most recent call last):
+            VariableNotFoundError: You tried to calculate or to set a value...
+
+            >>> class tax(variables.Variable):
+            ...     definition_period = periods.MONTH
+            ...     value_type = float
+            ...     entity = that
```
benjello, line 107: Why using different entities, `this` and `that`  here ?

openfisca_core/entities/_core_entity.py
```diff
@@ -90,16 +126,47 @@ def check_variable_defined_for_entity(self, variable_name: t.VariableName) -> No
...
+            >>> this.set_tax_benefit_system(tax_benefit_system)
+
+            >>> this.check_variable_defined_for_entity("tax")
+            Traceback (most recent call last):
+            VariableNotFoundError: You tried to calculate or to set a value...
+
+            >>> class tax(variables.Variable):
+            ...     definition_period = periods.WEEK
+            ...     value_type = int
+            ...     entity = that
```
benjello, line 153: Same as before.
Why using different entities, `this` and `that`  here ?

openfisca_core/entities/_core_entity.py
bonjourmauko: To show that this method is going to get you the variable regardless of who is this variable being defined for. It was a confusing behaviour for me (a couple of years ago I implemented an actual filter, to search within the variables defined for the entity, but I'll repropose that one later as it is a breaking change).

openfisca_core/entities/_core_entity.py
bonjourmauko: Same answer as above.

openfisca_core/entities/_core_entity.py
bonjourmauko: (The test in itself has exactly that purpose, for people to think "Wait, this is weird behaviour". If tomorrow someone changes this, the test will fail, and that is what I actually expect).

benjello (approved):

PR #1173, opened 2022-12-13
Synchronize Conda and Pypi deps

benoit-cty:
#### Technical changes

- Fix conda build in CI.
- Avoid duplicate run in CI.
- Build and test conda package in PR, no more on master.
- Deploy only on master.

PR #1257, opened 2024-10-02
ci: add matrix & fix conda

bonjourmauko:
Fixes #1173
Relates to openfisca/country-template#124
Relates to openfisca/extension-template#43

#### Technical changes

- Fix conda test and publish
- Add matrix testing to CI
  - Now it tests lower and upper bounds of python and numpy versions


benoit-cty (approved):
Thanks, it's better now !

bonjourmauko:
> Thanks, it's better now !

We still have some issues to tacke in other PRs:

- Should we use a common lock file, like with `poetry`, to make builds deterministic in PyPi and Conda
- The two-staged build process (core -> templates -> core-dev)
- The publishing of templates

PR #1283, opened 2024-10-18
build: remove coveralls

### LINKED ISSUE #1073: Remove unused Coveralls
Hi there!

### Here is what I did:

- Submit a PR from a third-party organisation.

- CI passes.

- CI fails on sending data to coveralls.io.

https://github.com/openfisca/openfisca-core/pull/1070#issuecomment-945785704

## Context

I identify more as a:

- Maintainer.

- - - - 

Coverage data storage was introduced by @maukoquiroga in b785382722eeb72b8cc72f18cb0942e0b6424c1e. It does not seem to be used by anyone in the build / release process, and a quick question on Slack did not yield any other reply. I am in favour of removing coveralls.io altogether, which I suggest to do unless someone chimes in within 4 days to say they use it, with a fast removal if @maukoquiroga confirms it can be removed.

The alternative would be to not execute the Coveralls CI job on third-party PRs.

benjello:
#### Technical changes

- Remove `covearalls`

Fixes #1073


benjello:
@bonjourmauko : got a doctest failure

bonjourmauko:
> @bonjourmauko : got a doctest failure

Weird that it didn't fail before. These are expected, however very useless, errors.

bonjourmauko:
> > @bonjourmauko : got a doctest failure
> 
> Weird that it didn't fail before. These are expected, however very useless, errors.

Actually not that useless.

For example, if you want to build a `recarray` containing `str`, you can define by hand the `itemsize`. Thanks to the testing matrix, we will now that in windows we can have an offset error of 10 bits, and that we should better pad automatically and increase the offset. These are very weird situations, so it is actually good to be able to catch them.

bonjourmauko (approved):

bonjourmauko:
@benjello it is your PR so I leave the merge to you.

PR #1298, opened 2024-11-19
Test to evaluate the evolution of performance

benoit-cty:
#### Technical changes

- Add a test on 50 000 persons to evaluate the evolution of performance


bonjourmauko:
By the look of it, maybe you would like to take a look at https://pytest-with-eric.com/pytest-advanced/hypothesis-testing-python/

bonjourmauko:
According to this test, there is no substantial difference between both Enum `__eq__` implementations.

Screenshot 2024-11-20 at 02 08 12

PR #1304, opened 2024-11-20
revert: restore line printing in tracer

### LINKED ISSUE #1310: Implement/restore line printing in tracer
Hi there!

### Here is what I did:

`openfisca test -v`

To print the computation log.

It did not.

In `computation_log`, the method for printing the log is noop.

via @guillett

guillett:
Fixes #1310
Depends on #1321

#### Technical changes

- Fix `computation_log`.

### DIFF
--- a/openfisca_core/tracers/computation_log.py
+++ b/openfisca_core/tracers/computation_log.py
@@ -46,8 +46,8 @@ class ComputationLog:
         If ``max_depth`` is set, for example to ``3``, only print computed
         vectors up to a depth of ``max_depth``.
         """
-        for _ in self.lines(aggregate, max_depth):
-            pass
+        for line in self.lines(aggregate, max_depth):
+            print(line)
 
     def _get_node_log(
         self,
--- a/openfisca_core/tracers/computation_log.py
+++ b/openfisca_core/tracers/computation_log.py
@@ -47,7 +47,7 @@ class ComputationLog:
         vectors up to a depth of ``max_depth``.
         """
         for line in self.lines(aggregate, max_depth):
-            print(line)
+            print(line)  # noqa: T201
 
     def _get_node_log(
         self,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+### 43.2.9 [#1304](https://github.com/openfisca/openfisca-core/pull/1304)
+
+#### Bugfix
+
+- Fix `computation_log`.
+
 ### 43.2.8 [#1322](https://github.com/openfisca/openfisca-core/pull/1322)
 
 #### Technical changes
--- a/setup.py
+++ b/setup.py
@@ -70,7 +70,7 @@ dev_requirements = [
 
 setup(
     name="OpenFisca-Core",
-    version="43.2.8",
+    version="43.2.9",
     author="OpenFisca Team",
     author_email="contact@openfisca.org",
     classifiers=[


bonjourmauko:
Hi! Was this a bug introduced recently? I'd love to add a failing test for this. If you could point me out at how to reproduce the error and the expected behaviour, I could help with the test.

bonjourmauko:
Ok I understood now: printing/logging has nothing to do in the computation log, which is just a log accumulator —like a writer monad transformer, sort of.

In `test_runner.py`, you could either:

1. Get the log buffer and log it to stdout there.
2. Pass down a function to the tracer with how you want to print/log the log buffer.

Something like this should give you what you want:

```py
# task_runner.py
@staticmethod
def print_computation_log(tracer, aggregate, max_depth) -> None:
    buffer = tracer.computation_log.lines(aggregate, max_depth)
    for line in buffer:
        print(line)
```

Also, using `print` is making the linter fail in CI, you have to use a logger for that.

Hope it helps :)

bonjourmauko (changes requested):
Ok I understood now: printing/logging has nothing to do in the computation log, which is just a log accumulator —like a writer monad transformer, sort of.

In `test_runner.py`, you could either:

1. Get the log buffer and log it to stdout there.
2. Pass down a function to the tracer with how you want to print/log the log buffer.

Something like this should give you what you want (just an idea):

```py
# task_runner.py
@staticmethod
def print_computation_log(tracer, aggregate, max_depth) -> None:
    buffer = tracer.computation_log.lines(aggregate, max_depth)
    for line in buffer:
        print(line)
```

Also, using `print` is making the linter fail in CI, you have to use a logger for that.

Hope it helps :)

guillett:
Thanks @bonjourmauko , I don't have much time to follow up. To me, this is a regression and I need this fix to work locally.

bonjourmauko:
@guillett If you give me the command to use I can propose you a stacked PR with the proposed fix.

Concerning `print`, even if I accept it, you won't be able to merge because of the linter.

bonjourmauko:
Oh I see the regression was introduced here https://github.com/openfisca/openfisca-core/commit/2f8072fdf270bfd204f1103b227a893f80a3bdaa#diff-fc2477fbba8037857fb9287895920ee5a1015c5eda4edb50832af14232fdd10b

openfisca_core/tracers/computation_log.py
```diff
@@ -46,8 +46,8 @@ def print_log(self, aggregate: bool = False, max_depth: int = sys.maxsize) -> No
         If ``max_depth`` is set, for example to ``3``, only print computed
         vectors up to a depth of ``max_depth``.
         """
-        for _ in self.lines(aggregate, max_depth):
-            pass
+        for line in self.lines(aggregate, max_depth):
+            print(line)
```
bonjourmauko, line 50:
```suggestion
            print(line)  # noqa: T201
```

benjello (approved):

benjello:
@guillett : I modified the changlog so the linter passes but we still have a problem with a conda build.
@benoit-cty: didn't we got this error already ? Any hint ?

benoit-cty:
> @guillett : I modified the changlog so the linter passes but we still have a problem with a conda build. @benoit-cty: didn't we got this error already ? Any hint ?

Well, the error in Conda complains about Numpy : `If using `np.array(obj, copy=False)` replace it with `np.asarray(obj)` to allow a copy when needed (no behavior change in NumPy 1.x).`
Because despite the `np1.6.24` in the name of the test, it is a Numpy 2 version that has been installed:

`  + numpy                               2.2.3  py310hefbff90_0       conda-forge                   8MB `

bonjourmauko:
There is definitely a problem with the conda's config.

bonjourmauko:
poke @sylvainipp it looks like the problem you were having could come from inadvertently using a numpy > 2.2 version, that could be unrelated to your proposed changes in #1319

PR #1319, opened 2025-02-07
fix: deepcopy variables in clone of tax_benefit_system

sylvainipp:
Depends on #1321

#### Technical changes

- Change the copy of the variables in tax_benefit_system.clone() to a deepcopy

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -70,7 +70,7 @@ dev_requirements = [
 
 setup(
     name="OpenFisca-Core",
-    version="43.2.9",
+    version="43.3.0",
     author="OpenFisca Team",
     author_email="contact@openfisca.org",
     classifiers=[
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 43.3.0 [#1319](https://github.com/openfisca/openfisca-core/pull/1319)
+
+#### Technical change
+
+- Use a deepcopy of variables in tax and benefit system cloning instead of a copy.
+
 ### 43.2.9 [#1304](https://github.com/openfisca/openfisca-core/pull/1304)
 
 #### Bugfix
--- a/openfisca_core/taxbenefitsystems/tax_benefit_system.py
+++ b/openfisca_core/taxbenefitsystems/tax_benefit_system.py
@@ -576,7 +576,7 @@ class TaxBenefitSystem:
 
         new_dict["parameters"] = self.parameters.clone()
         new_dict["_parameters_at_instant_cache"] = {}
-        new_dict["variables"] = self.variables.copy()
+        new_dict["variables"] = copy.deepcopy(self.variables)
         new_dict["open_api_config"] = self.open_api_config.copy()
         return new
 
--- a/tests/core/test_tracers.py
+++ b/tests/core/test_tracers.py
@@ -439,7 +439,7 @@ def test_log_aggregate(tracer) -> None:
     tracer._exit_calculation()
     lines = tracer.computation_log.lines(aggregate=True)
 
-    assert lines[0] == "  A<2017> >> {'avg': 1.0, 'max': 1, 'min': 1}"
+    assert lines[0] == "  A<2017> >> {'avg': 1.0, 'max': 1, 'min': 1}" or lines[0] == "  A<2017> >> {'avg': np.float64(1.0), 'max': np.int64(1), 'min': np.int64(1)}"
 
 
 def test_log_aggregate_with_enum(tracer) -> None:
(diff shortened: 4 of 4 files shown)


sylvainipp:
I think this is mainly used in survey-manager, @clallemand do you think it is appropriate ? I need it when creating a reform based on inflate_parameters (an openfisca-survey-manager function), otherwise I have a problem with variables losing their end_date.

benjello:
A simple copy creates a new object but references the same memory addresses for nested objects. Changes to nested objects in the copy will affect the original.

Maybe you changed the variable on the original (or the copy) and got side effects.

But I agree, I cannot see situation where we want attributes changes transmitted between the copy and the original of a tbs.

clallemand:
I have no objection !

sylvainipp:
Thanks for your feedback ! The idea is that it is still possible to make the change in variables before cloning, and after it is possible to change only one of them. I don't think this transmission of changes is intentionally used by anyone, so I think it can be a minor version change even if it is technically breaking.

setup.py
```diff
@@ -70,7 +70,7 @@
 
 setup(
     name="OpenFisca-Core",
-    version="43.2.7",
+    version="43.2.8",
```
benjello, line 73: It is a minor change not a bugfix.

sylvainipp:
Hi @bonjourmauko, there are some errors in the tests due to default value of arrays, do you think it is linked to your recent changes ? I can't figure why it appears here and not in the previous PR (maybe new dependencies ?), but as you worked recently on it I wondered if you had an easy fix.

bonjourmauko:
Hi @sylvainipp, my first impression is that there is a casting problem only afecting conda. As testing in conda was recently introduced, it may well be the case that the error is pre-existing.

openfisca_core/indexed_enums/_utils.py
```diff
@@ -117,7 +117,7 @@ def _int_to_index(
 
     """
     indices = enum_class.indices
-    values = numpy.array(value, copy=False)
+    values = numpy.asarray(value)
```
bonjourmauko, line 120: Why do you need this here?

openfisca_core/indexed_enums/_utils.py
```diff
@@ -171,7 +171,7 @@ def _str_to_index(
         array([1, 1], dtype=uint8)
 
     """
-    values = numpy.array(value, copy=False)
+    values = numpy.asarray(value)
```
bonjourmauko, line 174: Why do you need this here?

bonjourmauko:
The changes look ok. The `copy=false` is to reuse the `memoryview` when possible, to save on ram. The best way to test if copying has a noticeable impact on performance, is to run a simulation with a huge dataset with both.

(Be aware that there is no right answer, and it depends a lot on the specs of the machine runing it.)

For the test, I can see that it is wrong, because we should not mix DTypes within an array. This part looks like a hack:

```python
choicelist = [array('first_parent', dtype='<U12'), array('second_parent', dtype='<U13'), array('child', dtype='<U5'), 0]
condlist   = [array([ True, False]), array([False,  True]), array([False, False])]
```

Have you tried replacing `0` with an `str` default value? Or with an array or either `intxx` or `str`?

openfisca_core/indexed_enums/_utils.py
sylvainipp: It has no link with the original modification, it was to fix one of the errors in the conda test (here : https://github.com/openfisca/openfisca-core/actions/runs/13202963909/job/36859430121)

openfisca_core/indexed_enums/_utils.py
sylvainipp: Idem

bonjourmauko:
tests/core/parameters_fancy_indexing/test_fancy_indexing.py:177
tests/core/test_dump_restore.py:19

bonjourmauko:
Looking at the message given by NumPy, it looks like we're on the right track about the cause:

> TypeError: Choicelist and default value do not have a common dtype: The DType <class 'numpy.dtypes._PyLongDType'> could not be promoted by <class 'numpy.dtypes.StrDType'>. This means that no common DType exists for the given inputs. For example they cannot be stored in a single array unless the dtype is `object`. The full list of DTypes is: (<class 'numpy.dtypes.StrDType'>, <class 'numpy.dtypes.StrDType'>, <class 'numpy.dtypes.StrDType'>, <class 'numpy.dtypes._PyLongDType'>)

sylvainipp:
Thanks ! I don't understand how it is possible but I see other lines at those place : 
tests/core/parameters_fancy_indexing/test_fancy_indexing.py:177 : `assert_near(P.single.owner[zone], [100, 200, 200, 100])`
tests/core/test_dump_restore.py:19 `simulation_dumper.dump_simulation(simulation, directory)`
I'm sorry if I miss something obvious but I search this `0` that is not of the good dtype and I don't manage to find it.

bonjourmauko:
Tu as le stack trace ici https://github.com/openfisca/openfisca-core/actions/runs/13264258879/job/37027895878?pr=1319

Ce `0` vient de NumPy. À vue rapide, il semblerait que le de/encodage est bypassé au moment du dump/restore.

bonjourmauko (approved):

PR #1332, opened 2025-07-11
Fix `return` instead of `raise`.

### LINKED ISSUE #1331: Error not raised when building an erroneous reform
I ran thiis test

```python 
from openfisca_core.reforms import Reform

class CustomReform(Reform):
    def apply(self):
        self.modify_parameters(modifier_function=self.modify_my_parameters)

    @staticmethod
    def modify_my_parameters(parameters):
        parameters.benefits.basic_income.update(start="2015-01-01", value=60)
        # return parameters  # Work when this line is not commented


tbs = CountryTaxBenefitSystem()

reform = CustomReform

reformed_tbs = reform(tbs)

assert reformed_tbs.parameters.benefits.basic_income("2015-01-01") == 60, "Reform did not apply correctly"
assert reformed_tbs.parameters.benefits.basic_income("2016") == 60, "Reform did not apply correctly"
```

And the reform didn't apply correctly since `modify_my_parameters` didn't return a `ParameterNode`.
I thought I was warned because this is supposed to be tested [here](https://github.com/openfisca/openfisca-core/blame/8861ea0704b229779f341c432a66eb9a9dc0dd0b/openfisca_core/reforms/reform.py#L81)

I don't understand why I didn't get this error.

cc @benoit-cty @eraviart @clallemand

benoit-cty:
#### Technical changes
- the modify_my_parameters function should return the ParameterNode, but you're not getting the expected error.  In the modify_parameters method on line 81, there's a bug in the code. It should raise a ValueError, but instead it returns a ValueError. This means the error is created but not actually thrown, so your code continues executing without the expected error.

- Will close #1331

### DIFF
--- a/.github/workflows/merge.yaml
+++ b/.github/workflows/merge.yaml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        os: [ubuntu-22.04, windows-2019]
+        os: [ubuntu-22.04, windows-2025]
         numpy: [1.26.4]
         # Patch version must be specified to avoid any cache confusion, since
         # the cache key depends on the full Python version. If left unspecified,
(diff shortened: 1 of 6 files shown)


benoit-cty:
> @benoit-cty that's a lot of added tests to review, and some checks fail. Could you please clarify if you wrote these tests yourself, or if they were generated by an LLM? If so, which review mechanism did you put in place to ensure that all these tests are relevant and warrant another human to review them? Thank you!

These tests was indeed added by Copilot to improve the test coverage on the area where @benjello discovered an error.

PR #1334, opened 2025-09-21
Allow string parameters

guillett:
#### New features

- Allow parameters to be string values

#### Technical changes

- Simplify fancing index logic (to allow string value subnodes)


MattiSG:
Wow, this is quite a big new feature! I am a bit concerned with the potential for abuse it has 😄 I fear that it might not be obvious to contributors how much of a performance hit there is using strings compared to enums, for example.
Would you be able to clarify what the use case for string parameters is? 🙂

guillett:
We have a multiple period study to realize spread on 24 months and we have to track grade changes over time and we have a grade to next grade map.

guillett:
If performance are really bad as we experiment, we'll find ways to rely on Enums. If it great very convenient and probably safe to keep on relying on str.

guillett:
We plan on dynamically creating the mappings from CSV files (and Grist tables actually).

PR #1342, opened 2025-11-11
Add Contributor Covenant Code of Conduct

ricardomiron:
Added Contributor Covenant 3.0 Code of Conduct to promote a safe and inclusive community.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,90 @@
+
+# Contributor Covenant 3.0 Code of Conduct
+
+## Our Pledge
+
+We pledge to make our community welcoming, safe, and equitable for all.
+
+We are committed to fostering an environment that respects and promotes the dignity, rights, and contributions of all individuals, regardless of characteristics including race, ethnicity, caste, color, age, physical characteristics, neurodiversity, disability, sex or gender, gender identity or expression, sexual orientation, language, philosophy or religion, national or social origin, socio-economic position, level of education, or other status. The same privileges of participation are extended to everyone who participates in good faith and in accordance with this Covenant.
+
+## Encouraged Behaviors
+
+While acknowledging differences in social norms, we all strive to meet our community's expectations for positive behavior. We also understand that our words and actions may be interpreted differently than we intend based on culture, background, or native language.
+
+With these considerations in mind, we agree to behave mindfully toward each other and act in ways that center our shared values, including:
+
+1. Respecting the **purpose of our community**, our activities, and our ways of gathering.
+2. Engaging **kindly and honestly** with others.
+3. Respecting **different viewpoints** and experiences.
+4. **Taking responsibility** for our actions and contributions.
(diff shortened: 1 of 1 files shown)


MattiSG:
The relevance of adding a Code of Conduct is reinforced by the ramp-up of LLM-driven contributions. See for example https://github.com/openfisca/openfisca-core/issues/1380#issuecomment-5045402771, where a denial of handling generated the following reply, likely from an LLM-assisted human (emphasis mine):

> […]
> I acknowledge the substantive feedback. I also hope OpenFisca will reflect on how it was expressed. The changing scope, deadline, submission volume, and use of tooling could all have been criticised directly. Indeed, the top maintainer of Linux is quoted just this week in a different but related context. Inferring that the work was commercially motivated or that the account was an LLM, without first asking for clarification, introduced assumptions that weren't needed to make those points.
> I think you probably do need to be aware that in many of the countries that OpenFisca is based, your passing comment about commercial interests could be considered an allegation of financial corruption. Internationally borders can become irrelevant in these contexts (e.g. the US). Unfortunately (and related to my work), New Zealand has some of the most lax public integrity standards internationally, but **fortunately for you that means I'm not too fussed about the misfire. However, I think it's important to be aware that misattribution of commercial motivation to other people who have made an honest mistake, might result in a considerably different response- FOSS or not**. Ironically, LLMs could be quite good for checking future comments before posting.

The veiled threat to maintainers is toxic to the community, as is the pace of “contribution” due to LLMs and the automated notifications sent by LLM agents, both in #1382 and in https://github.com/BetterRules/openfisca-aotearoa/pull/200#issuecomment-4943749806.

I wonder how much the [Contributor Covenant 3.0](https://www.contributor-covenant.org/version/3/0/code_of_conduct/) is enough to address these issues. I was part of the feedback process before it was finalised, and realise that it operates under the assumption that parties involved are human. For example:

> 2. Engaging kindly and honestly with others.
> 5. Gracefully giving and accepting constructive feedback.

is from my perspective unwarranted and just too much to ask from human maintainers towards LLM agents who have no feelings to be hurt but are able to spawn massive amounts of contributions. I don't believe we have a moral obligation to extend kindness and grace to LLMs, and the cost of doing so is too high for open source work.

Equally important, how can we expect LLM agents to:

> 4. Taking responsibility for our actions and contributions.

What does “taking responsibility” look like for an LLM?

I also wonder how the chilling effect towards maintainers in the quoted comment above would be addressed by the Contributor Covenant. Would it go into “7. Behaving in other ways that promote and sustain the well-being of our community.” and leave it to maintainer enforcement?

- - -

At the very least, I would recommend editing the CoC to start with:

```diff
- We pledge to make our community welcoming, safe, and equitable for all.
+ We pledge to make our community welcoming, safe, and equitable for all humans.
```

…and only extend CoC protections to humans. This would be enforceable with the parallel adoption of mandatory tooling disclosure in #1372. This would however be very easily cheated by LLMs and LLM-assisted humans. The question of proof of humanity to benefit from Covenant protections thus remains.

PR #1330, opened 2025-07-10
Update changelog and version number

benjello:
#### Technical changes

- Add a check for creating a simulation when building a test

PR #1350, opened 2025-12-18
Build test

MattiSG:
This pull request restores #1330 whose contents were destroyed following a [Shai Hulud 2](https://about.gitlab.com/blog/gitlab-discovers-widespread-npm-supply-chain-attack/) attack.

As the original contents were force-pushed after PR closure, we are not able to restore the original PR. See the original PR for discussion and context.


benjello:
ping @guillett @benoit-cty

benoit-cty (approved):

PR #1352, opened 2025-12-29
Change license

benoit-cty:
#### Technical changes

The PyPi release break for a License syntax problem:

```sh
ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/        
         'License :: AGPL-3.0-only' is not a valid classifier. See              
         https://packaging.python.org/specifications/core-metadata for more     
         information.                                                           
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 #### Technical changes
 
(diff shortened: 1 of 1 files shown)


benoit-cty:
Tested with:

```sh
make install-dist
make build
twine upload --verbose  --repository testpypi dist/*
```

PR #1354, opened 2025-12-30
Fix/type error choicelist

benoit-cty:
#### Technical changes

Numpy 2.x check for typing in numpy.select. This behavior break OpenFisca-Core, but as some unit test was not run in CI, it has not been discovered by the upgrade.

The default value for `default` is 0, which is not compatible with a string for Enum.

So this PR:
- Add unit test in CI
- Specify the default value in a compatible type.

Here is the error:

` TypeError: Choicelist and default value do not have a common dtype: The DType <class 'numpy.dtypes._PyLongDType'> could not be promoted by <class 'numpy.dtypes.StrDType'>. This means that no common DType exists for the given inputs. For example they cannot be stored in a single array unless the dtype is `object`. The full list of DTypes is: (<class 'numpy.dtypes.StrDType'>, <class 'numpy.dtypes.StrDType'>, <class 'numpy.dtypes._PyLongDType'>)
`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+### 44.0.4 [#1354](https://github.com/openfisca/openfisca-core/pull/1354)
+
+#### Technical changes
+
+- Add missing unit test in CI
+- Specify the default value in a compatible type in numpy.select for Numpy 2.x.
+
 ### 44.0.3 [#1352](https://github.com/openfisca/openfisca-core/pull/1352)
 
 #### Technical changes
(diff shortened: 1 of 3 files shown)


benoit-cty:
Also tested with France:

```sh
# Create fresh venv with Python 3.13.11
python3.13 -m venv /tmp/test-ci-max
source /tmp/test-ci-max/bin/activate

# Build like CI
pip install --upgrade pip build twine
python -m build
pip install dist/*.whl
pip install tomli 'openfisca-core[web-api]'

# Check versions - should be maximal (latest)
pip show openfisca_core
pip show numpy

# Run tests
shopt -s globstar
openfisca test --country-package openfisca_france tests/**/*.py
```
=> Failed

After `pip install -e ../openfisca-core`, it works.

PR #1358, opened 2026-01-15
Add --ignore-default option to openfisca test command

benjello:
#### New features

- Add `--ignore-default` option to `openfisca test` command
  - When used with `--verbose`, filters out variables that have their default value from the computation log output


benoit-cty (approved):
Nice !
Maybe add something in the README or CONTRIBUTING.md ?

Issue #1359, opened 2026-01-16
Suggestion de Claude que l'on gagnerait à regarder


benjello:
Suggestion de Claude :
```
  Issue found in in_yaml_test_reform.py:15:
  reformed_parameters=dict,  # Should be: reformed_parameters: dict
  This is a type annotation bug - =dict sets the default to the dict class itself, not an empty dict. It should be : dict for type hint or = None for a default value.

  Also in apply() method (line 26-27): The modify_parameters function ignores local_parameters and completely replaces the parameter tree. This seems intentional but the docstring says "similar to a parameter tree" - might be worth clarifying if this fully replaces or merges parameters.

```

_Originally posted by @benoit-cty in https://github.com/openfisca/openfisca-core/issues/1358#issuecomment-3761570816_

Issue #1372, opened 2026-03-13
RFC: AI contribution policy


benoit-cty:
AI contribution is ramping up. To prevent unwanted behavior by the community I open a discussion on the AI policy we want for OpenFisca.

Not written by AI. Greatly inspired by https://github.com/kornia/kornia/blob/main/AI_POLICY.md

# 🤖 OpenFisca AI & Authorship Policy

**Version:** 1.0
**Enforcement:** Strict
**Applicability:** All Pull Requests (Human & Bot)

## 1. Core Philosophy

OpenFisca accepts AI-assisted code (e.g., using Copilot, Cursor, etc.), but strictly rejects AI-generated contributions where the submitter acts merely as a proxy. The submitter is the **Sole Responsible Author** for every line of code, comment, and design decision.

### Law 1: Proof of Verification

AI tools frequently write code that looks correct but fails execution. Therefore, "vibe checks" are insufficient.

**Requirement:** Every PR introducing functional changes must be carefully tested locally by the human contributor before submition. This is mandatory for all contributors and is particularly important for first-time contributors.

### Law 2: The Hallucination & Redundancy Ban

AI models often hallucinate comments or reinvent existing utilities.

**Requirement:** You must use existing method and library and never reinvent the wheel.

**Failure Condition:** Creating new helper functions when an OpenFisca equivalent exists is grounds for immediate rejection.

**Failure Condition:** "Ghost Comments" (comments explaining logic that was deleted or doesn't exist) will result in a request for a full manual rewrite. Unnecessary comments are not allowed. Example: "This function returns the input".

### Law 3: The "Explain It" Standard

**Requirement:** If a maintainer or reviewer asks during code review, you must be able to explain the logic of any function you submit.

**Failure Condition:** Answering a review question with "That's what the AI outputted" or "I don't know, it works" leads to immediate closure.

### Law 4: Transparency in AI Usage Disclosure

**Requirement:** All PRs must accurately complete the "AI Usage Disclosure" section in the pull request template. This disclosure is mandatory and must reflect the actual use of AI tools.

**When to mark as 🟥 AI-vibecoded:**
- An AI agent (e.g., Cursor, GitHub Copilot, ChatGPT, etc.) generated the code, PR description, or commit messages
- You cannot explain the logic without referring to the AI's output
- The PR was created primarily by an agent with minimal human review or modification
- Car analogy : the car drive by itself, your are outside it and just tell it where to go.

**When to mark as 🟠 AI-generated:**
[...]

PR #1373, opened 2026-03-19
Fix InYamlTestReform typing and clarify YAML parameter replacement

benjello:
## Summary
This PR fixes the YAML test reform typing issue and clarifies behavior, then updates project metadata accordingly.

### What changed
- Fix  constructor signature:
  - replace  with a typed optional dict and safe default initialization
- Clarify docstring behavior:
  - YAML test  replaces the full parameter tree
- Bump package version in :
  -  ->  (patch bump, non-breaking bugfix)
- Update  with a  entry referencing #1359
- Add/extend repository  guidance so future agents consistently:
  - use  for Python commands
  - open a PR after commits
  - apply version bump + changelog updates for functional changes

## Why
 sets a default value to the  class object instead of expressing a type hint. This is ambiguous and error-prone. The new signature is explicit and safer.

## Validation
- 
- Result: 

## Compatibility
- No breaking API change expected.
- This is a patch-level release change.


MattiSG (changes requested):
Do NOT merge any PR until the codebase revert that has been [described in our community space](https://openfisca.slack.com/archives/C3NFSJMPG/p1772763473615479) or equivalent has been implemented (copied below for reference), otherwise this will bring back the behaviour that had to be yanked:

> I believe we should revert the codebase to 8d26af3c424ff635d9cb57574aef67e37ca0e912, i.e. align it with the `44.2.2`  version that is the one currently published on Pypi, with:
> 1. A revert commit.
> 2. An update of the `CHANGELOG` to mention yanked versions
> 3. A version bump to `44.6.0` to restore continuous deployment capabilities.

AGENTS.md
MattiSG, line 1: I don't think we should add an `AGENTS` file until we have clarity on an AI contribution policy, cf. #1372.

benjello: Removed

PR #1383, opened 2026-08-17
Fix import failure on NumPy 2.5

### LINKED ISSUE #1379: Test broken with numpy >=2.5 (metaclass conflict)
Hi ! In Openfisca-France, the yaml tests broke with numpy ==2.5 with a metaclass conflict (https://github.com/openfisca/openfisca-france/actions/runs/28084812855/job/84046072485?pr=2765). The same tests have no error with numpy==2.4, even if it might be due to numpy dependencies and not directly numpy.
In the short run, the numpy version will be restricted in OpenFisca-France in order to keep the tests working, but it might also be a problem in other constituencies and it will restrict future developments if not tackled.

HugoFara:
`import openfisca_core` fails on NumPy >= 2.5 with `TypeError: metaclass conflict` at `types.py:182`. It is an import-time crash, so `openfisca test` and every other entry point goes down with it. 

NumPy 2.5 made `numpy.typing.NDArray` a PEP 695 `typing.TypeAliasType`, which cannot be a runtime base class. `types.EnumArray` now derives from `numpy.ndarray` directly, keeping its dtype parameters. 

**It also adds one NumPy 2.5.2 entry to `push.yaml`**, to `setup-pip` and `test-pip` both because `test-pip` reuses a cache key containing the version. The matrices stopped at 2.3.0, which is why nothing caught this. There is no unit test so that entry is the regression guard. Happy to split it out.

Under 2.5.2 the suite goes from crashing on import to 473 passed, matching the 2.4.6 baseline, and the change is a no-op on 2.4.6. `check-style` and `lint-doc` pass. `check-types` fails under 2.5 for an unrelated reason. `python_version = 3.10` in `setup.cfg` cannot parse NumPy's PEP 695 stubs, so the lint matrix stays at 2.3.0 and I would file that separately.

Closes #1379.

Disclosure, in the factual form discussed in #1372: Claude Opus 5 (Claude Code) located the failure, bisected the behaviour between NumPy 2.4.6 and 2.5.2, ran the suites and drafted this description; I chose the fix and the scope, reviewed the diff, and reproduced the tests and both lint targets locally.

---
*Pull request draft by Claude Opus 5, corrected manually.*

### DIFF
--- a/.conda/openfisca-core/meta.yaml
+++ b/.conda/openfisca-core/meta.yaml
@@ -49,10 +49,8 @@ test:
 
 outputs:
   - name: openfisca-core
-    type: conda_v2
 
   - name: openfisca-core-api
-    type: conda_v2
     build:
       noarch: python
     requirements:
@@ -68,7 +66,6 @@ outputs:
         - {{ pin_subpackage('openfisca-core', exact=True) }}
 
   - name: openfisca-core-dev
-    type: conda_v2
     build:
       noarch: python
     requirements:
--- a/.github/workflows/_before-conda.yaml
+++ b/.github/workflows/_before-conda.yaml
@@ -65,10 +65,6 @@ jobs:
         use-mamba: true
       if: steps.cache-env.outputs.cache-hit != 'true'
 
-    - name: Install dependencies
-      run: mamba install boa rattler-build
-      if: steps.cache-env.outputs.cache-hit != 'true'
-
     - name: Update conda & dependencies
       uses: conda-incubator/setup-miniconda@v3
       with:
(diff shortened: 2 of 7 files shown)


benoit-cty (approved):
Thanks for the fix !
- Python 3.13 + numpy 2.5.2: import openfisca_core succeeds on the branch; core suite 977 passed / 3 skipped, country-template 35 passed. The 7 openfisca_core/populations doctest failures are pre-existing on master and unrelated.
- The new base class is runtime-equivalent: EnumArray.__mro__ is (EnumArray, numpy.ndarray, object) both before and after, since NDArray[X] and ndarray[tuple[int, ...], dtype[X]] resolve to the same MRO entry. ndarray.__class_getitem__ exists since numpy 1.22, so the oldest matrix entry (1.24.2) is safe.
make check-types on master vs. branch: byte-identical error lists (50 both sides, all pre-existing). Tightening the shape param from Any to tuple[int, ...] adds no new type errors.
- CI plumbing checks out: 2.5.2 is in both setup-pip and test-pip with matching cache keys, install-test uses --no-deps so the pinned version is really what's exercised, setup.py's numpy >=2.1.0, <=3 for py3.13 permits it, and the 44.7.0 → 44.7.1 bump + CHANGELOG entry satisfy is-version-number-acceptable.sh.

MattiSG:
🎉 thanks @HugoFara and @benoit-cty! Looking forward to seeing this land in downstream country models 👍

Issue #956, opened 2020-08-17
Spiraling


magemax:
Hi there!

I really enjoy OpenFisca, but I recently encountered an issue.

### Here is what I did:

Created a recursive formula to model some mechanism that is defined the following way :
    

    - name: Montant total 2019 et 2020
      output:
        dsr_montant_total_fraction_cible: 
          2019: 323_780_451
          2020: 360_336_634
    
    - name: Montant total 2020
      output:
        dsr_montant_total_fraction_cible: 
          2020: 360_336_634

## Context

I identify more as a:

- Analyst _(I make macroscopic computations on real populations)_.
- Business expert _(I create tests and model legislation)_.
- Commenter _(I make data visualisations)_.
- Developer _(I create tools that use the existing OpenFisca code)_.
- Historian _(I accumulate data based on past legislation values)_.
- Lobbyist _(I model reforms to make them exist)_.
- Producer _(I make computations on individual situations)_.

(remove this line as well as all items in the list that don't fit your situation)

Morendil:
Thanks for the heads-up @magemax - taking a quick 👀  before getting back to you with an assessment.

Morendil:
**Suggested workaround**: remove the "recursion", replace it instead with an explicit loop and check on the requested period, i.e. something like `if year <= 2019 then base_value else base_value * (increase_rate ^ (year-2019))`.

The exception (SpiralError) is raised here:
 https://github.com/openfisca/openfisca-core/blob/2fd3a2ce40df17222c2b559a49fbfba26e24d544/openfisca_core/simulations.py#L307

This is part of `_check_for_cycle`, a utility method on the Simulation object which holds the context of a computation; it is called as a guard clause at each step of the recursive operation of the API method `Simulation. calculate`.

The condition under which a SpiralError is raised is that the computation of a variable at a period P involves the computation of the same variable at a period P'. With the default value of `max_spiral_loops` your test case is indeed considered problematic. (I agree about the naming; it's not ideal.)

That setting can only be changed in the Python API, not via YAML; that could be a useful feature to add (and possibly to the Web API as well).

[...]

PR #1385, opened 2026-09-09
Fix `Simulation.clone` sharing mutable state with the original simulation

### LINKED ISSUE #1155: Possible bug in `clone` logic
Sometimes when cloning a simulation, I've found that the original simulation is inexplicably affected by actions in the cloned simulation. I've seen this for a while but have been unable to generate a MWE, but yesterday I think I found the issue, in [this line](https://github.com/openfisca/openfisca-core/blob/bee45222babbd8d9727feace04448264fdb89685/openfisca_core/populations/group_population.py#L21). When a group population is cloned, it looks like the clone still has access to the original `members`, which is unchanged.

Adelagric:
Fixes #1155.

### Problem

`Simulation.clone()` promises to "copy the simulation just enough to be able to run the copy without modifying the original simulation", but the clone actually shares most of its mutable state with the original. Four distinct defects compound:

1. **`Holder.clone` shares storage by reference**: the cloned holder reuses the original's `InMemoryStorage` and `OnDiskStorage` objects. Any `set_input`, `put_in_cache` or `delete_arrays` through one simulation silently alters the other. Worse, the original can end up internally inconsistent: change `salary` through a clone and the original now returns the new salary but keeps its stale cached `income_tax`.
2. **`GroupPopulation.clone` binds cloned holders to the original population**: `holder.clone(self)` is called instead of `holder.clone(result)`, so the clone's group-entity holders have `population` and `simulation` pointing at the original simulation.
3. **`GroupPopulation.clone` keeps the original persons as `members`** — this is the mechanism @nikhilwoodruff suspected in #1155: `clone.household.members is original.persons`.
4. **The clone shares `_data_storage_dir`**: when both simulations lazily create a holder for the same variable (with a `memory_config` active), their two distinct `OnDiskStorage` objects write `{period}.npy` files into the *same* directory. One simulation then reads the other's values, and garbage-collecting either simulation `rmtree`s the directory, physically destroying the other's data.

Minimal reproduction of defect 1 (before this fix):

```python
simulation = SimulationBuilder().build_from_entities(tbs, situation)  # salary = 2000
simulation.calculate("income_tax", "2026-01")   # 300, cached
clone = simulation.clone()
clone.delete_arrays("salary", "2026-01")
clone.set_input("salary", "2026-01", [4000])
simulation.calculate("salary", "2026-01")       # 4000 — the original was modified
simulation.calculate("income_tax", "2026-01")   # 300 — stale, inconsistent with salary
```

### Fix

- `InMemoryStorage.clone()` / `OnDiskStorage.clone(storage_dir)`: new methods returning independent copies. The in-memory clone copies the period-to-array mapping only (arrays stay shared, as the engine treats them as immutable). The on-disk clone copies the `.npy` files into the new directory and re-registers the enum mapping under the new directory key, so `EnumArray` values survive the copy.
- `Holder.clone` uses them instead of sharing references.
- `GroupPopulation.clone` binds holders to the cloned population and takes the cloned persons (`simulation.persons`) as `members`.
- `Simulation.clone` resets `_data_storage_dir` (the clone lazily gets its own temporary directory) and copies `invalidated_caches`.

### Trade-offs

- Cloning a simulation with on-disk storage now copies the stored files (I/O cost proportional to spilled data). The alternative — sharing files read-only — was rejected because either simulation's cleanup deletes the other's files.
- Arrays themselves are still shared between original and clone, consistently with how they are already shared within a single simulation's cache.
- Code that *relied* on the sharing behaviour (e.g. writing into a clone to warm the original's cache) will observe a change; that behaviour contradicted the documented contract of `clone`.

### Test plan

- 6 new tests in `tests/core/test_simulations.py`, each failing on `master` (except the enum guard, which protects the new copy logic): holder-storage isolation, group holders bound to the clone, cloned `members`, `invalidated_caches` copy, on-disk isolation incl. garbage-collection of the clone, enum preservation on disk clone.
- Full test suite and doctests pass; `isort`/`black`/`flake8`/`pylint` clean on the changed files.

### DIFF
--- a/openfisca_core/populations/group_population.py
+++ b/openfisca_core/populations/group_population.py
@@ -20,10 +20,13 @@ class GroupPopulation(Population):
         self._ordered_members_map = None
 
     def clone(self, simulation):
-        result = GroupPopulation(self.entity, self.members)
+        # ``simulation.persons`` must already be the cloned persons population,
+        # so that members of the cloned group belong to the new simulation.
+        result = GroupPopulation(self.entity, simulation.persons)
         result.simulation = simulation
         result._holders = {
-            variable: holder.clone(self) for (variable, holder) in self._holders.items()
+            variable: holder.clone(result)
+            for (variable, holder) in self._holders.items()
         }
         result.count = self.count
         result.ids = self.ids
--- a/tests/core/test_simulations.py
+++ b/tests/core/test_simulations.py
@@ -1,8 +1,12 @@
+import gc
+
+import numpy
 import pytest
 
 from openfisca_country_template.situation_examples import single
 
 from openfisca_core import errors, periods
+from openfisca_core.memory_config import MemoryConfig
 from openfisca_core.simulations import SimulationBuilder
 
 
@@ -57,6 +61,101 @@ def test_clone(tax_benefit_system) -> None:
     assert salary_holder != salary_holder_clone
     assert salary_holder_clone.simulation == simulation_clone
     assert salary_holder_clone.population == simulation_clone.persons
+
+
+def new_simulation(tax_benefit_system):
+    return SimulationBuilder().build_from_entities(
+        tax_benefit_system,
+        {
+            "persons": {
+                "bill": {"salary": {"2017-01": 3000}},
+            },
+            "households": {"household": {"adults": ["bill"]}},
+        },
+    )
+
+
+def test_clone_does_not_share_holder_storage(tax_benefit_system) -> None:
+    simulation = new_simulation(tax_benefit_system)
+    income_tax_before = simulation.calculate("income_tax", "2017-01")
+
+    clone = simulation.clone()
+
+    # Writing in the clone must not pollute the original
+    clone.delete_arrays("salary", "2017-01")
+    clone.set_input("salary", "2017-01", numpy.array([6000.0]))
+    assert simulation.calculate("salary", "2017-01") == 3000
+    assert simulation.calculate("income_tax", "2017-01") == income_tax_before
+
+    # Deleting in the clone must not destroy the original's values
+    clone.delete_arrays("salary")
+    assert simulation.get_array("salary", "2017-01") is not None
+
+
+def test_clone_group_entity_holders_bound_to_clone(tax_benefit_system) -> None:
+    simulation = new_simulation(tax_benefit_system)
+    simulation.calculate("housing_tax", "2017")
+
+    clone = simulation.clone()
+    holder_clone = clone.household.get_holder("housing_tax")
+
+    assert holder_clone.population is clone.household
+    assert holder_clone.simulation is clone
+
+
+def test_clone_members_are_cloned(tax_benefit_system) -> None:
+    simulation = new_simulation(tax_benefit_system)
+
+    clone = simulation.clone()
+
+    assert clone.household.members is clone.persons
+
+
+def test_clone_does_not_share_invalidated_caches(tax_benefit_system) -> None:
+    simulation = new_simulation(tax_benefit_system)
+
+    clone = simulation.clone()
+
+    assert clone.invalidated_caches is not simulation.invalidated_caches
+
+
+@pytest.mark.filterwarnings("ignore")
+def test_clone_disk_storage_is_isolated(tax_benefit_system) -> None:
+    simulation = new_simulation(tax_benefit_system)
+    simulation.memory_config = MemoryConfig(max_memory_occupation=0.0)
+    # Materialize the on-disk storage before cloning, so that the clone
+    # actually has to disentangle itself from the original's directory.
+    simulation.set_input("age", "2017-01", numpy.array([30]))
+
+    clone = simulation.clone()
+
+    clone.delete_arrays("age", "2017-01")
+    clone.set_input("age", "2017-01", numpy.array([99]))
+    assert simulation.calculate("age", "2017-01") == 30
+
+    # "birth" has no holder yet: each simulation lazily creates its own,
+    # which must not end up writing into the same directory.
+    simulation.set_input("birth", "2017-01", numpy.array(["1987-01-01"]))
+    clone.set_input("birth", "2017-01", numpy.array(["1999-01-01"]))
+    assert simulation.calculate("birth", "2017-01") == numpy.datetime64("1987-01-01")
+
+    # Garbage-collecting the clone must not destroy the original's files
+    del clone
+    gc.collect()
+    assert simulation.calculate("age", "2017-01") == 30
+    assert simulation.calculate("birth", "2017-01") == numpy.datetime64("1987-01-01")
+
+
+@pytest.mark.filterwarnings("ignore")
+def test_clone_disk_storage_preserves_enums(tax_benefit_system) -> None:
+    simulation = new_simulation(tax_benefit_system)
+    simulation.memory_config = MemoryConfig(max_memory_occupation=0.0)
+    simulation.set_input("housing_occupancy_status", "2017-01", numpy.array(["tenant"]))
+
+    clone = simulation.clone()
+
+    status = clone.calculate("housing_occupancy_status", "2017-01")
+    assert status.decode_to_str() == ["tenant"]
 
 
 def test_get_memory_usage(tax_benefit_system) -> None:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## 44.7.2 [#1385](https://github.com/openfisca/openfisca-core/pull/1385)
+
+#### Bug fixes
+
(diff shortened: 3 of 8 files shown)


Adelagric:
@nikhilwoodruff since you reported #1155 and have been living with this at PolicyEngine — if you have a chance to run this branch against your workloads, that would be a much stronger validation than my regression tests alone. Happy to adjust anything that surfaces.

guillett (approved):
That looks good to me. Many thanks @Adelagric

Adelagric:
Thanks @guillett! One thing: the CI workflows haven't run on this PR yet — as a first-time contributor they need a maintainer to approve the run. Happy to rebase if the version bump conflicts with anything landing before this.

While you're at it: #1386 (spiral warning, #956) and #1387 (`restore_simulation` counts) are the two sibling PRs from the same pass, also awaiting a first look.

tests/core/test_simulations.py
```diff
@@ -59,6 +63,101 @@ def test_clone(tax_benefit_system) -> None:
...
+    assert simulation.calculate("birth", "2017-01") == numpy.datetime64("1987-01-01")
+
+    # Garbage-collecting the clone must not destroy the original's files
+    del clone
+    gc.collect()
+    assert simulation.calculate("age", "2017-01") == 30
+    assert simulation.calculate("birth", "2017-01") == numpy.datetime64("1987-01-01")
+
+
+@pytest.mark.filterwarnings("ignore")
```
guillett: Why are warnings ignored here?

Adelagric: Good catch — no good reason. The only warning raised there is `MemoryConfigWarning` (the "experimental feature" notice), and `test_holders.py` already uses `MemoryConfig(max_memory_occupation=0)` without filtering anything. Removed in the latest commit.

guillett: Did you try to run tests directly on your own machine?

guillett: I'm trying to figure out, I much AI assisted you are.

Adelagric: Fair question, and I'd rather answer it plainly than have it hang over the PR.

Yes, the tests were run locally several ways (`pytest`, `make test-core`, full suite), all green with the project config. The `filterwarnings` was a defensive reflex from a run with `-W error`, where `MemoryConfigWarning` fails the disk tests; since `test_holders.py` uses `MemoryConfig` without filtering, I removed it to follow the existing convention.

On AI: I use AI tooling in my workflow, the way I'd use an IDE — but the analysis is mine, and I can defend every line. The five compounding defects behind #1155 were each reproduced by hand before a line of fix was written; the subtle one is the shared `_data_storage_dir`, where fixing the holder storage alone just moves the bug to the filesystem. Ask me anything about it.

To be clear about where I stand: I'm here because these are real silent-data-loss bugs in a codebase people rely on for policy decisions, not for the merge count. If the project would rather not take contributions from people who use these tools, say so and I'll respect it — no hard feelings either way. If the concern is maintainability, I'm happy to stay reachable on this code after merge.

Adelagric: Fair question. Yes — the tests were run locally several ways (`pytest`, `make test-core`, full suite), all green with the project config. The `filterwarnings` came from a run with `-W error`, where `MemoryConfigWarning` fails the disk tests; since `test_holders.py` uses `MemoryConfig` without filtering, I removed it to follow the convention.

On AI: I use AI tooling in my workflow, the way I'd use an IDE. The analysis is mine and I can defend every line — the five compounding defects behind #1155 were each reproduced by hand before any fix was written, and the subtle one (the shared `_data_storage_dir`, where fixing the holder storage alone just moves the bug to the filesystem) is the kind of thing you only find by running it. Ask me anything about it.

I'm here because these are real silent-data-loss bugs in code people rely on for policy decisions. If the project prefers not to take contributions from people who use these tools, just say so — I'll respect it. Otherwise I'm happy to stay reachable on this code after merge.

openfisca_core/populations/group_population.py
guillett: Shoudn't these cloned? I guess they are array references, no?

Adelagric: Array references, yes — this block predates the PR. Nothing in core mutates them in place (every write path reassigns a new array; the indexed write in `members_position` is on a freshly built one), so sharing them was safe under the same immutability policy as holder values. It's a convention, not a guarantee, and copying costs nothing: done in e14e0d6, with a test.

MattiSG:
Thanks for the clarifications @Adelagric.

As you said, this is “code people rely on for policy decisions”, and unfortunately that means we have to deal with real risks of reputation farming, privilege escalation, or even exploit planting (prepositioning). Core was never, until LLM assistance, the entry point for new contributors due to the complexity of the codebase and the potential global impact of changes.
It is super exciting to have new contributors, and we're glad to have you come along as someone who seems to be able to participate in discussions beyond dumping an LLM-generated changeset 🙂 However, we also need to be very careful about supply chain vulnerabilities, and about the work of volunteers who have to handle the consequences of a bug in a new deployed version (aka “I'm not super keen on another sleepless night of emergency rollbacks across deployment sites and PR management on behalf of an anonymous third party” 😄).

We are slowly learning to deal with the size and complexity of AI-generated contributions, but this is an area under active definition, and I'm sorry that everything is not super clear up front. Please bear with us!

In particular, the OpenFisca community is trying to define its approach with regards to AI-assisted contributions. You can find of our current line of thinking in #1372. In particular, defining the level of assistance and model(s) used would be helpful, as @guillett mentioned in https://github.com/openfisca/openfisca-core/pull/1385#discussion_r4007349437 🙂 

It would also make a significant difference in the trust and effort we can allocate to such changesets if we understood a bit more of your context for using OpenFisca and contributing to it 🙂 I could gather from your [public website](https://kaleche.dev/) that you claim to be based in France, but a little hi and an intro in our community chat would be great as well! I sent you an invite to your publicly listed email. Feel free to join there or drop a line on contact@openfisca.org. In particular, understanding which production uses you have of OpenFisca, what's your experience with it, and what motivates you to contribute to Core would be very helpful.

Note that this is not an official precondition as we have not made this into a strict prerequisite policy, and I'll let @guillett and other Core maintainers judge the relevance of the changeset itself. I just hope to bring some guidance on what can make the integration of your work into the codebase smoother 🙂 OpenFisca is not merely an open source project, it is a digital common and that means governance and the human connections that make the codebase evolve are an integral part of it.

Have a good day! 😊

MattiSG:
I've checked the list of changed files: none modify CI. I've checked that the introduced code does not make suspicious new logs (secret extraction) or remote calls: all good.
I've thus approved workflows to run 🙂 CI will report in a bit.

Adelagric:
Thanks @MattiSG — both for taking the time to write this and for reviewing the diff and approving the workflows. I read #1372, and I appreciate that the policy is being worked out in the open.

**On AI usage:** under the RFC terminology, I would classify the submitted fix and tests as 🟠 AI-generated: I use an LLM coding assistant under my direction to write code, while the investigation and reproduction work here was mine. Every defect in #1155 was first reproduced independently against the actual code; the fix was then developed test-first, with the tests confirmed red on `master` before going green with the change. I also run an adversarial review pass before submitting — in this case it caught a disk test that was not actually discriminating.

Everything I submit is reread, executed and verified by me, and I consider myself responsible for it. I can explain every line and maintain the code afterward.

For tooling, I use an agentic coding setup with an LLM assistant. I'm happy to disclose the exact tool/model explicitly once the contribution template has a place for it.

**On who I am:** I'm Adel Kaleche, based in Occitanie, France. I teach part-time and run a small software and operations-research consultancy on the side (kaleche.dev). I do not use OpenFisca in production, so I want to be explicit about that.

Most of my open-source work has converged on the same kind of problem: silent data loss and silent failures. I have contributed fixes in projects such as rust-bio, Rostam, conda and x402, and I currently have work open in several data/infrastructure projects. The recurring pattern is the same: find a case where the system loses, corrupts or mis-associates data without failing loudly; reproduce it deterministically; then fix it behind a red-then-green regression test.

**Why OpenFisca:** part of the work I'm building a specialty around sits at the boundary between probabilistic and deterministic systems. On the probabilistic side, I recently wrote a preprint on detecting silent distribution drift: https://arxiv.org/abs/2608.01268. On the deterministic side, I tend to hunt exactly the kind of failures I found in the clone/dump/spiral paths here.

As more code is produced with AI assistance, I think this boundary — systems appearing to work while silently producing the wrong state — is going to matter more, not less.

The second reason is more personal: France took me in, and contributing to a digital common the country relies on felt like a small way to give something back.

**On future involvement:** I can't promise a regular contribution rate — teaching and client work make that difficult — but I do consider myself responsible for the code I submit. I'll stay available for review questions, regressions, or follow-up work on #1385, #1386 and #1387 after merge.

Beyond that, I'll probably continue following this same thread across other codebases whenever I find another silent failure worth digging into.

I'll say hi in the community chat as well. Thanks for the invitation.

PR #1386, opened 2026-09-09
Warn when a spiral substitutes a variable's default value

### LINKED ISSUE #956: Spiraling

Adelagric:
Closes #956.

### Problem

When a circular calculation is detected (`SpiralError`), the engine silently returns the variable's default value instead of evaluating its formula ([simulation.py](https://github.com/openfisca/openfisca-core/blob/master/openfisca_core/simulations/simulation.py#L166)). As reported in #956, nothing tells the user that a formula they wrote was skipped — models end up wrong in ways that are very hard to notice (the reporter only found out by diffing test results that depended on calculation order).

### Fix

The substitution now emits a `SpiralWarning` (new, exported from `openfisca_core.warnings`, following the existing `TempfileWarning` pattern) that names the variable and period whose formula was skipped, and explains the two ways out: rewrite the formula to avoid the recursion, or allow deeper recursion with `simulation.max_spiral_loops` — which, since #956 was opened, can also be set in YAML tests with the `max_spiral_loops` keyword, so the warning mentions that too.

The calculation behaviour is unchanged: same results, same caching, same purge of invalidated values. Users who rely on spirals intentionally can filter the warning out with `warnings.filterwarnings("ignore", category=SpiralWarning)`.

```
SpiralWarning: A circular calculation was detected while computing 'montant_total' for period 2017:
the formula was not evaluated, and the variable's default value was returned instead. If this is
unexpected, rewrite the formula to avoid the recursion (e.g. with a closed-form expression or an
explicit check on the requested period), or allow deeper recursion by raising
`simulation.max_spiral_loops` (currently 1; also settable in YAML tests with the `max_spiral_loops` keyword).
```

### Test plan

- New test in `tests/core/test_cycles.py` asserting the warning is emitted on a spiraling calculation (fails without this change).
- Reproduced the exact scenario from #956 (a yearly variable whose `formula_2018` depends on `period.last_year`) and checked the warning is now emitted where there used to be silence.
- Full test suite passes; `isort`/`black`/`flake8` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## 44.8.0
+
+#### New features
+
+- Emit a `SpiralWarning` when a circular calculation is detected and a variable's default value is silently returned instead of evaluating its formula (closes [#956](https://github.com/openfisca/openfisca-core/issues/956)).
+  - Previously the substitution was completely silent, so there was no way to know that a formula had not been used.
+  - The warning names the variable and period whose formula was skipped, and explains how to avoid the recursion or raise `max_spiral_loops` (including from YAML tests).
+  - The calculation behaviour is unchanged; users who expect spirals can silence the warning with `warnings.filterwarnings("ignore", category=SpiralWarning)`.
+
 ## 44.7.1 [#1383](https://github.com/openfisca/openfisca-core/pull/1383)
 
 #### Bug fixes
--- a/openfisca_core/simulations/simulation.py
+++ b/openfisca_core/simulations/simulation.py
@@ -165,6 +165,25 @@ class Simulation:
 
         except errors.SpiralError:
             array = holder.default_array()
+            message = [
+                (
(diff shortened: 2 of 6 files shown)


Adelagric:
@magemax @benjello you were both directly affected in #956 — does the warning message and its guidance look right to you? The behaviour is unchanged, this only makes the substitution visible.

PR #1387, opened 2026-09-09
Fix `restore_simulation` entity counts

Adelagric:
### Problem

`restore_simulation` rebuilds each group entity's `count` from the members mapping (`max(members_entity_id) + 1`) and the persons' `count` from the members of the last group entity restored. Two consequences:

- **A group entity with no member is silently dropped** from the restored simulation's `count`, while still listed in its `ids` — an internally inconsistent state. Calculations on the restored simulation then return arrays with missing entities, with no error:

```python
simulation.household.count            # 2 (h1, h2 — h2 has no member)
dump_simulation(simulation, directory)
restored = restore_simulation(directory, tbs)
restored.household.count              # 1 — h2 silently gone
restored.calculate("total_benefits", "2017-01")  # 1 value instead of 2
```

When a value *was* stored for the group entity, restoring instead fails with `ValueError: length is 2 while there are 1 households`.

- **A tax and benefit system with no group entity cannot be restored at all**: the `person_count` local is never assigned and `restore_simulation` raises `UnboundLocalError`.

### Fix

The dump already contains each entity's `ids` (`id.npy`); each entity's `count` is now restored from them. This also removes the `person_count` plumbing and the ordering constraint between the two restore passes.

### Test plan

Two new tests in `tests/core/test_dump_restore.py`, both failing on `master`: roundtrip with a memberless household (count and stored group values preserved), and roundtrip of a person-only tax and benefit system.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## 44.7.2
+
+#### Bug fixes
+
+- Fix `restore_simulation` rebuilding entity counts from the members mapping instead of the dumped ids.
+  - A group entity with no member was silently dropped from the restored simulation's `count` (while still listed in `ids`), so subsequent calculations returned arrays with missing entities — or failed with a `ValueError` when a value had been stored for the group entity.
+  - Restoring a simulation whose tax and benefit system has no group entity raised an `UnboundLocalError`.
+  - Each entity's `count` is now restored from its dumped `ids`.
+
 ## 44.7.1 [#1383](https://github.com/openfisca/openfisca-core/pull/1383)
 
 #### Bug fixes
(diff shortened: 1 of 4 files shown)


Adelagric:
For prioritization context: the main consumer of `dump_simulation`/`restore_simulation` I could find is openfisca-survey-manager (`abstract_scenario.py`), where simulations run on survey data — a setting where a group entity with no member is much more plausible than in handcrafted test cases (several group entity types, filtered datasets). A silently shrunk entity count there means silently wrong aggregates after a restore, with `ids` still listing the dropped entities.

PolicyEngine's `policyengine-core` forked this module verbatim and carries the same behaviour; I'm reporting it to them separately.

PR #1392, opened 2026-09-17
Remove conda support

### LINKED ISSUE #1384: [RFC] Drop Conda support
According to https://anaconda.org/channels/openfisca/packages/openfisca-core/overview there is 0 download in the last six month.

Conda build in the CI is specific and consume time to maintain.

For example it's broken right now : https://github.com/openfisca/openfisca-core/actions/runs/32033678797/job/100930956941?pr=1383 (fixed by https://github.com/openfisca/openfisca-core/pull/1383/commits/8fde2db98079f2756e630b2f03687d5a5c73b58d )

That's why I suggest to drop it.

guillett:
#### Breaking changes

- Remove Conda support

It is funny because it is definitely a breaking change but as no newer version will be publishd on Conda what version update should we really apply.

Fixes #1384


guillett:
Parameter additions in the country template are breaking CI. Cf #1393 that should remove the issue (or more precisely move it to the dedicated package"

benoit-cty (approved):
conda removal is good.

`setup.py` and `.conda/*` are not in `IGNORE_DIFF_ON`, so check-version break. A bump for technical change is needed.

benoit-cty:
Thanks Thomas !

May I suggest to keep the test-on-windows by replacing Conda with UV ?

In, merge.yaml ` publish-to-pypi:` add

```
    outputs:
      version: ${{ steps.get-version.outputs.version }}
[...]

    - name: Get version
      id: get-version
      run: |
        source venv/bin/activate
        echo "version=$(python setup.py --version)" >> $GITHUB_OUTPUT

```

Then:

```

  test-on-windows:
    runs-on: windows-2025
    needs: [publish-to-pypi]
    defaults:
      run:
        shell: bash

    steps:
    - name: Install uv
      uses: astral-sh/setup-uv@v6

    - name: Install openfisca-core from PyPi
      env:
        VERSION: ${{ needs.publish-to-pypi.outputs.version }}
      # PyPi's index can lag a few seconds behind an upload, hence the retries.
      run: |
        uv venv --python ${{ env.UPPER_PYTHON_VERSION }}
        installed=false
        for attempt in $(seq 1 5); do
          if uv pip install "openfisca-core==$VERSION"; then
            installed=true
            break
          fi
          echo "openfisca-core $VERSION not on PyPi yet, retrying in 30s (attempt $attempt/5)…"
          sleep 30
        done
        if [ "$installed" = false ]; then
          echo "Could not install openfisca-core $VERSION from PyPi."
          exit 1
        fi

    - name: Test openfisca
      run: |
        source .venv/Scripts/activate
        openfisca --help

```

I can do the commit if you want.

guillett:
Thanks @benoit-cty I added your suggestion myself because I would rather have you help on #1393

guillett:
@benoit-cty I'm not sure if you're ok with a patch version increment so I requested another review from you. Thanks!

benoit-cty (approved):
I'm OK, thanks for asking 👍

PR #1394, opened 2026-09-21
Fix Windows install check in continuous deployment

guillett:
#### Technical changes

- Fix Windows install in continuous deployment


guillett:
@benoit-cty we were in a hurry my changes in the continuous deployment pipeline of #1392 were not fully functional.

benoit-cty (approved):
CI is always a pain to update...

guillett:
I actually tested a too basic use case. git tags are currently not published. It had to be fixed before merging this PR as is. I included a _temporary fix_ that we'll be able to remove when moving completely to uv.

PR #1396, opened 2026-09-21
Migrate to pyproject.yaml and use uv in CI/CD

guillett:


benoit-cty:
Hello, don't know if you have finish, here is some pre-review:

- `uv.lock` The lockfile isn't excluded from codespell. Bare codespell in make check-style reports astroid ==> asteroid at lines 13 and 2101 and exits 65 — deterministic lint failure.
- `pyproject.toml` skip = "./venv" was dropped from [tool.codespell]. The lint job restores venv/ into the repo root before make check-style, so codespell now scans the whole virtualenv (confirmed: a typo under venv/ is reported on this branch, skipped on master).
- `pyproject.toml` setuptools >= 61.0 is too low for the PEP 639 license = "AGPL-3.0-or-later" string plus license-files. With setuptools 70 it's a hard ValueError: invalid pyproject.toml config: 'project.license'. Needs >= 77.0.3.
- `pyproject.toml` Stray [dependency-groups] dev = ["ruff>=0.16.7"] shadows the dev extra: uv sync installs ruff only, not the lint/type toolchain, and declares a second conflicting ruff constraint.
- `.github/workflows/_before-pip.yaml` Cache keys still hash setup.py, but dependencies now live in pyproject.toml, so dependency changes no longer bust the cache. Same in _lint-pip.yaml and merge.yaml.
- `pyproject.toml` New template extra has uncapped versions, against the file's own capping rule (and the templates depend back on openfisca-core[web-api]).

guillett:
Sorry, I forgot to set this PR as draft. It is not ready at all and I don't know if I'm tacking this the right way.

benoit-cty:
Ok, you may have a look to https://github.com/openfisca/openfisca-france/pull/2547

